PR #2, opened 2021-09-24
Merge jdk-17+35

AntonKozlov:
Merge tag jdk-17+35. All CRaC-specific tests pass, although there are only few of them.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/crac pull/2/head:pull/2` \
`$ git checkout pull/2`

`$ git checkout pull/2` \
`$ git pull https://git.openjdk.java.net/crac pull/2/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2`

`$ git pr show -t 2`

</details>

<a href="https://git.openjdk.java.net/crac/pull/2.diff">https://git.openjdk.java.net/crac/pull/2.diff</a>

</details>


AntonKozlov:
I'm going to tag the merge as `crac-17+1`. That is, first build of JDK17 with CRaC.

DanHeidinga:
@AntonKozlov What's the preferred process on this kind of change?  I'm new to managing merges of the JDK mainline into project specific repos so not yet clear on what level of review something like this would need.  

> All CRaC-specific tests pass, although there are only few of them.

As this is a merge of mainline (well, 17) I'm OK with giving this the +1 assuming that's the process we typically follow for this?

Do projects typically follow the releases or the mainline development?  Still learning the norms here :)

AntonKozlov:
> What's the preferred process on this kind of change? I'm new to managing merges of the JDK mainline into project specific repos so not yet clear on what level of review something like this would need.

I'm not sure also :) But I wanted to provide a heads up at least for the bump of the major version. Merges of minor versions probably are OK without review, if we'll do them often.

> > All CRaC-specific tests pass, although there are only few of them.
> 
> As this is a merge of mainline (well, 17) I'm OK with giving this the +1 assuming that's the process we typically follow for this?

Yeah, I think this would be useful, a simple acknowledgment from other member(s). 

> Do projects typically follow the releases or the mainline development? Still learning the norms here :)

We are free to choose something most suitable for us. I think having both makes sense: following mainline development for fast resolution of the problems with merges and a stable release for users to try. However, it is an additional burden in maintenance. Having 17, the merge of the early pre-18 state should be simple, the opposite looks harder. So for following 17 looks useful but simple enough. We can extend this any time we'd want, I think.

DanHeidinga:
Seems like a reasonable starting point.

AntonKozlov:
> github repeatedly fails to load the "files" tab with a "This page is taking too long to load." so I'm unable to actually review the PR in the UI. Anyone know of a workaround for this? 

Funny enough, it's actually impossible to provide review (e.g. after checking out changes locally) without github attempt to show the diff.

> Otherwise, please consider this reviewed.

OK, thanks!

PR #5, opened 2021-11-25
Invalidate and reseed SHA1PRND during checkpoint/restore

alexeybakhtin:
This is a proposal for the CRaC Critical Section interface and its usage in the SHA1PRNG SecureRandom implementation.
The changes in the SecureRandom implementation allow invalidating and reseeding SHA1PRNG secure random during checkpoint/restore.  SHA1PRNG can be invalidated and reseeded in case of being created with a default embedded seed generator. Also, SHA1PRNG is used as an additional seed generator to the SUN NativePRNG implementation, so it is desirable to have reseeded SHA1PRNG after restore.

### DIFF
--- a/src/java.base/share/classes/jdk/crac/CheckpointLock.java
+++ b/src/java.base/share/classes/jdk/crac/CheckpointLock.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2017, 2021, Azul Systems, Inc. All rights reserved.
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package jdk.crac;
+
+/**
+ * A {@code CheckpointLock} allows creating critical sections that are not interrupted during checkpoint/restore.
+ */
+public class CheckpointLock implements AutoCloseable {
+
+    /**
+     * Creates a {@code CheckpointLock}.
+     */
+    public CheckpointLock() {
+        jdk.internal.crac.Core.getJDKContext().addLock(this);
+    }
+
+    @Override
+    public void close() {
+        jdk.internal.crac.Core.getJDKContext().removeLock(this);
+    }
+}
--- a/src/java.base/share/classes/jdk/crac/impl/AbstractContextImpl.java
+++ b/src/java.base/share/classes/jdk/crac/impl/AbstractContextImpl.java
@@ -23,6 +23,7 @@
 package jdk.crac.impl;
 
 import jdk.crac.CheckpointException;
+import jdk.crac.CheckpointLock;
 import jdk.crac.Context;
 import jdk.crac.Resource;
 import jdk.crac.RestoreException;
@@ -44,6 +45,9 @@ public abstract class AbstractContextImpl<R extends Resource, P> extends Context
     private WeakHashMap<R, P> checkpointQ = new WeakHashMap<>();
     private List<R> restoreQ = null;
     private Comparator<Map.Entry<R, P>> comparator;
+    private List<CheckpointLock> lockQ = new LinkedList<CheckpointLock>();
+    private Object enterCriticalSectionLock = new Object();
+    private Object exitCriticalSectionLock = new Object();
 
     protected AbstractContextImpl(Comparator<Map.Entry<R, P>> comparator) {
         this.comparator = comparator;
@@ -55,32 +59,44 @@ public abstract class AbstractContextImpl<R extends Resource, P> extends Context
 
     @Override
     public synchronized void beforeCheckpoint(Context<? extends Resource> context) throws CheckpointException {
-        List<R> resources = checkpointQ.entrySet().stream()
-            .sorted(comparator)
-            .map(Map.Entry::getKey)
-            .collect(Collectors.toList());
-
-        CheckpointException exception = new CheckpointException();
-        for (Resource r : resources) {
-            if (DEBUG) {
-                System.err.println("jdk.crac beforeCheckpoint " + r.toString());
-            }
-            try {
-                r.beforeCheckpoint(this);
-            } catch (CheckpointException e) {
-                for (Throwable t : e.getSuppressed()) {
-                    exception.addSuppressed(t);
+        synchronized (enterCriticalSectionLock) {
+            synchronized (exitCriticalSectionLock) {
+                while (!lockQ.isEmpty()) {
+                    try {
+                        exitCriticalSectionLock.wait();
+                    } catch (InterruptedException iex) {
+                        // just skip
+                    }
                 }
-            } catch (Exception e) {
-                exception.addSuppressed(e);
             }
-        }
 
-        Collections.reverse(resources);
-        restoreQ = resources;
+            List<R> resources = checkpointQ.entrySet().stream()
+                    .sorted(comparator)
+                    .map(Map.Entry::getKey)
+                    .collect(Collectors.toList());
 
-        if (0 < exception.getSuppressed().length) {
-            throw exception;
+            CheckpointException exception = new CheckpointException();
+            for (Resource r : resources) {
+                if (DEBUG) {
+                    System.err.println("jdk.crac beforeCheckpoint " + r.toString());
+                }
+                try {
+                    r.beforeCheckpoint(this);
+                } catch (CheckpointException e) {
+                    for (Throwable t : e.getSuppressed()) {
+                        exception.addSuppressed(t);
+                    }
+                } catch (Exception e) {
+                    exception.addSuppressed(e);
+                }
+            }
+
+            Collections.reverse(resources);
+            restoreQ = resources;
+
+            if (0 < exception.getSuppressed().length) {
+                throw exception;
+            }
         }
     }
 
@@ -105,6 +121,23 @@ public abstract class AbstractContextImpl<R extends Resource, P> extends Context
 
         if (0 < exception.getSuppressed().length) {
             throw exception;
+        }
+    }
+
+
+    public void addLock(CheckpointLock lock) {
+        synchronized(enterCriticalSectionLock) {
+            synchronized (lockQ) {
+                lockQ.add(lock);
+            }
+        }
+    }
+    public void removeLock(CheckpointLock lock) {
+        synchronized(exitCriticalSectionLock) {
+            synchronized (lockQ) {
+                lockQ.remove(lock);
+            }
+            exitCriticalSectionLock.notify();
         }
     }
 }
--- a/src/java.base/share/classes/sun/security/provider/SecureRandom.java
+++ b/src/java.base/share/classes/sun/security/provider/SecureRandom.java
@@ -25,6 +25,10 @@
 
 package sun.security.provider;
 
+import jdk.crac.CheckpointLock;
+import jdk.crac.Context;
+import jdk.crac.Resource;
+
 import java.io.IOException;
 import java.security.MessageDigest;
 import java.security.SecureRandomSpi;
@@ -53,7 +57,7 @@ import java.security.NoSuchProviderException;
  */
 
 public final class SecureRandom extends SecureRandomSpi
-implements java.io.Serializable {
+implements java.io.Serializable, jdk.internal.crac.JDKResource {
 
     @java.io.Serial
     private static final long serialVersionUID = 3581829991155417889L;
@@ -63,6 +67,8 @@ implements java.io.Serializable {
     private byte[] state;
     private byte[] remainder;
     private int remCount;
+    private boolean clearStateOnCheckpoint = true;
+    private boolean reseedAfterRestore = false;
 
     /**
      * An empty constructor that creates an unseeded SecureRandom object.
@@ -119,6 +125,7 @@ implements java.io.Serializable {
         if (seed != null) {
            engineSetSeed(seed);
         }
+        jdk.internal.crac.Core.getJDKContext().register(this);
     }
 
     /**
@@ -154,15 +161,19 @@ implements java.io.Serializable {
      * @param seed the seed.
      */
     @Override
+    @SuppressWarnings("try")
     public synchronized void engineSetSeed(byte[] seed) {
-        if (state != null) {
-            digest.update(state);
-            for (int i = 0; i < state.length; i++) {
-                state[i] = 0;
+        try (CheckpointLock lock = new CheckpointLock() ) {
+            if (state != null) {
+                digest.update(state);
+                for (int i = 0; i < state.length; i++) {
+                    state[i] = 0;
+                }
             }
+            state = digest.digest(seed);
+            remCount = 0;
+            clearStateOnCheckpoint = false;
         }
-        state = digest.digest(seed);
-        remCount = 0;
     }
 
     private static void updateState(byte[] state, byte[] output) {
@@ -190,6 +201,39 @@ implements java.io.Serializable {
         }
     }
 
+    @Override
+    public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+        if (clearStateOnCheckpoint) {
+            if (state != null) {
+                for (int i = 0; i < state.length; i++) {
+                    state[i] = 0;
+                }
+            }
+            state = null;
+            if (remainder != null) {
+                for (int i = 0; i < remainder.length; i++) {
+                    remainder[i] = 0;
+                }
+            }
+            remainder = null;
+            remCount = 0;
+        }
+    }
+
+    @Override
+    public void afterRestore(Context<? extends Resource> context) throws Exception {
+        if (reseedAfterRestore) {
+            byte[] b = new byte[DIGEST_SIZE];
+            SeedGenerator.generateSeed(b);
+            engineSetSeed(b);
+        }
+    }
+
+    @Override
+    public int getPriority() {
+        return 0;
+    }
+
     /**
      * This static object will be seeded by SeedGenerator, and used
      * to seed future instances of SHA1PRNG SecureRandoms.
@@ -206,10 +250,13 @@ implements java.io.Serializable {
              * seed material (likely from the Native implementation).
              */
             seeder = new SecureRandom(SeedGenerator.getSystemEntropy());
-            byte [] b = new byte[DIGEST_SIZE];
+            seeder.reseedAfterRestore = true;
+            byte[] b = new byte[DIGEST_SIZE];
             SeedGenerator.generateSeed(b);
             seeder.engineSetSeed(b);
+            seeder.clearStateOnCheckpoint = true;
         }
+
     }
 
     /**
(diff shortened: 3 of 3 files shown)


src/java.base/share/classes/jdk/crac/impl/AbstractContextImpl.java
```diff
@@ -55,32 +59,44 @@ protected synchronized void register(R resource, P payload) {
...
-            .map(Map.Entry::getKey)
-            .collect(Collectors.toList());
-
-        CheckpointException exception = new CheckpointException();
-        for (Resource r : resources) {
-            if (DEBUG) {
-                System.err.println("jdk.crac beforeCheckpoint " + r.toString());
+        synchronized (enterCriticalSectionLock) {
+            synchronized (exitCriticalSectionLock) {
+                while (!lockQ.isEmpty()) {
```
AntonKozlov, line 64: This use of `lockQ` is not synchronized. I assume there is no race with add-to/remove-from lockQ, since they are synced with enterCriticalSectionLock/exitCriticalSectionLock as well. But a comment will be useful.

Is it possible to remove enterCriticalSectionLock, exitCriticalSectionLock to make code more simple? And use a single lock to sync the list, to block entering ot checkpoint-critical section, and to notify the checkpointRestore the last section is finished. Or is it so to prioritize checkpoint over checkpoint-critical sections?

AntonKozlov (commented):
Thanks! It may be useful to split changes for CheckpointLock and in SecureRandom to make discussion more focused.

alexeybakhtin:
Hi @AntonKozlov ,
Thank you for your review.
As suggested, I've excluded SecureRandom part from the review and focused on the CRaC Critical Section API only.
AbstractContextImpl and SecureRandom classes are reverted to the original state.
JTREG Test is added to demonstrate and verify new API.

src/java.base/share/classes/jdk/crac/impl/AbstractContextImpl.java
```diff
@@ -55,32 +59,44 @@ protected synchronized void register(R resource, P payload) {
...
-        restoreQ = resources;
+            Collections.reverse(resources);
+            restoreQ = resources;
 
-        if (0 < exception.getSuppressed().length) {
-            throw exception;
+            if (0 < exception.getSuppressed().length) {
+                throw exception;
+            }
         }
```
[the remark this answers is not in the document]
alexeybakhtin: Yes, you are right. I've moved enterCriticalSectionLock to the Core class to lock checkpoint and restore stages.
afterRestore() methods are not protected by enterCriticalSectionLock, so it is possible to use methods with critical sections inside afterRestore()

src/java.base/share/classes/jdk/crac/impl/AbstractContextImpl.java
```diff
@@ -44,6 +45,9 @@ public Boolean run() {
     private WeakHashMap<R, P> checkpointQ = new WeakHashMap<>();
     private List<R> restoreQ = null;
     private Comparator<Map.Entry<R, P>> comparator;
+    private List<CheckpointLock> lockQ = new LinkedList<CheckpointLock>();
```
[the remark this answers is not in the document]
alexeybakhtin: Moved to Core class (see comments above)

src/java.base/share/classes/jdk/crac/impl/AbstractContextImpl.java
alexeybakhtin: Thank you. You are right. It could be simplified. It is not possible to use only one lock but I've merged enterCriticalSectionLock and exitCriticalSectionLock locks into one criticalSectionsLock

test/jdk/jdk/crac/Selector/checkpointLock/Test.java
```diff
@@ -0,0 +1,103 @@
...
+    private static class TestThread extends Thread {
+        private long timeout;
+
+        TestThread(long timeout) {
+            System.out.println("TestThread " + timeout);
+            this.timeout = timeout;
+        }
+
+        @Override
+        @SuppressWarnings("try")
```
AntonKozlov, line 44: Sorry, I missed this initially. What warning does the "try" generate? I.e. will every user will need this?

test/jdk/jdk/crac/Selector/checkpointLock/Test.java
```diff
@@ -0,0 +1,103 @@
+// Copyright 2019-2021 Azul Systems, Inc.  All Rights Reserved.
```
AntonKozlov, line 1: test/jdk/jdk/crac/checkpointLock seems more appropriate (Selector is a set of regression tests for nio.Selector Resource implementation)

test/jdk/jdk/crac/Selector/checkpointLock/Test.java
alexeybakhtin: OpenJDK build generates warning : "warning: [try] auto-closeable resource lock is never referenced in body of corresponding try statement" 
I never see this warning compiling tests or apps. It safe to remove from test

test/jdk/jdk/crac/Selector/checkpointLock/Test.java
alexeybakhtin: OK

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -50,6 +52,9 @@
 
     private static boolean traceStartupTime;
 
+    private static final List<CheckpointLock> lockQ = new LinkedList<CheckpointLock>();
```
DanHeidinga, line 55: LinkedList is rarely the class we actually want to use.  Better to use `ArrayList`

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -206,4 +227,20 @@ private static void checkpointRestoreInternal() {
...
+            synchronized (lockQ) {
+                lockQ.add(lock);
+            }
+        }
+    }
+
+    static void removeLock(CheckpointLock lock) {
+        synchronized (lockQ) {
+            lockQ.remove(lock);
+            lockQ.notify();
```
DanHeidinga, line 242: This will wake up the `checkpointRestore1` wait loop on every lock removal.  If there are a lot of CheckpointLock's this code will spend a lot of time waking up the other loop only for it to check and go back to sleep.

Would it make more sense to only `notify` the `lockQ` when the queue is empty?

DanHeidinga (commented):
I'd like to understand the use of this api a little better.  Do you have some examples of where and how you expect it to be used?

The CheckpointLock seems to be a way to block checkpoints from being taken during operations that are "unsafe" for checkpoints to occur during or would leave the application in an unstable state.

The idea of blocking checkpoints at certain points makes sense but I'm wondering how common do you expect this to be used?  Will library developers need to deploy this throughout their libraries?  Or does it target a smaller set of use cases?

How does this work in a non-CRIU runtime?  We don't want to introduce an API that library authors need to use for checkpoints but costs them performance / complexity when running on a vanilla JVM.  Maybe to gain some experience with the problem but it'll be hard to get buy-in for a solution that costs the common case for a corner case.

src/java.base/share/classes/jdk/crac/CheckpointLock.java
```diff
@@ -0,0 +1,45 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package jdk.crac;
+
+/**
+ * A {@code CheckpointLock} allows creating critical sections that are not interrupted during checkpoint/restore.
+ */
+public class CheckpointLock implements AutoCloseable {
```
DanHeidinga, line 32: I'm tempted to call this a `CheckpointBlocker` rather than a Lock as it seems the purpose is to prevent checkpoints from being able to proceed.

src/java.base/share/classes/jdk/crac/Core.java
alexeybakhtin: Good catch. Thank you. 
Even if notify does not really wakeup another thread, it would be better to notify on empty queue only

src/java.base/share/classes/jdk/crac/Core.java
alexeybakhtin: Thank you. After comparing ArrayList and LinkedList implementations, I can see - ArrayList has better performance even for general LinkedList use cases.

alexeybakhtin:
@DanHeidinga 
The main idea of CheckpointLock is to introduce Critical Sections that can not be interrupted by checkpoint/restore operation. It could be used by JDK and Application code. One of the use cases was introduced in the first commit to this review as an example:
https://github.com/openjdk/crac/blob/a3ea1f6eadb772e70085cbba4d23a80ba6dca9e2/src/java.base/share/classes/sun/security/provider/SecureRandom.java#L166
Sha1PRNG Secure random should be reseeded to generate a new set of random bytes after restore. Also, the state should be cleared to prevent the calculation of the previously used randoms. As soon as the state is cleared, an implementation should prevent checkpoint in the middle of the methods that use a secure random state. Similar actions could/should be applied for other secure sensitive resources.
It could be necessary for application and libraries to use critical sections if they modify resources in the beforeCheckpoint/afterRestore methods.
The overhead of the CheckpointLock is very limited - just creation of the small object, add to and remove from the static ArrayList - so, I do not expect performance issues for the non-CRIU runtime or normal behavior after restore.

DanHeidinga:
> @DanHeidinga The main idea of CheckpointLock is to introduce Critical Sections that can not be interrupted by checkpoint/restore operation. It could be used by JDK and Application code. 

I understand why this is an appealing idea - there are operations that ideally should be treated atomically with respect to the checkpoint/restore mechanism.  Conceptually very similar to Hotspot's GCLocker blocking further access to jni critical while a GC is pending.

And while this kind of approach works for ie: jni critical, it works in part because the spec requires that critical regions are well bounded and the operations allowed in a critical section are heavily restricted.

Those properties don't hold for CheckpointLocks.  What properties need to be specified to ensure these critical sections are short lived?  

> One of the use cases was introduced in the first commit to this review as an example:
> 
> https://github.com/openjdk/crac/blob/a3ea1f6eadb772e70085cbba4d23a80ba6dca9e2/src/java.base/share/classes/sun/security/provider/SecureRandom.java#L166
> 
> 
> Sha1PRNG Secure random should be reseeded to generate a new set of random bytes after restore. Also, the state should be cleared to prevent the calculation of the previously used randoms. As soon as the state is cleared, an implementation should prevent checkpoint in the middle of the methods that use a secure random state. Similar actions could/should be applied for other secure sensitive resources.

Thanks for the example.  Looking at the SecureRandom code, the `afterRestore` hook calls the `engineSetSeed` method which takes a `CheckpointLock` to prevent additional checkpoints happening during that operation.

There's a larger problem here though, that isn't resolved by stopping the future checkpoints from happening during the `engineSetSeed` and looking at that problem will hopefully lead to better understanding of thread behaviour around checkpoints.

The intention of updating the SecureRandom seed is to prevent attackers from being able "guess" future random numbers by restoring, observing the randoms, and then restoring a second time - basically to avoid replay attacks.

Using CheckpointLock prevents a second checkpoint from happening while the seeds are being reset, but I'm not sure what that actually gains us here.  Presumably, the second checkpoint will run through the `{before,after}Checkpoint` hooks and reset the state on checkpoint and reseed on the second restore.  What additional advantage has been gained by locking around that operation?

And to the larger threading question, if we have N threads operating prior to the checkpoint using the SecureRandom, by taking the checkpoint, we leave those threads with potentially inconsistent results.  The checkpoint operation doesn't quiesce the threads so they are free to observe the state being modified by the `beforeCheckpoint` hook and on restore, to race with `afterCheckpoint` hook potentially seeing the old or new seed.

If I've missed something that forces the running threads to quiesce, then please correct me.

> It could be necessary for application and libraries to use critical sections if they modify resources in the beforeCheckpoint/afterRestore methods.

I don't think CheckpointLock provides a strong enough guarantee for these operations as they race with other active threads.

> The overhead of the CheckpointLock is very limited - just creation of the small object, add to and remove from the static ArrayList - so, I do not expect performance issues for the non-CRIU runtime or normal behavior after restore.

Look at this like the use of the AccessController::doPrivileged operations which are hard to place correctly, and for most users, are pure overhead as they don't run with the SecurityManager enabled.  I'm hesitant about introducing APIs that have the same characteristics as apis that are in the process of being removed [1] unless we have good understanding of why the new apis are different.

[1] https://github.com/openjdk/jdk/blob/f7237793ffa3a5a804fea49f165c8b9f1935cfac/src/java.base/share/classes/java/security/AccessController.java#L281

alexeybakhtin:
Hi @DanHeidinga,

Thank you a lot for your review. I see a lot of open issues. Let me answer to some of them. The others are homework for me. 
> Using CheckpointLock prevents a second checkpoint from happening while the seeds are being reset, but I'm not sure what that actually gains us here. Presumably, the second checkpoint will run through the `{before,after}Checkpoint` hooks and reset the state on checkpoint and reseed on the second restore. What additional advantage has been gained by locking around that operation?

I think we should restrict nested checkpoints. So, application should not be allowed to request second checkpoint from the beforeCheckpoint or afterRestore methods. Proposed implementation partially implement it with checkpointRestoreLock
https://github.com/openjdk/crac/blob/db6f36e9f67110f750d4e4361f0c8fabdbe8d7e9/src/java.base/share/classes/jdk/crac/Core.java#L199
but it does not work if called directly (not from threads) from beforeCheckpoint / afterRestore methods.
Also, I think CheckpointException should be thrown on the attempt to a nested checkpoint. 

> And to the larger threading question, if we have N threads operating prior to the checkpoint using the SecureRandom, by taking the checkpoint, we leave those threads with potentially inconsistent results. The checkpoint operation doesn't quiesce the threads so they are free to observe the state being modified by the `beforeCheckpoint` hook and on restore, to race with `afterCheckpoint` hook potentially seeing the old or new seed.
> 
> If I've missed something that forces the running threads to quiesce, then please correct me.
> 

I think this exact scenario is addressed here. The checkpoint operation holds the lock on the checkpointRestoreLock object. Running threads will be stopped till the end of restore trying to walk into the critical section 
https://github.com/openjdk/crac/blob/db6f36e9f67110f750d4e4361f0c8fabdbe8d7e9/src/java.base/share/classes/jdk/crac/Core.java#L231
It is the application responsibility to correctly surround all code sections that access resources modified in the beforeCheckpoint method. 
Threads, suspended on the enter of critical section will be released and continue execution after Restore is complete and state is modified in beforeCheckpoint/afterRestore methods. These threads will enter a critical section with the new state of the resource

> Look at this like the use of the AccessController::doPrivileged operations which are hard to place correctly, and for most users, are pure overhead as they don't run with the SecurityManager enabled. I'm hesitant about introducing APIs that have the same characteristics as apis that are in the process of being removed [1] unless we have good understanding of why the new apis are different.
> 
Yes, I see your point. I think I can modify the code to do nothing in case of application started without checkpoint properties CRaCChecpointTo and CRaCRestoreFrom.

AntonKozlov:
> The idea of blocking checkpoints at certain points makes sense but I'm wondering how common do you expect this to be used? Will library developers need to deploy this throughout their libraries? Or does it target a smaller set of use cases?

Libraries may find this useful, but this is not mandatory, of course. A similar mechanism can be implemented by an application.
But since the JDK is will likely have this implemented, and some libraries may need this, it's worth exposing this as a public API. The main driver is the need to provide a control for something that should not end up in the image, like secrets that are fine to flow in the memory of the running process but are too critical to be stored. This relates to any implementation (e.g. not based on CRIU) that stores the image of the JVM process.

> And while this kind of approach works for ie: jni critical, it works in part because the spec requires that critical regions are well bounded and the operations allowed in a critical section are heavily restricted.
> Those properties don't hold for CheckpointLocks. What properties need to be specified to ensure these critical sections are short lived?

In contrast with JNI critical (which affects negatively the whole JVM), a non-minimal critical section has a potentially negative effect on the checkpoint itself only. Making the section shorter is beneficial, but not mandatory. As with custom beforeCheckpoint's, critical sections are part of the application's checkpoint procedure. JDK itself should not impose restrictions on this -- the user should have the control and take responsibility.

> Look at this like the use of the AccessController::doPrivileged operations which are hard to place correctly, and for most users, are pure overhead as they don't run with the SecurityManager enabled. I'm hesitant about introducing APIs that have the same characteristics as apis that are in the process of being removed [1] unless we have good understanding of why the new apis are different.

This is indeed an interesting analogy. Actually, the API similar to the doPrivileged looks even better :) (instead of try-with-resources that generates the warning).

From the critique of the SecurityManager from JEP-411 [1], the difficult programming model and poor performance concerns should be evaluated. But both are associated with the need of traversing the stack for making the decision.
For CheckpointLock we need to check the global state that may be implemented efficiently, especially for no-checkpoint-possible mode.

[1] https://openjdk.java.net/jeps/411

DanHeidinga:
Lock ordering between `synchronized` methods and the `jdk.crac.Core.checkpointRestoreLock` can also result in deadlocks.

There's a non-obvious deadlock in this example (assuming I have the ordering right) between the callers of the `CheckpointDeadLock.operate()` method and the act of taking a checkpoint:

```
class CheckpointDeadLock {
   Object state;
   
  synchronized void clearState() {
    state = null;
  }

  synchronized void setState(Object o) {
    try (new CheckpointLock()) {
       // lock order: {this, jdk.crac.Core.checkpointRestoreLock }
       state = o;
    }
  }
   
   void operate {
      while(true) {
        setState(new Object());
      }
    }

    public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
       // { jdk.crac.Core.checkpointRestoreLock, this } when calling synchronized 'clearState'
        clearState();
    }
}
```

AntonKozlov:
> It's also non-obvious how the two different `new CheckpointLock()` calls interact with each other and the class's existing `synchronized` methods. A developer needs to think about two kinds of synchronization now:
> 
> 1. The existing concept of synchronization to protect against concurrent access
> 2. The new checkpoint consistency to guard against checkpoint-related data modification.
>

I suggest looking at the checkpoint-critical section (or similar) as the Read part of the ReadWrite [1] lock, while checkpoint/restore is performed with the Write lock acquired. Checkpoint-critical sections can intersect, they are effectively no-op in absence of checkpoint/restore. Checkpoint/restores never intersect with each other and checkpoint-critical sections.

Similarly, with the Lock [2], I think we can specify checkpoint-lock independent of any native monitor. However, this implementation needs to have a minimal effect on the checkpoint-critical sections and Resource implementations, and the effect should be documented.

[1] https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReadWriteLock.html
[2] https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html 

> All of a class invariants now have to be guarded against changes from the `{before/after}Checkpoint` methods in addition to multi-threaded concerns.

They should have such guards before this PR as well -- the async checkpoint request may arrive any time and the implementor should take care about proper sync of between the main thread(s) and the thread doing checkpoint/restore (and calling before/afterRestore's, the thread is not specified).

So the new API should only sync arbitrary java code with checkpoint/restore; but not arbitrary java code with before/afterRestore -- for this, there is a better way with fine-grained user synchronization.

Thanks for the deadlock example. That and the occasional dependency on the lock implementation pushes to the thinking: should not we narrow checkpointRestoreLock to checkpointRestore1. So beforeCheckpoint will run in parallel with checkpoint-critical sections. 

> The overall thing - and in general, not just in this PR - that needs to be clarified is the threading and locking behaviour around a checkpoint. What guarantees does CRaC provide to application threads about the state of the system during their `beforeCheckpoint` hooks and their `afterRestore` hooks?
> 
> Are application / framework / library authors responsible to ensure their threads are quiesced?

What kind of guarantees would we need?  So far we don't assume any special state of the system and threads during beforeCheckpoint/afterRestore, as they are arbitrary java code. 

In general, it's app / lib responsibility to sync user resources with the checkpoint.
I doubt jdk would implement this flexibly and efficiently, so the best is to provide the control for users.

> If we intend to have CRaC optimize the checkpoint by e.g. aggressively GCing before a checkpoint, then it would be good to ensure that app threads aren't allocating garbage at the same time. Acting in a single threaded manner here is a great conceptual model but introduces other issues e.g. deadlocks - at any rate, we need to start to specify what we're going to guarantee and what the apps have to build themselves.

For this, checkpoint implemented in several phases:
1. beforeCheckpoint's are called
2. in the safepoint, JVM checks preconditions and do necessary preparation, including GC
3. still in the safepoint, the image is stored

> I'm concerned about how easy it will be to get this wrong with the proposed api.
> 
> And common synchronization patterns, like privatizing copies to operate on, will violate the intent of the CheckpointLock api rendering user's existing intuition wrong around how to protect data.

Intuition plays against us. Process memory is usually seen as vault protected by OS means. However, it may leak on the file system as a core dump. The same is with CRaC, although the image is a part of the expected workflow. With the new API, the image still could not be completely safe from secrets unless a complete code review is performed. But at least it is feasible.

> The point I'm trying to make is that we need to uplevel the discussion to look at the guarantees we want CRaC to provide and then can tune the apis and features we create to provide those guarantees.

Do you mean something in particular? I think we have a pile of concrete concerns about CRaC like the security, we don't have good mechanisms to deal with. Adding new features and trying to use them is the way to notice similarities and to come with something better (the API is not stable yet and may be extended and reduced). However, knowing the flaws in advance are preferable. Just not sure what are they.

alexeybakhtin:
> Thanks for the deadlock example. That and the occasional dependency on the lock implementation pushes to the thinking: should not we narrow checkpointRestoreLock to checkpointRestore1. So beforeCheckpoint will run in parallel with checkpoint-critical sections.

I think we can not do it this way. It is exactly what we are going to prevent. beforeCheckpoint() prepares resources for checkpoint and the critical section prevents resources from concurrent access/modification during checkpoint/restore.
In case of a concurrent run of beforeCheckpoint() and critical sections, we'll have a situation when beforeCheckpoint() prepares resources for checkpoint and then critical sections modify or restore it. So, I think the lock around the whole checkpoint/ restore is the right solution.

alexeybakhtin:
Hi @AntonKozlov @DanHeidinga
Thank you for your review and comments.
I've proposed the new API method Core.criticalSection() instead of CheckpointLock.
CheckpointLock causes javac warning and the new proposed API method does not have this issue.
Also, this version optimizes criticalSection implementation in case CRaCCheckpointTo option is not specified in cmdline.
jmh benchmark is added to assure - critical section does not cause performance degradation for normal and after restore JDK runs

alexeybakhtin:
Hi @DanHeidinga, all
After additional discussions, we decided to postpone this API. As Dan pointed out, additional checkpoint locks introduce excessive complexity of the code and could lead to deadlocks that are hard to find and avoid.
In the absence of a general implementation of checkpoint locks, it is the application's responsibility to protect its own resources from concurrent access/modification during checkpont/restore.
I will continue to implement protection of security-related data in another PR.

PR #7, opened 2021-12-17
Reseed secure random on checkpoint restore

alexeybakhtin:
Proposed changes in the SecureRandom implementation allow invalidating and reseeding SHA1PRNG secure random during checkpoint/restore. SHA1PRNG can be invalidated and reseeded in case of being created with a default embedded seed generator. Also, SHA1PRNG is used as an additional seed generator to the SUN NativePRNG implementation, so it is desirable to have reseeded SHA1PRNG after restore.
Two jtreg tests added: 
- verify if no deadlocks introduced by checkpoint/restore
- verify if SHA1PRNG is reseeded if created with default embedded seed generator

---------
### Progress

### Reviewers
 * [Anton Kozlov](https://openjdk.java.net/census#akozlov) (@AntonKozlov - **Reviewer**) ⚠️ Review applies to f33c6ab0470c439b2e229a4e32505157769f22ed

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/crac pull/7/head:pull/7` \
`$ git checkout pull/7`

`$ git checkout pull/7` \
`$ git pull https://git.openjdk.java.net/crac pull/7/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 7`

`$ git pr show -t 7`

</details>

<a href="https://git.openjdk.java.net/crac/pull/7.diff">https://git.openjdk.java.net/crac/pull/7.diff</a>

</details>


src/java.base/share/classes/sun/security/provider/SecureRandom.java
```diff
@@ -190,6 +204,46 @@ private static void updateState(byte[] state, byte[] output) {
...
+        } finally {
+            objLock.unlock();
+        }
+    }
+
+    @Override
+    public int getPriority() {
+        if (isSeedGenerator)
+            return -12;
+        return -11;
```
DanHeidinga, line 244: How are priorities determined?  I see other implementors of this method mostly setting their priority to 0.

src/java.base/share/classes/sun/security/provider/SecureRandom.java
```diff
@@ -154,15 +162,21 @@ private void init(byte[] seed) {
...
+        try {
+            if (state != null) {
+                digest.update(state);
+                for (int i = 0; i < state.length; i++) {
+                    state[i] = 0;
+                }
             }
+            state = digest.digest(seed);
+            remCount = 0;
+            clearStateOnCheckpoint = false;
```
DanHeidinga, line 176: I'm not clear on why we need this boolean.  Why does the code say not to clear this Random on a checkpoint if someone has called `engineSetSeed`?

src/java.base/share/classes/sun/security/provider/SecureRandom.java
alexeybakhtin: Hi @DanHeidinga 
Default global CRaC Context implementation uses resource priorities to define the order of checkpoint notifications (beforeCheckpoint). Resource priorities are automatically assigned incrementally upon resource registration in the context. The lowest priority is 0. Order of checkpoint notification is the reverse order of registration.
There are JDKContext and JDKResource internal implementations. These classes allow assigning priorities to specific resources. The order of checkpoint notifications for JDKResources is from higher to lower priorities.

src/java.base/share/classes/sun/security/provider/SecureRandom.java
alexeybakhtin: By default, SHA1PRNG implementation depends on the initial seed and is not reseeded internally. The random sequence is a function of initial seed. So, an application could expect a particular sequence on depends on the seed provided. This is a reason we can not reseed secure random if the application provides its own seed. In case of application does not provide seed, the seed is auto-assigned by implementation. It means the application is not depends on particular random sequence. It means we can re-seed such secure randoms during checkpoint/restore.
Secure Randoms, seeded by the application should be handled by this application layer because of only application knows How to handle it.
So, we'll re-seed secure randoms only in case of we sure it does not change application logic

src/java.base/share/classes/sun/security/provider/SecureRandom.java
DanHeidinga: Given that explanation, how were `-12` & `-11` chosen here?  Randomly as low enough to ensure these are changed late?

src/java.base/share/classes/sun/security/provider/SecureRandom.java
DanHeidinga: That makes a lot of sense - thanks for clarifying!

src/java.base/share/classes/sun/security/provider/SecureRandom.java
alexeybakhtin: Yes and No :) I left some space for the future JDK resources that depend on the secure randoms (use secure random inside own beforeCheckpoint method)
Maybe these priorities should be collected in a single class, e.g. defined in the JDKresource interface

alexeybakhtin: Added pull request to collect JDKResorce priorities in the single enumeration: https://github.com/openjdk/crac/pull/8
Current PR will be updated as soon as JDKResource priorities is accepted and comitted

src/java.base/share/classes/sun/security/provider/SecureRandom.java
```diff
@@ -218,53 +274,56 @@ private static void updateState(byte[] state, byte[] output) {
...
-        int todo;
-        byte[] output = remainder;
-
-        if (state == null) {
-            byte[] seed = new byte[DIGEST_SIZE];
-            SeederHolder.seeder.engineNextBytes(seed);
-            state = digest.digest(seed);
-        }
+    public void engineNextBytes(byte[] result) {
+        objLock.lock();
```
AntonKozlov, line 278: A minor drawback, the keyword in the method declaration makes the synchronization more explicit. An alternative is to retain the synchronization the on object monitor by e.g.

```
private boolean blocked = false;

public synchronized void engineNextBytes {
   while (blocked) {
     this.wait()
   }
   ...
}

public synchronized void beforeCheckpoint {
  blocked = true
  ...
}
public synchronized void afterRestore {
  ...
  blocked = false;
  this.notifyAll();
}
```

However, a dedicated ReentrantLock makes the implementation a bit shorter. So, up to you.

AntonKozlov (changes requested):

src/java.base/share/classes/sun/security/provider/SecureRandom.java
```diff
@@ -206,9 +260,11 @@ private static void updateState(byte[] state, byte[] output) {
              * seed material (likely from the Native implementation).
              */
             seeder = new SecureRandom(SeedGenerator.getSystemEntropy());
-            byte [] b = new byte[DIGEST_SIZE];
+            seeder.isSeedGenerator = true;
+            byte[] b = new byte[DIGEST_SIZE];
             SeedGenerator.generateSeed(b);
             seeder.engineSetSeed(b);
+            seeder.clearStateOnCheckpoint = true;
```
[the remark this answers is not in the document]
alexeybakhtin: Unfortunately sun.security.provider.SecureRandom class is final. I think it was done intentionally for security reasons.
isSeedGenerator can be easily replaced by comparison with SeederHolder.seeder but it'll eat time during restore.

src/java.base/share/classes/sun/security/provider/SecureRandom.java
alexeybakhtin: Yes, it makes sense. But using Lock object in the beforeCheckpoint() -> afterRestore() methods looks more natural.
Also, in your case, we'll have to check for interruptedException and modify engineSetSeed() accordingly.
If there are no more objections, I would use objLock for synchronization.

src/java.base/share/classes/sun/security/provider/SecureRandom.java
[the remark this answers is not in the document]
AntonKozlov: Another option is to make `SeederHolder` to be the `Resource` that cleans `seeder` on checkpoint and creates one on restore. I think `seeder.clearStateOnCheckpoint = true` won't be needed, as the next checkpoint will clear-up the seeder again.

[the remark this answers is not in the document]
alexeybakhtin: > Another option is to make `SeederHolder` to be the `Resource` that cleans `seeder` on checkpoint and creates one on restore. I think `seeder.clearStateOnCheckpoint = true` won't be needed, as the next checkpoint will clear-up the seeder again.

You are right. Thank you. Reimplemented as you suggested

src/java.base/share/classes/sun/security/provider/SecureRandom.java
```diff
@@ -190,25 +204,78 @@ private static void updateState(byte[] state, byte[] output) {
...
             /*
              * Call to SeedGenerator.generateSeed() to add additional
              * seed material (likely from the Native implementation).
              */
             seeder = new SecureRandom(SeedGenerator.getSystemEntropy());
-            byte [] b = new byte[DIGEST_SIZE];
+            byte[] b = new byte[DIGEST_SIZE];
             SeedGenerator.generateSeed(b);
             seeder.engineSetSeed(b);
+            jdk.internal.crac.Core.getJDKContext().register(new SeederHolder());
```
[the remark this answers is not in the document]
alexeybakhtin: Thank you. Fixed in new version

src/java.base/share/classes/sun/security/provider/SecureRandom.java
```diff
@@ -190,25 +213,77 @@ private static void updateState(byte[] state, byte[] output) {
         }
     }
 
+    private void invalidate() {
```
AntonKozlov, line 216: I would like to have "assert objLock.isHeldByCurrentThread()" or another "objLock.lock()" here. This method is called from SeederHolder.beforeCheckpoint. There is no race here only if that executes after seeder's beforeCheckpoint and in the same thread. While it is so, additional safety would not harm.

src/java.base/share/classes/sun/security/provider/SecureRandom.java
```diff
@@ -190,25 +204,78 @@ private static void updateState(byte[] state, byte[] output) {
...
-            byte [] b = new byte[DIGEST_SIZE];
+            byte[] b = new byte[DIGEST_SIZE];
             SeedGenerator.generateSeed(b);
             seeder.engineSetSeed(b);
+            jdk.internal.crac.Core.getJDKContext().register(new SeederHolder());
+        }
+
+        @Override
+        public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+            objLock.lock();
```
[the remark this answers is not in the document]
AntonKozlov: The CheckpointException does not look like a perfect exception here. To avoid controversial requirements, could it be RuntimeException or another unchecked exception, like e.g. IllegalStateException?

src/java.base/share/classes/jdk/crac/CheckpointException.java
```diff
@@ -29,7 +29,7 @@
 /**
  * Suppresses exceptions thrown during checkpoint notification.
  */
-public class CheckpointException extends Exception {
+public class CheckpointException extends RuntimeException {
```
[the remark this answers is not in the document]
alexeybakhtin: This is an internal jdk.crac Exception that is not visible to users. It is used for JVM resources only and handled explicitly in jdk.crac.Core. This exception will be thrown from existing JDK classes, so it was changed to unchecked to make it possible to throw exception without changing signature of the existing public API (e.g. https://github.com/openjdk/crac/blob/3ade9580452ab2db193e9c2c1b458a2ff17a8597/src/java.base/share/classes/sun/security/provider/SecureRandom.java#L296 or https://github.com/openjdk/crac/blob/dd46160142a3ec490a400f56738d0251d128494a/src/java.base/unix/classes/sun/security/provider/NativePRNG.java#L556 or https://github.com/openjdk/crac/blob/dd46160142a3ec490a400f56738d0251d128494a/src/java.base/unix/classes/sun/security/provider/NativePRNG.java#L490)
Also, it was approved already in the https://github.com/openjdk/crac/pull/9

src/java.base/share/classes/sun/security/provider/SecureRandom.java
```diff
@@ -154,7 +163,20 @@ private void init(byte[] seed) {
      * @param seed the seed.
      */
     @Override
-    public synchronized void engineSetSeed(byte[] seed) {
+    public void engineSetSeed(byte[] seed) {
+        objLock.lock();
+        try {
+            // verify if objLock is already acquired in beforeCheckpoint
```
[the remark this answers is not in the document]
alexeybakhtin: Thank you, changed.

src/java.base/share/classes/sun/security/provider/SecureRandom.java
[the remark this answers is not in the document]
alexeybakhtin: CheckpointException is CheckpointException now.
We can not use IllegalStateException because we should catch this exception and correctly roll back the checkpoint.
Also, see my comments in CheckpointException https://github.com/openjdk/crac/pull/7#discussion_r793379918

src/java.base/share/classes/jdk/crac/CheckpointException.java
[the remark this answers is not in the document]
alexeybakhtin: After additional discussions, it was decided to revert these changes and use checked CheckpointException.
JDKResources can throw another RuntimeExceptions like IllegalStateException if required to indicate a failure during checkpoint/restore. All exceptions are processed in the AbstractContextImpl.beforeCheckpoint(), restore resources and throw ChecpointException to the user with the real cause.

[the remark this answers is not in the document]
AntonKozlov: Thanks for fixing this and keeping #9 consistent

AntonKozlov (approved):
Looks good to me!

PR #8, opened 2021-12-20
JDKResource priorities

alexeybakhtin:
Added priority enumeration for the JDK resources
It will allow better handling the order of checkpoint notifications for dependent resources

---------
### Progress

### Reviewers
 * [Anton Kozlov](https://openjdk.java.net/census#akozlov) (@AntonKozlov - **Reviewer**) ⚠️ Review applies to 58f23f6739f050f46bc031843355451bf5d624ef
 * [Dan Heidinga](https://openjdk.java.net/census#heidinga) (@DanHeidinga - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/crac pull/8/head:pull/8` \
`$ git checkout pull/8`

`$ git checkout pull/8` \
`$ git pull https://git.openjdk.java.net/crac pull/8/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 8`

`$ git pr show -t 8`

</details>

<a href="https://git.openjdk.java.net/crac/pull/8.diff">https://git.openjdk.java.net/crac/pull/8.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/crac/JDKContext.java
+++ b/src/java.base/share/classes/jdk/internal/crac/JDKContext.java
@@ -35,7 +35,7 @@ public class JDKContext extends AbstractContextImpl<JDKResource, Void> {
     static class ContextComparator implements Comparator<Map.Entry<JDKResource, Void>> {
         @Override
         public int compare(Map.Entry<JDKResource, Void> o1, Map.Entry<JDKResource, Void> o2) {
-            return o1.getKey().getPriority().ordinal() - o2.getKey().getPriority().ordinal();
+            return o1.getKey().getPriority().compareTo(o2.getKey().getPriority());
         }
     }
 
--- a/src/java.base/share/classes/jdk/internal/crac/JDKContext.java
+++ b/src/java.base/share/classes/jdk/internal/crac/JDKContext.java
@@ -35,7 +35,7 @@ public class JDKContext extends AbstractContextImpl<JDKResource, Void> {
     static class ContextComparator implements Comparator<Map.Entry<JDKResource, Void>> {
         @Override
         public int compare(Map.Entry<JDKResource, Void> o1, Map.Entry<JDKResource, Void> o2) {
-            return o2.getKey().getPriority().ordinal() - o1.getKey().getPriority().ordinal();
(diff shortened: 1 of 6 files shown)


src/java.base/share/classes/jdk/internal/crac/JDKResource.java
```diff
@@ -29,5 +29,23 @@
...
+     * Most resources should use NORMAL.
+     * Other priorities define sequence of
+     * checkpoint notification for dependent resources
+     */
+    enum Priority {
+        /**
+         * Priority of the
+         * sun.nio.ch.EPollSelectorImpl resource
+         */
+        EPOLLSELECTOR,
```
AntonKozlov, line 43: This uncovers the future mess we'll have with manually setting the order, similar to the mutex order in the Hotspot. However, I like this central place to control them. This approach is by no means worse than the existing code but highlights a point of growth.

AntonKozlov (approved):

src/java.base/share/classes/jdk/internal/crac/JDKContext.java
```diff
@@ -35,7 +35,7 @@
     static class ContextComparator implements Comparator<Map.Entry<JDKResource, Void>> {
         @Override
         public int compare(Map.Entry<JDKResource, Void> o1, Map.Entry<JDKResource, Void> o2) {
-            return o2.getKey().getPriority() - o1.getKey().getPriority();
+            return o1.getKey().getPriority().ordinal() - o2.getKey().getPriority().ordinal();
```
DanHeidinga, line 38: Enums are already comparable so we can use the built in compareTo method here but this isn't critical and shouldn't block merging

```suggestion
            return o1.getKey().getPriority().compareTo(o2.getKey().getPriority());
```

DanHeidinga (approved):

src/java.base/share/classes/jdk/internal/crac/JDKContext.java
alexeybakhtin: Hi @DanHeidinga !
Thank you for review. Yes, compareTo looks better. Updated.

DanHeidinga (approved):
lgtm

PR #9, opened 2021-12-23
Reseed NativePRNG on checkpoint restore

alexeybakhtin:
NativePRNG should be re-seeded during checkpoint/restore because it uses SHA1PRNG secure random for additional seed. It is seeded at initialization, so it is not re-seeded automatically during checkpoint/restore
Also, the internal buffer should be cleared at the checkpoint.

---------
### Progress

### Reviewers
 * [Anton Kozlov](https://openjdk.java.net/census#akozlov) (@AntonKozlov - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/crac pull/9/head:pull/9` \
`$ git checkout pull/9`

`$ git checkout pull/9` \
`$ git pull https://git.openjdk.java.net/crac pull/9/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 9`

`$ git pr show -t 9`

</details>

<a href="https://git.openjdk.java.net/crac/pull/9.diff">https://git.openjdk.java.net/crac/pull/9.diff</a>

</details>


src/java.base/unix/classes/sun/security/provider/NativePRNG.java
```diff
@@ -566,7 +579,31 @@ private void implNextBytes(byte[] data) {
...
+
+        @Override
+        public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+            crLock.lock();
+            mixRandom = null;
+            buffered = 0;
+            lastRead = 0;
+            for(int i=0; i<nextBuffer.length; i++) {
+                nextBuffer[i] = 0;
+            }
```
mbien, line 595: could use Arrays.fill

alexeybakhtin: Yes. Thank you. I will update

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
```diff
@@ -566,7 +579,31 @@ private void implNextBytes(byte[] data) {
...
                 } catch (IOException e){
                     throw new ProviderException("nextBytes() failed", e);
+                } finally {
+                    crLock.unlock();
                 }
         }
+
+        @Override
+        public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+            crLock.lock();
```
AntonKozlov, line 589: The new lock is very related to `LOCK_GET_BYTES`, however, this code does not acquire `LOCK_GET_BYTES`. Now all `synchronized (LOCK_GET_BYTES)` blocks are executed under `crLock`. But I would prefer at least `assert crLock.isHeldByCurrentThread()` near those synchronized blocks, or any other way to ensure / document locks relation. Or use the LOCK_GET_BYTES, considering the other problem with premature buffer and mixRandom initialzation.

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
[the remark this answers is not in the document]
alexeybakhtin: For 2. I think it is just a matter of JDKResource.Priority. The use case you described is about dependent resources (see https://github.com/openjdk/crac/pull/8). In this case, the priority of the NativePRNG should be adjusted in the JDKResource.Priority.
Also, I would suggest adding a debug option that enables detection of the incorrect usage of the JDKResource A inside beforeCheckpoint of JDKResource B (priority A > priority B)

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
alexeybakhtin: I do not see any issues between `crLock` and `LOCK_GET_BYTES.` `beforeCheckpoint` does not acquire any additional locks, so it should not cause deadlocks.
It is normal if different threads try to acquire `crLock.` In this case, one of them waits for the completion of another. E.g. `beforeCheckpoint` waits for completion `implNextBytes()` and vice versa. So, `assert crLock.isHeldByCurrentThread()` is not required.
In case of `implNextBytes` called from the `beforeCheckpoint` and `crLock` already acquired by `RandomIO.beforeCheckpoint` will have improper priorities of dependent JDKResources. It should be properly adjusted in the JDKResource.Priority (see #8)

alexeybakhtin: I do not see any issues between `crLock` and `LOCK_GET_BYTES.` `beforeCheckpoint` does not acquire any additional locks, so it should not cause deadlocks.
It is normal if different threads try to acquire `crLock.` In this case, one of them waits for the completion of another. E.g. `beforeCheckpoint` waits for completion `implNextBytes()` and vice versa. So, assert `crLock.isHeldByCurrentThread()` is not required.
In case of `implNextBytes` called from the `beforeCheckpoint` and `crLock` already acquired by `RandomIO.beforeCheckpoint` will have improper priorities of dependent `JDKResources.` It should be properly adjusted in the `JDKResource.Priority` (see #8)
The only problem is performance. `INSTANCE` is a single object. So, locking the whole  `implNextBytes` method will affect performance dramatically. I think It can be fixed by `ReentrantReadWriteLock`. Will update implementation

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
```diff
@@ -383,6 +387,9 @@ protected void engineNextBytes(byte[] bytes) {
         // mutex lock for setSeed()
         private final Object LOCK_SET_SEED = new Object();
 
+        // lock for checkpoint/restore
+        private final ReentrantLock crLock = new ReentrantLock();
```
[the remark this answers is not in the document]
alexeybakhtin: Will do

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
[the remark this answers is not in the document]
AntonKozlov: Without the debug option, the image may be compromised -- so there should be no option of running without the checks. The currently employed ReentrantReadWriteLock does not block taking Read lock (access to PRNG) while holding Write lock (that is supposed to block access to the PRNG during checkpoint/restore -- but not from the same thread). It would be much more useful to throw an exception "the image is going to be wrong", or at worst to deadlock. But now the implementation does guarantee the security of the image.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/locks/ReentrantReadWriteLock.html

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
AntonKozlov: crLock and LOCK_GET_BYTES should be acquired in the particular order to ensure no deadlock can happen. The order is correct now AFAICS, but in the code these locks are used in distant places, checking the locks relation is nontrivial. Someone in the future may add another use of LOCK_GET_BYTES that would go out-of-sync with crLock.

I support using a ReadWriteLock. I'm not sure about how bad the performance was without one effect (the LOCK_GET_BYTES is there anyway). But some ReadWriteLock is certainly better suited for our task. I'm still not sure the ReentrantReadWriteLock is a perfect one, see another thread.

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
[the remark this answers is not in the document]
alexeybakhtin: Thank you. Added the code to verify if NativePRNG is used from the beforeCheckpoint/afterRestore. In such a case, CheckpointException will be thrown.

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
alexeybakhtin: I do not like to replace LOCK_GET_BYTES with crLock. It will affect performance because of NativePRNG is singleton and the original LOCK_GET_BYTES synchronizes much smaller pieces of code.
The only standard implementation of ReadWriteLock is ReentrantReadWriteLock. So I think it is OK to use ReentrantReadWriteLock with additional checks I just proposed

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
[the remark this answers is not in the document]
AntonKozlov: Thanks, now it looks good. Just as a note, in the similar case I've used IllegalSelectorException [1]. I think IllegalStateException may fit here as well -- we probably need to converge what exception(s) to use. But it is not necessary now.

[1] https://github.com/openjdk/crac/blob/43aba3d502832a5a3d2e9712558f62e0cf93dbbb/src/java.base/linux/classes/sun/nio/ch/EPollSelectorImpl.java#L384

src/java.base/unix/classes/sun/security/provider/NativePRNG.java
AntonKozlov: OK, I don't insist. Two related locks complicate the code, but crLock intent and the difference from LOCK_GET_BYTES is clearer now. I think the current state is good.

AntonKozlov (approved):

alexeybakhtin:
After additional discussions, it was decided to revert these changes and use checked CheckpointException.
JDKResources can throw another RuntimeExceptions like IllegalStateException if required to indicate a failure during checkpoint/restore. All exceptions are processed in the AbstractContextImpl.beforeCheckpoint(), restore resources and throw ChecpointException to the user with the real cause.

AntonKozlov (approved):

PR #3, opened 2021-10-28
Decouple JVM and CRIU

AntonKozlov:
CREngine option to specify a program to do checkpoint/restore.
* `criuengine` to use CRIU as it was used before.
* `pauseengine` that stores PID in a file and pauses JVM until `java -XX:CRaCRestoreFrom=` is called. This should allow external mechanisms that may operate on a process.
* `simengine` is something that simulates restore immediately, with the same effect as `-XX:+CRAllowToSkipCheckpoint`. It is intended to be simplest example and could be a starting point to implement Checkpoint/Restore with e.g. the help of a VM/container/... in which JVM is running.

---------
### Progress
- [x] Change must not contain extraneous whitespace

### Reviewers
 * @DanHeidinga (no known github.com user name / role)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/crac pull/3/head:pull/3` \
`$ git checkout pull/3`

Update a local copy of the PR: \
`$ git checkout pull/3` \
`$ git pull https://git.openjdk.java.net/crac pull/3/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 3`

View PR using the GUI difftool: \
`$ git pr show -t 3`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/crac/pull/3.diff">https://git.openjdk.java.net/crac/pull/3.diff</a>

</details>

PR #16, opened 2022-02-11
Provide arguments for restore

AntonKozlov:
This change adds an ability to receive a new set of command-line arguments in the restored Java instance. The supplied demo code shows a faster replacement for `javac`.

---------
### Progress

### Reviewers
 * [Alexey Bakhtin](https://openjdk.java.net/census#abakhtin) (@alexeybakhtin - no project role) ⚠️ Review applies to b45a5b85adefece83ba2843c06ceeeb59d3f2778

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/crac pull/16/head:pull/16` \
`$ git checkout pull/16`

`$ git checkout pull/16` \
`$ git pull https://git.openjdk.java.net/crac pull/16/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 16`

`$ git pr show -t 16`

</details>

<a href="https://git.openjdk.java.net/crac/pull/16.diff">https://git.openjdk.java.net/crac/pull/16.diff</a>

</details>


AntonKozlov:
The workflow for checkpoint could be:

```
$BUILDDIR/images/jdk/bin/java
-XX:CRaCCheckpointTo=$CRDIR
-Xshare:off # disable CDS for faster start after restore
-XX:-UsePerfData # disable jps for
-jar $BUILDDIR/images/jdk/demo/crac/JavaCompilerCRaC/JavaCompilerCRaC.jar
        # provide a set of warm-up workloads
        -nowarn                            -d tmp @${BUILDDIR}/jdk/modules/java.base/_the.java.base_batch.filelist --
        -nowarn -XDignore.symbol.file=true -d tmp @${BUILDDIR}/jdk/modules/java.desktop/_the.java.desktop_batch.filelist --
        -nowarn -XDignore.symbol.file=true -d tmp @${BUILDDIR}/jdk/modules/java.xml/_the.java.xml_batch.filelist
```

And after that, compile anything by
```
$BUILDDIR/images/jdk/bin/java
        -XX:CRaCRestoreFrom=$CRDIR
        Compile # does not mean anything
        "${TARGET_ARGS[@]}"
```
Or more concrete:
```
$ time jdk/bin/java -XX:CRaCRestoreFrom=./cr Compile HelloWorld.java
real    0m0.124s
user    0m0.269s
sys     0m0.075s

$ time jdk/bin/javac HelloWorld.java
real    0m0.380s
user    0m0.817s
sys     0m0.072s
```

src/java.base/share/classes/javax/crac/Core.java
```diff
@@ -82,4 +82,13 @@ public static void checkpointRestore() throws
             throw newException;
         }
     }
+
+    /**
+     * Gets new arguments provided after restore.
```
alexeybakhtin, line 87: May be description could be extended. I think it's worth mentioning that the new arguments do not replace original arguments, it is application-level arguments and it's application's responsibility to parse and apply these additional arguments. An empty array is returned in case of no additional arguments are provided.

DanHeidinga (commented):
Is there a design doc that describes the intended behaviour and limitations of this approach?  I've read the description in the PR but there's more to call out about what options can reasonably change.  ie: jvm options are unlikely to be successfully changed while passing new application args may be OK (if the app is written to handle that)

AntonKozlov:
> Is there a design doc that describes the intended behaviour and limitations of this approach? I've read the description in the PR but there's more to call out about what options can reasonably change. ie: jvm options are unlikely to be successfully changed while passing new application args may be OK (if the app is written to handle that)

At the momement only CLI arguments are passed. I think it won't be hard to extended this for system properties (-D...). Environment variables would be more tricky. JVM options are of course under a big question. I'm looking at CLI arguments as they cannot change,  in contrast with properties or the environment (I'm thinking e.g. an another java thread that may change them). Some JVM properties can be changed by existing mechanisms (`MANAGEABLE`). But unlikely all JVM options can be made manageable -- although in theory possible, this won't be practical (resize heaps because Xmx, regenerate code because codegen options,..). But some of options are not manageable but can be allowed, like PrintCompilation. Either they need be turned to manageable, or another term should be invented, or we may use ad-hoc checks for each and warn the user the option may not be changed. Let's try to find a useful and practical set? It looks to be CLI arguments, system properties, environment, and a subset of JVM options.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -5869,7 +5932,11 @@ static int checkpoint_restore() {
     return JVM_CHECKPOINT_ERROR;
   }
 
-  return info.si_int;
+  if (0 < info.si_int) {
+    *argp = get_new_args(info.si_int);
```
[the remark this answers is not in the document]
AntonKozlov: Idea is exactly to avoid data corruption. For criu, it's a PID of the restoring process, that is, the PID that would java process receive with the usual start (without checkpoint/restore). The restoring process exists for the whole life of the restored java instance, so it is impossible for another restoring process to obtain the same ID. 

For pauseengine, there is a slight change of ID reuse, as restoring process may exit earlier than the target VM pick up the data. But since chances are small, and pauseengine is mostly for debugging, I don't think it is a big problem. To mitigate, another ID generation scheme could be used, e.g. the ID could be the PID of the target VM.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -5835,7 +5841,64 @@ static int call_crengine() {
...
+
+    int argslen = strlen(args);
+    int wret = write(shmfd, args, argslen);
+    if (argslen != wret) {
+        if (wret < 0) {
+            perror("write shm");
+        } else {
+            fprintf(stderr, "write shm truncated");
+        }
+        close(shmfd);
```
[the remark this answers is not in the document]
AntonKozlov: Nice catch, thanks!

src/java.base/share/classes/javax/crac/Core.java
AntonKozlov: I was trying to provide better docs without over-specifying current implementation and realized that the new functionality can be done without an API change. Please look at the latest state.

AntonKozlov:
I think I found a better approach to achieve the same. I didn't like the first argument was going to be ignored most of the time (considering use-cases like in the demo). So instead, we can require arguments provided for restore need to be an actual Class, which main method will be called, and the arguments. So `java -XX:CRaCRestore Main arg1" will restore the instance and execute Main.main(arg1) in that instance. 

Comparing with the previous approach, it's like a program did
```
checkpointRestore();
mainClass, arguments = getAndParseNewArguments()
invoke(mainClass, "main", arguments)
```

Now the target Main class may handle new arguments right away, or store them to be used for later, like it was done in versions up to v3[1].

The demo is extended for Compile to be the Class, so examples above are still valid. Start-up time is roughly the same (actually a bit better, but it should be a noise).

[1] https://mail.openjdk.java.net/pipermail/crac-dev/2022-March/000132.html

src/java.base/share/classes/javax/crac/Core.java
alexeybakhtin: New approach looks great. Thank you

AntonKozlov:
@alexeybakhtin Thank you very much for the review!

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -140,18 +146,45 @@ private static void checkpointRestore1() throws
...
+            String[] args = newArguments.split(" ");
+            if (args.length > 0) {
+                try {
+                    Class<?> newMainClass = Class.forName(args[0], false,
+                        callerClass.getClassLoader());
+                    Method newMain = newMainClass.getDeclaredMethod("main",
+                        String[].class);
+                    newMain.invoke(null,
+                        (Object)Arrays.copyOfRange(args, 1, args.length));
+                } catch (Throwable e) {
```
DanHeidinga, line 172: Is `Throwable` the right thing to catch here?  I think Error-subclasses, like OutOfMemoryError or VerifyError, should propagate past here without being added to the suppressed set of the `restoreException`.

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -179,7 +213,7 @@ public static void checkpointRestore() throws
             if (!checkpointInProgress) {
                 try {
                     checkpointInProgress = true;
-                    checkpointRestore1();
+                    checkpointRestore1(Reflection.getCallerClass());
```
DanHeidinga, line 216: The caller Class is being passed in to get access to a Classloader object to load the new main class.  This means that if the `Core::checkpointRestore` caller is loaded by for example, application specific classloader, it may not have access to the core classpath.

Would it be a cleaner model if we require that the new main class must be a sibling of the original main class and therefore on the classpath/modulepath?  We could avoid using `getCallerClass` and directly pass the `ClassLoader::getSystemClassLoader` in that case.

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -140,18 +147,49 @@ private static void checkpointRestore1() throws
...
+
+        if (newArguments != null && newArguments.length() > 0) {
+            String[] args = newArguments.split(" ");
+            if (args.length > 0) {
+                try {
+                    Class<?> newMainClass = Class.forName(args[0], false,
+                        ClassLoader.getSystemClassLoader());
+                    Method newMain = newMainClass.getDeclaredMethod("main",
+                        String[].class);
+                    newMain.setAccessible(true);
```
DanHeidinga, line 171: Good catch on adding the `setAccessible` call.  

Given that the VM is doing the lookup based on the command line parameters, we should be wrapping this with a `AccessController::doPrivileged` block as a SecurityManager shouldn't be able to block this accessibility request.

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -140,18 +147,49 @@ private static void checkpointRestore1() throws
...
+                    newMain.invoke(null,
+                        (Object)Arrays.copyOfRange(args, 1, args.length));
+                } catch (ClassNotFoundException    |
+                         InvocationTargetException |
+                         NoSuchMethodException     |
+                         IllegalAccessException e) {
```
DanHeidinga, line 177: Thanks for adapting this to be more specific on what it catches.

I don't think it's quite right yet though as the current code will allow exceptions thrown by the `newMain` method (and its callees) to propagate past the checkpoint spot and makes the caller of the checkpoint code need to handle them.  

At the time `Core.checkpointRestore();` is called, we may have stack that looks llike:
```
TOS
Core.checkpointRestore();
Foo.bar();
Foo.foobar();
SomeOtherClass.method();
OriginalClass.main();
```

And when we restore, we load an execute a new `main()` method as though it was called where `Core.checkpointRestore();` was previously on the stack resulting in:
```
TOS
newMain.main();
Core.checkpointRestore();
Foo.bar();
Foo.foobar();
SomeOtherClass.method();
OriginalClass.main();
```

So exceptions thrown by code called from `newMain` should not propagate past `Core.checkpointRestore();` without being wrapped in a `RestoreException`.  `Error`-subclasses should propagate.

I think the code should be refactored to something like:
```
               } catch(Exception t) {
                   assert checkpointException == null :
                        "should not have new arguments";
                    if (restoreException == null) {
                        restoreException = new RestoreException();
                    }
                    restoreException.addSuppressed(e);
                }
```
as that correctly catches all Exceptions but lets the Errors propagate past

ashu-mehra:
I feel it is a useful feature be able to pass new arguments to the restored process. But looking from the user perspective, I see few concerns in the current approach to implement it:
- the current implementation is forcing user to create two classes with the main() method
- consequently the command line to run the application also differs depending on whether the user is running for the first time to take checkpoint, or restoring from the checkpoint
- the stack trace of the main thread after restore has remnant frames from the "before checkpoint" state, which at the least would be confusing to the user

When I was thinking of ways to work around these concerns, I ended up with a solution very similar to the initial approach where an additional API to get the new env variables was added without the requirement to introduce a new class with main method. I feel that would have been a better approach.

I guess the concern with that approach is that the first argument is to be ignored as mentioned [here](https://github.com/openjdk/crac/pull/16#issuecomment-1059730025). I think that shouldn't matter from the user perspective. A user would prefer to use same class name every time to start the application and therefore the JVM can safely ignore the first argument. If the user specifies a different class name during restore, then the JVM should interpret it as a different application and the current checkpoint should not be used for restoring it. In this case the JVM would bail out with an exception.
Does that make sense?

AntonKozlov:
I'm looking at the new arguments as the way to specify an additional service action on restore. In the javac example, the action is the main reason to restore the instance, after which the original program completes. But in general, it's a way to influence the program behavior on restore, after it has been executed for a while before the checkpoint. After the action is completed, the execution returns back to the original program. The purpose is to allow more flexible images that can be purposed better to the current environment, like in general command line arguments allow more flexible program.

> * the current implementation is forcing user to create two classes with the main() method

It's not necessary two (initializing and restoring) main methods, there could be more. With the old approach, users would be forced to dispatch execution themselves. I had some not pretty array juggling code, and it will be repeated over and over in every program that attempts to use this feature. 

The new arguments are optional. And JDK may provide predefined services. Such as a store for arguments for later consuming, to get back to the old approach. 

> * consequently the command line to run the application also differs depending on whether the user is running for the first time to take checkpoint, or restoring from the checkpoint

It may be a good feature. Restore brings us into the middle of execution, not to the start of the original program. I think the opposite looks confusing, if the same class name means different things if used for checkpoint (start the main from the class) or to restore (continue from some point of the main of the specified class).

> * the stack trace of the main thread after restore has remnant frames from the "before checkpoint" state, which at the least would be confusing to the user

Some documentation indeed would be useful. I'm thinking about a comment for checkpointRestore(): "implementation may accept an additional action and arguments to be run after restore, after which they will be used to locate a class which `main` method will be called with provided arguments". There should be less surprise then.

> I guess the concern with that approach is that the first argument is to be ignored as mentioned [here](https://github.com/openjdk/crac/pull/16#issuecomment-1059730025). 

It's better to avoid the first argument then. I just did not want to modify the launcher code too much.

With the new approach, the first argument of restore means something very similar to the first argument of the normal start.

> A user would prefer to use same class name every time to start the application and therefore the JVM can safely ignore the first argument. If the user specifies a different class name during restore, then the JVM should interpret it as a different application and the current checkpoint should not be used for restoring it. In this case, the JVM would bail out with an exception.

For java -jar, would it be required to match -jar or with the name of the main class?

I see this may be appealing, but then two commands in @DanHeidinga 's Option 2 are too similar to each other, and the second one means something very different. 

```
$ java -XX:CRaCRestoreFrom=./cr Foo arg1 arg2 arg3
```

On the opposite, treating Foo as the class name allows us to implement a fallback if CRaCRestoreFrom fails. `Foo` then may proceed as usual, or it may detect the VM was not restored and fail gracefully.

-XX:+CRaCIgnoreRestoreIfUnavailable was designed for the fallback, but unfortunately, I broke it in #3.

src/java.base/share/classes/jdk/crac/Core.java
AntonKozlov: > the current code will allow exceptions thrown by the `newMain` method (and its callees) to propagate past the checkpoint spot and makes the caller of the checkpoint code need to handle them.

An exception from the newMain will be wrapped in InvocationTargetException, which is caught and is suppressed by the RestoreException. Right? Otherwise, this java code won't compile, as checkpointRestore is declared to throw only {Checkpoint,Restore}Exception.

DanHeidinga: I had to double check with `jshell` as I thought runtime exceptions would propagate past.  Turns out your 100% right about the all Exceptions/Errors/Throwables being wrapped with `InvocationTargetException`.

src/java.base/share/classes/jdk/crac/Core.java
AntonKozlov: I was unsure about SecurityManager, having it will be removed soon. However, while it's still there, and adding necessary code is not hard, it's worth fixing.

I'm trying to push the fix, but GitHub sends me "Internal Server Error". Hope this will be resolved soon.

ashu-mehra:
@AntonKozlov it appears I am looking at this feature from a different perspective. My model was  the first invocation of the application would create the checkpoint, and for subsequent invocation, JVM would restore the process from the checkpoint. For javac kind of applications, user would want to pass different arguments for every invocation, and that's where this PR would be helpful.

>  it's a way to influence the program behavior on restore, after it has been executed for a while before the checkpoint. After the action is completed, the execution returns back to the original program. 

This is interesting approach as well. Can you share some scenarios where the application would benefit from this feature. While the javac example shows how this approach would work, in my opinion, it doesn't fall into the category of "influence the program behavior on restore" because the javac has already completed its original task when the checkpoint is taken.

AntonKozlov:
> the first invocation of the application would create the checkpoint, and for subsequent invocation, JVM would restore the process from the checkpoint.

I don't want to give too much control to JVM (and take the control from the user). Not every program is sensible to restart. Now we always restore to the point of checkpoint. Defining it elsewhere (e.g. start of the app) is hard problem.

> Can you share some scenarios where the application would benefit from this feature. While the javac example shows how this approach would work, in my opinion, it doesn't fall into the category of "influence the program behavior on restore" because the javac has already completed its original task when the checkpoint is taken.

Here the example program is a loop over javac, not a pure javac. The current example exits right after restore https://github.com/openjdk/crac/pull/16/files#diff-4d20e6a15e5a09de1e430f18a7c84eb257eb3d2b98721a35f4853d3fd2dadf37R27, but with arguments we arrange another javac iteration before the exit. 

I can imagine that a micro-service may need to know coordinates of other services on restore. Command line arguments may be used to provide the path to a config file then.

In general, I would like arguments on restore to be as powerful as arguments for normal java start, to parameterize program as necessary.

ashu-mehra:
> I can imagine that a micro-service may need to know coordinates of other services on restore. Command line arguments may be used to provide the path to a config file then.

@AntonKozlov Thanks for the explanation. This makes the requirement a lot more clear now.

I have a slightly different take on the application's perspective for this requirement. I am looking at the need for new arguments as no different than the post-restore adjustments that the application may have to do. Currently this is implemented using `Resource` interface and I feel the same mechanism can be extended for handling the new arguments at the application level as well. This can be done by treating the class responsible for processing arguments as a `Resource` which would get the `afterRestore` notification and can then get the new arguments from the Context passed to it.

For instance, in the JavaCompilerCRaC example, `JavaCompilerCRaC` can extend `Resource`, implement `afterRestore()`, get the new args from the `Context` and make internal adjustments which could be reading the config file, or opening a socket on the port number, or in this example call `runJavac()`.

I have created a [patch](https://github.com/ashu-mehra/crac/commit/3360123f4d4671bfdcee129115300d0f108d3d8f) on top of your changes to highlight this behavior.

This approach also removes the requirement for the application to create another main class to be used only on restore. In fact, the user can just do `java -XX:CRaCRestoreFrom=<dir> arg1 arg2` and the new arguments  would be available to the application on restore. 

Your thoughts on this approach?

AntonKozlov:
> I am looking at the need for new arguments as no different than the post-restore adjustments that the application may have to do. 

Indeed, I was thinking about the same, and now I realized there is a subtle difference. There are a lot Resources, if each of them may process arguments, how would they interpret them? E.g. does `-f` parameter means the same thing for all resources in the system? If not, how to specify a parameter is designated to a particular Resource? Technically nothing prevents two Resources to assume they are in position to process arguments.

Resources are required when an object of a class may require checkpoint/restore notification, but for the program the existence of the object is not clear (the object may not be created, the object may be GCed, the class may not be loaded at all in this program configuration). If it's possible to program without Resource's, it's better than with them.

BTW, the JavaCompilerCRaC object on line 44 won't survive GC, so it won't be notified.
https://github.com/ashu-mehra/crac/commit/3360123f4d4671bfdcee129115300d0f108d3d8f#diff-4d20e6a15e5a09de1e430f18a7c84eb257eb3d2b98721a35f4853d3fd2dadf37R44

> Currently this is implemented using `Resource` interface and I feel the same mechanism can be extended for handling the new arguments at the application level as well. This can be done by treating the class responsible for processing arguments as a `Resource` which would get the `afterRestore` notification and can then get the new arguments from the Context passed to it.

I still think that taking args as method parameters is cleaner and more streamlined. I would agree that arguments should be stored somewhere and then referred by a method call, if regular CLI arguments are stored somewhere and then referred by the main method. But they are provided as parameters to the main method, so I think we should not diverge in this. I mean, we don't write java programs in such way:
```
class Main {
  public void main() {
    String args = getArgs();
  }
}
```

An additional benefit of arguments as parameters -- they could be GCed as soon as a method processing them completes. Having arguments stored would make them eternally consuming memory.

The context to store args does not look as the right place: context's are forming hierarchy, so you'd need to set the args for downstream context when it's notified from the parent context. And arguments make sense only during afterRestore execution.

ashu-mehra:
> There are a lot Resources, if each of them may process arguments, how would they interpret them? E.g. does -f parameter means the same thing for all resources in the system? If not, how to specify a parameter is designated to a particular Resource? Technically nothing prevents two Resources to assume they are in position to process arguments.

> Resources are required when an object of a class may require checkpoint/restore notification, but for the program the existence of the object is not clear (the object may not be created, the object may be GCed, the class may not be loaded at all in this program configuration). If it's possible to program without Resource's, it's better than with them.

So I am assuming that if an application needs to accept new arguments to restore, it would be to make some internal adjustments. And I would expect the application to model any adjustments post restore using Resource interface. So essentially, what I am trying to say is if the application is using new arguments, there would be _some_ `Resource` that would take _some_ action in `afterRestore()` notification based on the new args.
Do you see any other way the new arguments could be used?

> An additional benefit of arguments as parameters -- they could be GCed as soon as a method processing them completes. Having arguments stored would make them eternally consuming memory.

Well, this can be achieved if the new arguments are passed as parameters to `afterRestore()`. We then wouldn't need to store them anywhere.


I think the main concern from my point of view with the current approach was the new main class that the user needs to create to access the new arguments. And I realized another implication of this approach which somehow I missed out earlier (not sure if this has already been thought about) - it essentially gives unlimited control to the user to do anything on restore.
It allows the application developer to update the objects, instead of doing through the Resource interface. So now they have to choose which approach to use for post-restore adjustments. More importantly, it allows the user to basically run a different application on restore! I think this is not what we want, right?

AntonKozlov:
> So I am assuming that if an application needs to accept new arguments to restore, it would be to make some internal adjustments. And I would expect the application to model any adjustments post restore using Resource interface. So essentially, what I am trying to say is if the application is using new arguments, there would be _some_ `Resource` that would take _some_ action in `afterRestore()` notification based on the new args. Do you see any other way the new arguments could be used?

The problem is that you cannot guarantee a single resource will handle the arguments. Providing arguments for every resource will encourage Resources using them, while only a single Resource should actually do that.

Resources are required to register different unrelated modules of the program, from third-party libraries to JDK. Each Resource at the smallest is a file handler, or network connection, or e.g. a native part of Selector on Linux (EPoll). Resources are more suited for non-precise checkpoint which comes at a random moment in time -- when you cannot provide a single routine to de/re-initialize, or it's hard.

E.g. microservice main endpoint and service logging endpoint -- each endpoint should be likely a resource, as they are likely handled with different parts of the program and fitting them in the single resource will be rather hard (was the logging endpoing is configured, was it initialized,...).

> Well, this can be achieved if the new arguments are passed as parameters to afterRestore()

This will encourage resources using new arguments even more. I've thought about adding another kind of resources (related/completely unrelated to the existing ones), but I see only a little value in them, but they are not better than the class.

> I think the main concern from my point of view with the current approach was the new main class that the user needs to create to access the new arguments. 

I don't completely understand the concern. How a new Resource is different from this point of view? It's a change anyway. But with the new class it's possible to make it easier and to provide a common class in JDK for all users (I'm not sure the particular class is a great idea):

```
public class jdk.crac.util.StoreArgs {
  String[] newArgs;
  public static main(String[] args) {
    newArgs = args;
  }
  public static String[] getNewArgs() {
    return newArgs;
  }
}
```

That is, a class that called if you really need to store new arguments somewhere. 

Or another class that prints image state (Thread dump?) and exits.

> And I realized another implication of this approach which somehow I missed out earlier (not sure if this has already been thought about) - it essentially gives unlimited control to the user to do anything on restore. It allows the application developer to update the objects, instead of doing through the Resource interface. So now they have to choose which approach to use for post-restore adjustments.

I hope I've addressed this above. I expect that Resources won't need arguments as they do not need them now, and a separate mean for handling arguments would not necessary need to be a Resource.

> More importantly, it allows the user to basically run a different application on restore! I think this is not what we want, right?

Why not? An image could be suited for different purposes. A regular java program may have different entry points (classes with the main method) and only one of them employed in a single run. With the restore, the old program does not go away, just another class's main method is called before returning from checkpointRestore().

AntonKozlov:
I propose to move with the PR in its current form, unless it is wrong. If not, we can fix it later based on user's feedback. I deliberately don't add the documentation to preserve some flexibility.

ashu-mehra:
Dan and I have a discussion internally about this PR and we agree its worth going ahead with what we have in this PR. At this point there isn't much clarity on the the way the applications would need to access the new arguments, but providing some mechanism to make the new arguments available on restore would at least enable us to play with more real-world applications and use-cases having these requirements. Based on that experience we can always revisit and make adjustments if required.

AntonKozlov:
Thanks for discussing this! Yes, let's try to use this and then adjust.

PR #18, opened 2022-04-07
Collect CREngine child

AntonKozlov:
When VM calls for CREngine it does not properly awaits termination of the engine. In case of `criuengine`, this could lead for a dangling process in the CRIU image [1]. This patch adds missing waitpid that eliminates the problem.

[1] https://github.com/openjdk/crac/blob/crac/src/java.base/unix/native/criuengine/criuengine.c#L68

---------
### Progress

### Reviewers
 * [Dan Heidinga](https://openjdk.java.net/census#heidinga) (@DanHeidinga - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/crac pull/18/head:pull/18` \
`$ git checkout pull/18`

`$ git checkout pull/18` \
`$ git pull https://git.openjdk.java.net/crac pull/18/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 18`

`$ git pr show -t 18`

</details>

<a href="https://git.openjdk.java.net/crac/pull/18.diff">https://git.openjdk.java.net/crac/pull/18.diff</a>

</details>

### DIFF
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -5837,7 +5837,11 @@ static int call_crengine() {
   }
 
   pid_t pid = fork();
-  if (!pid) {
+  if (pid == -1) {
+    perror("cannot fork for crengine");
+    return -1;
+  }
+  if (pid == 0) {
     execl(_crengine, _crengine, "checkpoint", CRaCCheckpointTo, NULL);
     perror("execl");
     exit(1);
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -5826,13 +5826,27 @@ static bool compute_crengine() {
 }
 
 static int call_crengine() {
-  if (_crengine && !fork()) {
+  if (!_crengine) {
+    return -1;
+  }
+
+  pid_t pid = fork();
+  if (!pid) {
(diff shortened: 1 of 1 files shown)


AntonKozlov:
Ping? This a bit annoying problem that we checkpoint/restore an unnecessary process.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -5826,13 +5826,27 @@ static bool compute_crengine() {
 }
 
 static int call_crengine() {
-  if (_crengine && !fork()) {
+  if (!_crengine) {
+    return -1;
+  }
+
+  pid_t pid = fork();
```
DanHeidinga, line 5833: Should we check for fork() to fail?  If so, we need to handle `-1` separately from `0` (child) and `> 0` (parent)

DanHeidinga (commented):
Apart from the possible fork failure case question, this looks reasonable

DanHeidinga:
And not really related to this PR, but didn't we disable the non linux x64 pr builds?  Do we need to relook at that change?

AntonKozlov:
I've added the check for the failed fork, thanks! I forgot to merge upstream changes to the branch, now the status of the PR is green.

DanHeidinga (approved):
Looks good to me

AntonKozlov:
Thanks for review!

/integrate

PR #28, opened 2022-07-05
Secret keys zeroisation

alexeybakhtin:
The main idea is to destroy secrets as soon as possible after use. Otherwise, the secret could be checkpointed and restored multiple times on restore. It causes predictable secrets after restore.
* SecretKey implements Destroyable, so destroy() method should be called just after secret is used (it could affect performance)
* In the most cases the jdk.internal.ref.CleanerFactory.java handles secrets end of life and cleans memory. However, it happens on GC only, so it does not guaranty the keys cleaned on checkpoint 

Another alternative/additional solution is to make SecretKey is checkpoint resource and verify it on checkpoint. In case of secret initialized and not yet destroyed - throw CheckpointException or log a warning

Additionaly, TLS SessionCache should be cleaned on checkpoint -> SSLSessionContextImpl should implement Checkpoint Resource and clean sessionCache/sessionHostPortCache on beforeCheckpoint()

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/28/head:pull/28` \
`$ git checkout pull/28`

`$ git checkout pull/28` \
`$ git pull https://git.openjdk.org/crac.git pull/28/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 28`

`$ git pr show -t 28`

</details>

<a href="https://git.openjdk.org/crac/pull/28.diff">https://git.openjdk.org/crac/pull/28.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/sun/security/provider/HmacDrbg.java
+++ b/src/java.base/share/classes/sun/security/provider/HmacDrbg.java
@@ -60,9 +60,11 @@ public class HmacDrbg extends AbstractHashDrbg {
 
     // 800-90Ar1 10.1.2.2: HMAC_DRBG Update Process
     private void update(List<byte[]> inputs) {
+        SecretKeySpec keySpec = new SecretKeySpec(k, macAlg);
         try {
             // Step 1. K = HMAC (K, V || 0x00 || provided_data).
-            mac.init(new SecretKeySpec(k, macAlg));
+            mac.init(keySpec);
+            keySpec.destroy();
             mac.update(v);
             mac.update((byte) 0);
             for (byte[] input: inputs) {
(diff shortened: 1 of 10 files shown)


src/java.base/share/classes/sun/security/provider/HmacDrbg.java
```diff
@@ -84,13 +88,17 @@ private void update(List<byte[]> inputs) {
                 k = mac.doFinal();
 
                 // Step 5. V=HMAC(K,V).
+                keySpec = new SecretKeySpec(k, macAlg);
                 mac.init(new SecretKeySpec(k, macAlg));
```
snazarkin, line 92: Missed key replacement.

rvansa:
Hi @alexeybakhtin do you intend to work on this further/reply to the reviewers' comments?

PR #30, opened 2022-09-29
Environment vars propagation into restored process

wkia:
This PR provides functionality to propagate actual environment variables to a restored process, as well as the test for this functionality.

Env propagation is done in few steps:
- Store the actual environment before restoring
- After restoring, replace the restored `environ` with a new one.
- On `afterRestore` event, propagate the new environment into a restored process via `ProcessEnvironment`.

---------
### Progress

### Reviewers
 * [Anton Kozlov](https://openjdk.org/census#akozlov) (@AntonKozlov - **Reviewer**) ⚠️ Review applies to [0270dc4b](https://git.openjdk.org/crac/pull/30/files/0270dc4b3d21bebfcc25e0c3c6d8a779093f754f)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac pull/30/head:pull/30` \
`$ git checkout pull/30`

`$ git checkout pull/30` \
`$ git pull https://git.openjdk.org/crac pull/30/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 30`

`$ git pr show -t 30`

</details>

<a href="https://git.openjdk.org/crac/pull/30.diff">https://git.openjdk.org/crac/pull/30.diff</a>

</details>

### DIFF
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 1999, 2021, Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2017, 2021, Azul Systems, Inc. All rights reserved.
+ * Copyright (c) 2017, 2022, Azul Systems, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -291,6 +291,7 @@ class CracRestoreParameters : public CHeapObj<mtInternal> {
     jlong _restore_time;
     jlong _restore_counter;
     int _nprops;
+    int _nenvs;
   };
 
   static bool write_check_error(int fd, const void *buf, int count) {
@@ -340,7 +341,8 @@ class CracRestoreParameters : public CHeapObj<mtInternal> {
     header hdr = {
       restore_time,
       restore_counter,
-      system_props_length(props)
+      system_props_length(props),
+      0
     };
 
     if (!write_check_error(fd, (void *)&hdr, sizeof(header))) {
@@ -356,6 +358,37 @@ class CracRestoreParameters : public CHeapObj<mtInternal> {
         return false;
       }
       p = p->next();
+    }
+
+    {
+      // Write env vars
+      int envCnt = 0;
+      for (auto env = environ; *env; ++env)
+      {
+        if (!write_check_error(fd, *env, strlen(*env) + 1)) {
+          return false;
+        }
+        ++envCnt; 
+      }
+
+      // Update written header with env counter
+      const auto curPos = lseek(fd, 0, SEEK_CUR);
+      if (0 > curPos) {
+        perror("shm seek error");
+        return false;
+      }
+      hdr._nenvs = envCnt;
+      if (0 > lseek(fd, 0, SEEK_SET)) {
+        perror("shm seek error");
+        return false;
+      }
+      if (!write_check_error(fd, (void *)&hdr, sizeof(header))) {
+        return false;
+      }
+      if (0 > lseek(fd, curPos, SEEK_SET)) {
+        perror("shm seek error");
+        return false;
+      }
     }
 
     return write_check_error(fd, args, strlen(args)+1); // +1 for null char
@@ -6439,6 +6472,19 @@ bool CracRestoreParameters::read_from(int fd) {
     int idx = _properties->append(cursor);
     int prop_len = strlen(cursor) + 1;
     cursor = cursor + prop_len;
+  }
+
+  for (int i = 0; i < hdr->_nenvs; i++) {
+    const auto envname = cursor;
+    while ('=' != *cursor++) {
+      assert(cursor <= contents + st.st_size, "environment var name length exceeds shared memory size");
+    }
+    cursor[-1] = 0;
+    const auto envvalue = cursor;
+    while (0 != *cursor++) {
+      assert(cursor <= contents + st.st_size, "environment var value length exceeds shared memory size");
+    }
+    setenv(envname, envvalue, 1);
   }
 
   _args = cursor;
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -353,36 +353,8 @@ class CracRestoreParameters : public CHeapObj<mtInternal> {
     return write_check_error(fd, args, strlen(args)+1); // +1 for null char
   }
 
-  bool read_from(int fd) {
-    struct stat st;
-    if (fstat(fd, &st)) {
-      perror("fstat (ignoring restore parameters)");
-      return false;
-    }
+  bool read_from(int fd);
 
-    char *contents = NEW_C_HEAP_ARRAY(char, st.st_size, mtInternal);
-    if (read(fd, contents, st.st_size) < 0) {
-      perror("read (ignoring restore parameters)");
-      FREE_C_HEAP_ARRAY(char, contents);
-      return false;
-    }
-
-    _raw_content = contents;
-
-    // parse the contents to read new system properties and arguments
-    header* hdr = (header*)_raw_content;
-    char* cursor = _raw_content + sizeof(header);
-
-    for (int i = 0; i < hdr->_nprops; i++) {
-      assert((cursor + strlen(cursor) <= contents + st.st_size), "property length exceeds shared memory size");
-      int idx = _properties->append(cursor);
-      int prop_len = strlen(cursor) + 1;
-      cursor = cursor + prop_len;
-    }
-
-    _args = cursor;
-    return true;
-  }
 };
 
 class VM_Crac: public VM_Operation {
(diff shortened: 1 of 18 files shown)


src/hotspot/os/linux/os_linux.cpp
```diff
@@ -358,6 +368,14 @@ class CracRestoreParameters : public CHeapObj<mtInternal> {
       p = p->next();
     }
 
+    // Write env vars
+    for (char** env = environ; *env; ++env)
+    {
```
AntonKozlov, line 373:
```suggestion
    for (char** env = environ; *env; ++env) {
```

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6441,6 +6459,20 @@ bool CracRestoreParameters::read_from(int fd) {
     cursor = cursor + prop_len;
   }
 
+  {
+    char* env_mem = NEW_C_HEAP_ARRAY(char, hdr->_env_memory_size, mtArguments); // left this pointer unowned, it is freed when process dies
```
AntonKozlov, line 6463:
```suggestion
  char* env_mem = NEW_C_HEAP_ARRAY(char, hdr->_env_memory_size, mtArguments); // left this pointer unowned, it is freed when process dies
```

Here and below, we don't hesitate to introduce few local variables without creating a new syntax block. Please note the change once applied makes the code uncompilable.

AntonKozlov (approved):
LGTM, thanks!

Just few minor questionable style issues.

DanHeidinga:
One concern with this approach - it means that environment variables will change values after a restore.

It seems odd to say this is a concern when it's the intended behaviour of this PR but it is a concern.  Users typically cache environment variables in static fields or use them to make a one time decision.  They don't expect them (at least at the Java layer) to change value throughout a run of the same process.

This change means two reads of the same env var can give different results at different times which may put unsuspecting applications into inconsistent states if two locations read the env var before vs after a restore.  That's going to be a hard to debug issue.

The VM may also read env vars and bind tightly to the value.  Native code after a restore will still have the original env while java code the modified env.  Do we foresee any issues there?

wkia:
@DanHeidinga 
Hi, 
You're right in your concerns. Indeed the suggested enhancement changes the usual workflow, so users may be confused. 
That is why we expect users to explicitly adapt their applications in accordance with the behaviour and make sure it works, otherwise there is no guarantee the application run with CRaC is successful.

wkia:
@DanHeidinga 
For a simple scenario, when nothing is changed in the environment, user applications don't need to be changed, the applications work out-of-the-box.

Could we consider a scenario when a container has a different environment for restoring a process rather than the environment it was checkpoint'ed? The different environment means something has changed in the system, it doesn't matter why. In case the application is not prepared and the process doesn't expect that environment could be changed after restoration (and/or doesn't handle this correctly), the process may have outdated view on the env. So the process may need to be reconfigured to continue working. It's not about particular env var values, but system changes. In case we don't propagate all the changed env vars to the process, the process doesn't have a chance to reconfigure itself. So propagation of all the changed env vars seems necessary. Of course, users need to make necessary changes to prepare their applications for checkpoint/restore events, otherwise the applications couldn't work properly.

Speaking about debugging, currently users are able to create simple apps to print out vars, see the example below:

```
> java -XX:CRaCRestoreFrom=./restore_folder TestApp

public class TestApp {
    public static void main(String args[]) throws Exception {
        for (Map.Entry<String, String> e : System.getenv().entrySet()) {
            System.out.println(e.getKey() + " = " + e.getValue());
        }
    }
}
```

DanHeidinga:
> @DanHeidinga For a simple scenario, when nothing is changed in the environment, user applications don't need to be changed, the applications work out-of-the-box.
> 

Agreed.  We restore the process and the environment is the same so the application code doesn't need to be updated to make "new" decisions regarding env vars.

> Could we consider a scenario when a container has a different environment for restoring a process rather than the environment it was checkpoint'ed? The different environment means something has changed in the system, it doesn't matter why. 

When using CRIU, we're restoring a full process so on restore, the env is the **same** as it was at the checkpoint time.  The only way to make it different is to **inject** something.  We're not talking just about env changes but about how to inject new configuration into the system so that system can respond to the change.

> In case the application is not prepared and the process doesn't expect that environment could be changed after restoration (and/or doesn't handle this correctly), the process may have outdated view on the env. So the process may need to be reconfigured to continue working. 

We agree here.  I want to point out the challenge here is that most Java applications are not written to expect env vars to change.  We have a huge body of applications and libraries that expect the env to be stable and allowing the env to change will result in strange inconsistencies from 3rd party code that the user has no idea ever used the env for anything.

Think of those old dusty jar files that no one has source for any more but is still in wide use.  Those are the dark corners these kinds of changes trip over.

Java has no standard mechanism for updating env vars so applications don't expect them to change.

> It's not about particular env var values, but system changes. In case we don't propagate all the changed env vars to the process, the process doesn't have a chance to reconfigure itself. So propagation of all the changed env vars seems necessary. 

So we need a way to inject new configuration data into the restored application.  Env vars are a convenient way to do this as they are already used when deploying containers to inject configuration.  Wholesale replacement of the env seems like a really scary way to pull the rug out from under existing applications that may have been only partially configured at the time of the checkpoint.  

A full scale env replacement mechanism requires every existing library needs to be reviewed for use of System.getEnv and updated to reconfigure if their env vars change.  And that means all users of those libraries also need to be reviewed and potentially reconfigured.  It makes adoption harder and less safe.

By limiting access to the env prior to checkpoint (only a subset of env vars are available, maybe configurable?), allowing new env vars to be injected at restore, and having a limited way to override env vars, we contain the potential side effects and allow users to reason about the code they are going to run more easily.

There's a really good discussion of this problem in the OpenJ9 issues [0] and you can see how the design evolved from add only, to eventually a limited amount of override.  There's also a writeup from the GraalVM team about capturing build time state that applies here too. [1] Both are worth a read.

[0] https://github.com/eclipse-openj9/openj9/issues/13545
[1] https://github.com/graalvm/taming-build-time-initialization#host-machine-data-leakage

> Of course, users need to make necessary changes to prepare their applications for checkpoint/restore events, otherwise the applications couldn't work properly.
> 

Agreed.  Our design here can make adoption easier or harder.  The more we can take into account the existing code that applications depend on, the easier we can make the adoption path.  We don't want to make the old code behave in new ways, but we can make it safer (or not) to use that old code.

> Speaking about debugging, currently users are able to create simple apps to print out vars, see the example below:
> 
> ```
> > java -XX:CRaCRestoreFrom=./restore_folder TestApp
> 
> public class TestApp {
>     public static void main(String args[]) throws Exception {
>         for (Map.Entry<String, String> e : System.getenv().entrySet()) {
>             System.out.println(e.getKey() + " = " + e.getValue());
>         }
>     }
> }
> ```

This is a one really cool capability with CRaC in that we can run different applications from the same checkpoint.  It doesn't really address the serviceability concerns though - when dealing with end user problems, we might get the `java -version` output and a system core file.  To make it debuggable we need a way to see that the env vars were changed to avoid chasing  inconsistencies related to timing of env var lookup (pre checkpoint vs post restore).

wkia:
@DanHeidinga 
In case of env propagation there is a chance to get an app in some inconsistent state (when app does both caches a var and reads it via getenv in different peices of code.) That's true, but only for old applications which are not prepared for checkpoint/restore events. There is no problem for new apps adapted for C/R. (And as I mentioned before, an old appswould have problems anyway in case of changed env not propagated to the app.)
OTOH there are env vars which depend on each other. In case we only add new vars, there is the same chance to get the inconsistent state because of new var is propagated, but its dependent var not (due to it already exists). This might be a problem for both old apps and new apps.

Another yet scnario, an app uses env var to know a port numer to open. After the app restoration it turns that the port is already busy, so we could change the var value to reconfigure the app, but the app cannot reconfigure itself because we don't propagate env into the restored app.

I guess there may be a lot similar scenarios.

AntonKozlov:
Thank you to all participants of this interesting discussion. I see the points in different aspects, in Roman's implementation of the mechanism and in Dan's concerns about policy how it should be used, and how to debug possible issues. I propose to move with smaller steps and implement missing features and polish the policy in the subsequent PR(s), on top of the simpler implementation.

AntonKozlov:
Makes sense. I'll try to draft something to start from.

wkia:
Thank you all for your comments! As far as I understand, we've got some different opinions about further directions, so we need to work on principles to make the approaches consistent. 
Does the document creation stop us from merging this PR now? If no, I'd merge it.

AntonKozlov:
Merging is fine to me. I think policy, etc will be implemented on top, extending the functionality of this PR.

DanHeidinga:
> Merging is fine to me. I think policy, etc will be implemented on top, extending the functionality of this PR.

Agreed.  Let's merge this, define the principles, and then iterate on it.

DanHeidinga:
I wanted to follow up on this and thank everyone involved in the discussion for being open to working through the issues here.  And a big thanks to @wkia for persevering on this patch!

Looking forward to collaborating with all of you on the principles document.

wkia:
> We should also consider doing something similar to the OpenJ9 approach where we restrict the set of env vars available prior to the checkpoint (minimize the accidental use of checkpoint env), and limit the env var changes to only add new env vars (no inconsistencies). This got them a long ways in their work with Liberty though they did find it necessary to eventually support overriding some env vars.

Recently I realized that if users don't like some variables to be propagated into a restored process, they can prepare their environment as they want before a restore by setting and/or unsetting any environment variables they want. So there is no need to implement additional solutions in CRaC. I've extended the test with the example (#42).

DanHeidinga:
The updated test doesn't really address the concern though.  The key concern was:
> limit the env var changes to only add new env vars (no inconsistencies).

OpenJ9 did this by constraining the checkpoint env to only capture a limited subset of env vars and allowing restore to expand that set with new env vars.

The current CRaC approach still allows overriding env vars, which is something Java (and developers, and library developers) doesn't expect to happen so they haven't written their code to be aware these values can change.  Changing existing env vars on restore will just lead to inconsistent systems.  See [0] for more details on why this is the case.

[0] https://github.com/openjdk/crac/pull/30#issuecomment-1268433596

PR #31, opened 2022-10-11
Disable rseq in libc on checkpoint

AntonKozlov:
Restartable sequences (rseq) [0] may be used by the glibc [1]. Without proper support of rseq in the ptrace [2], CRIU fails to create the checkpoint [3]. Some paravirtualized environments like Docker on Mac, and WSL, which are commonly used during development, still do not provide a proper rseq support, leading to the error. A simple usability workaround is to disable rseq in glibc. The patch disables rseq if JVM is started with -XX:CRaCheckpointTo and there is no explicit setting of rseq for glibc.  The workaround is not going to live forever, just until rseq support is implemented in the majority of environments. Alternatives like more clever detection of rseq in the ptrace, or detection of a paravirtualized environment seem too complex, having that positive impact from rseq on java performance is unknown. 

[0] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d7822b1e24f2
[1] https://sourceware.org/git/?p=glibc.git;a=commit;h=95e114a0919d844d8fe07839cb6538b7f5ee920e. 
[2] https://github.com/CRaC/criu/blob/cc01f191639a5c2c988f49f1e314d17b055497b2/criu/kerndat.c#L944
[3] https://github.com/CRaC/criu/blob/cc01f191639a5c2c988f49f1e314d17b055497b2/criu/cr-dump.c#L225

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac pull/31/head:pull/31` \
`$ git checkout pull/31`

`$ git checkout pull/31` \
`$ git pull https://git.openjdk.org/crac pull/31/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 31`

`$ git pr show -t 31`

</details>

<a href="https://git.openjdk.org/crac/pull/31.diff">https://git.openjdk.org/crac/pull/31.diff</a>

</details>

### DIFF
--- a/src/java.base/share/native/launcher/main.c
+++ b/src/java.base/share/native/launcher/main.c
@@ -106,6 +106,33 @@ main(int argc, char **argv)
         int i, main_jargc, extra_jargc;
         JLI_List list;
 
+#ifdef __linux__
+        for (int i = 1; i < argc; ++i) {
+            if (strstr(argv[i], "-XX:CRaCCheckpointTo")) {
+                char* tunables = getenv("GLIBC_TUNABLES");
+
+                // do not try overwrite an existing tunable setting
+                if (tunables && strstr(tunables, "glibc.pthread.rseq")) {
+                    break;
+                }
+
+                char newtunes[4096];
+                int sz = snprintf(newtunes, sizeof(newtunes), "GLIBC_TUNABLES=%s:glibc.pthread.rseq=0", tunables);
+                if (sz < 0 || (int)sizeof(newtunes) <= sz) {
+                    fprintf(stderr, "Cannot update GLIBC_TUNABLES: does not fit\n");
+                    return 1;
+                }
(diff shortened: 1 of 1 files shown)


wkia:
THe fix seems working on WSL.
Checked on Ubuntu 22.04 (5.10.16.3-microsoft-standard-WSL2 x86_64 GNU/Linux)

simonis:
This fix seems overly complicated and I don't like the fact that we exec in the launcher. This doesn't seem to be a very common use case so why do we not simply document it and instruct users who are affected to set the environment variable themselves? That looks like a pretty simple workaround to me.

AntonKozlov:
Hmm, that's reasonable. Since we anyway need to mention how to disable the workaround, documenting the workaround seems to be enough indeed. Thanks for looking! I'm closing this PR.

PR #35, opened 2022-11-29
Assume module jars persistent

AntonKozlov:
Jar files opened by URLClassLoaders are assumed persistent. We need to handle module Jars the same way to prevent CheckpointException as reported by [1] 

[1] https://mail.openjdk.org/pipermail/crac-dev/2022-November/000380.html

### DIFF
--- a/src/java.base/share/classes/jdk/internal/module/ModuleReferences.java
+++ b/src/java.base/share/classes/jdk/internal/module/ModuleReferences.java
@@ -32,10 +32,13 @@ import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.lang.module.ModuleReader;
 import java.lang.module.ModuleReference;
+import java.lang.reflect.Method;
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -48,8 +51,12 @@ import java.util.jar.JarFile;
 import java.util.stream.Stream;
 import java.util.zip.ZipFile;
 
+import jdk.crac.Context;
+import jdk.crac.Resource;
+import jdk.internal.crac.JDKResource;
 import jdk.internal.jmod.JmodFile;
 import jdk.internal.module.ModuleHashes.HashSupplier;
+import jdk.internal.util.jar.JarFileCRaCSupport;
 import sun.net.www.ParseUtil;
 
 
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/jdk/internal/module/ModuleReferences.java
```diff
@@ -236,6 +244,27 @@ static JarFile newJarFile(Path path) {
         JarModuleReader(Path path, URI uri) {
             this.jf = newJarFile(path);
             this.uri = uri;
+            this.resource = new JDKResource() {
```
DanHeidinga, line 247: Would it be cleaner to implement this JDKResource subclass as a named class in `JarFileCRaCSupport` and wrap this logic so we have something like:
```
this.resource = JarFileCRaCSupport.wrapAndRegisterJarFile(this.jf);
```

It centralizes more of the Jar file support in one place....

AntonKozlov: Indeed, I've created ClassLoaderJarFile for that purpose. Moving the whole class to be used by ModuleReferences.

src/java.base/share/classes/jdk/internal/util/jar/PersistentJarFile.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+        Method zipBeforeCheckpoint = ZipFile.class.getDeclaredMethod("beforeCheckpoint");
+        @SuppressWarnings("removal")
+        Void v = AccessController.doPrivileged(new PrivilegedAction<Void>() {
+            public Void run() {
+                zipBeforeCheckpoint.setAccessible(true);
+                return null;
+            }});
```
DanHeidinga, line 56: Something to think about that doesn't need to be addressed now:

We may want to lazily cache the accessible reflect Method in a static field to avoid looking it up for each PersistentJarFile.  The value of doing that depends somewhat on how many jars we expect here and how much we care about time spent in checkpoint hooks.

DanHeidinga (approved):

src/java.base/share/classes/jdk/internal/util/jar/PersistentJarFile.java
AntonKozlov: Oh, you're right. Not sure when it will be an another chance to improve, so better be done now :) This implementation with reflection predated the SharedSecrets, which cleaner, safer, faster -- so I just used that.

DanHeidinga (approved):
Agreed.  The SharedSecrets approach is even better

AntonKozlov:
Thanks for review!

/integrate

PR #36, opened 2022-12-01
Fix [ASSERT] compilation error after fdf4eb8dd5c2a11176eb1ec8f40b3f4133a872aa

jankratochvil:
```
../../src/hotspot/os/linux/attachListener_linux.cpp: In function 'void assert_listener_thread()':
../../src/hotspot/os/linux/attachListener_linux.cpp:389:16: error: aggregate 'ResourceMark rm' has incomplete type and cannot be defined
  389 |   ResourceMark rm; // For retrieving the thread names
```


AntonKozlov (approved):
Indeed, the build fails if configured with --disable-precompiled-headers, thanks!

PR #41, opened 2023-01-30
RFC: -XX:CPUFeatures=0xnumber for CPU migration

jankratochvil:
Currently if you `-XX:CRaCCheckpointTo` on a better CPU and `-XX:CRaCRestoreFrom` on a worse CPU the restored OpenJDK will crash.

1. An obvious reason is that JIT-compiled code is using CPU features not implemented on the CPU where the image is restored.
2. A second reason is that glibc has a similar problem, its PLT entries point to CPU optimized functions also crashing on the worse CPU. https://sourceware.org/glibc/wiki/GNU_IFUNC

(1) could be solved somehow automatically by deoptimizing and re-JITing all the JIT code. But that would defeat the performance goal of restoring a ready image in the first place. Therefore there had to be implemented a new OpenJDK option:

> use -XX:CPUFeatures=0xnumber with -XX:CRaCCheckpointTo when you get an error during -XX:CRaCRestoreFrom on a different machine

It is intended to specify the lowest common denominator of all CPUs in a farm. Instead of a possible crash of OpenJDK it will now refuse to run:

> Error occurred during initialization of VM
> You have to specify -XX:CPUFeatures=0x421801fcfbd7 during -XX:CRaCCheckpointTo making of the checkpoint; specified -XX:CRaCRestoreFrom file contains CPU features 0x7fff9dfcfbf7; this machine's CPU features are 0x421801fcfbd7; missing features of this CPU are 0x3de79c000020 = 3dnowpref, adx, avx512f, avx512dq, avx512cd, avx512bw, avx512vl, sha, avx512_vpopcntdq, avx512_vpclmulqdq, avx512_vaes, avx512_vnni, clflushopt, clwb, avx512_vbmi2, avx512_vbmi

(2) has been implemented according to Anton Kozlov's idea that glibc can just reset its IFUNC PLT entries any time later (after restore), not just during the first initialization of glibc. That has currently a problem that it has turned out to be very invasive into private glibc structures. It could work somehow with glibc debuginfo (*-debuginfo.rpm or *-dbg.deb) installed but that has been considered as unacceptable requirement just to run CRaC. Therefore I have provided this proof of concept while I will propose such feature for glibc upstream where it is sure easily implementable.

If upstream glibc maintainers do not like the IFUNC reset idea then I do not think this hacky IFUNC reset patching many glibc internal data structures is a good way forward for a 3rd party implementation like CRaC/OpenJDK. In such case I believe one should switch to using GLIBC_TUNABLES environment variable, re-execing OpenJDK after converting the `-XX:CPUFeatures` OpenJDK format into glibc GLIBC_TUNABLES format. Unfortunately there is a precedent OpenJDK upstream has already rejected such re-exec idea in the past: https://github.com/openjdk/crac/pull/31#issuecomment-1275707621

That IMO does not preclude trying the same for this case.

- Debian 11 x86_64: It does not work, glibc is too different and inlined there.
- Debian 12 x86_64: It works even without libc6-dbg as its offsets are the default.
- Fedora 36 x86_64: It works as on Fedoras glibc debuginfo is embedded.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/41/head:pull/41` \
`$ git checkout pull/41`

`$ git checkout pull/41` \
`$ git pull https://git.openjdk.org/crac.git pull/41/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 41`

`$ git pr show -t 41`

</details>

<a href="https://git.openjdk.org/crac/pull/41.diff">https://git.openjdk.org/crac/pull/41.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/41#issuecomment-1407943951)


src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6019,8 +6021,143 @@ class CracSHM {
...
+      while (errno = 0, ent = readdir(dir)) {
+        const char *name = ent->d_name;
+        if (strcmp(name, ".") == 0)
+          continue;
+        if (strcmp(name, "..") == 0)
+          continue;
+        char *endptr;
+        unsigned long l = strtoul(name, &endptr, 10);
+        pid_t ent_tid = l;
+        if (l >= LONG_MAX || ent_tid != (long) l) {
```
DanHeidinga, line 6050: `l` is an `unsigned long` and I think `LONG_MAX` is a signed value as there's also a `ULONG_MAX`.  Are we using this to validate that `l` is a positive value?  Otherwise, I'm not sure I understand this condition.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6019,8 +6021,143 @@ class CracSHM {
...
+      if (closedir(dir)) {
+        tty->print_cr("JVM: Error closing %s: %m", dirname);
+        return false;
+      }
+    } while (signalled);
+    return true;
+  }
+  static pthread_mutex_t signalled_mutex;
+  static pthread_cond_t signalled_cond;
+  static size_t signalled;
```
DanHeidinga, line 6074: The above `template<class Callback> bool all_threads` also uses a "signalled" local variable.  It might be clearer to rename one of these so future readers don't have to think about scoping and shadowing rules

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6019,8 +6021,143 @@ class CracSHM {
...
+    assert(handler_signo == signo, "handler signo");
+    int err;
+    err = pthread_mutex_lock(&signalled_mutex);
+    assert(!err, "pthread error");
+    ++signalled;
+    err = pthread_mutex_unlock(&signalled_mutex);
+    assert(!err, "pthread error");
+    err = pthread_cond_signal(&signalled_cond);
+    assert(!err, "pthread error");
+    pthread_mutex_t unused_mutex = PTHREAD_MUTEX_INITIALIZER;
```
DanHeidinga, line 6086: Waiting on a local mutex deserves a comment to explain the approach.  I think we're waiting until the thaw operation signals the resume condition but better to be explicit in the code on the intended interactions between the waiters

DanHeidinga (commented):
This is a nice piece of engineering to make it work.  I'm concerned about the maintainability and correctness of the approach though, when running on different systems / glibcs and over time as glibc is recompiled.

Ashutosh suggested an alternative approach back in Feb 2022 [0] [1] of providing a single "-XX:[+-]PortableCode" with a common enough baseline set of features and the use of GLIBC_TUNABLES to control the glibc bindings.  

This is very similar to the approach the OpenJ9 team has taken in their CRIU efforts and have found that it works reasonable well for the use cases they have targeted (mostly restore inside a container).

[0] http://cr.openjdk.java.net/~heidinga/crac/Portability_of_checkpoints.pdf
[1] https://mail.openjdk.org/pipermail/crac-dev/2022-February/000114.html

src/hotspot/os/linux/os_linux.cpp
jankratochvil: The number must fit in `pid_t` (otherwise it is a weird Linux kernel). Therefore `int`. But there is no `strtoi`. Also PID must not be negative (otherwise it is a weird Linux kernel again). So I found this as the most simple and safe code to validate `name` contains a non-negative number which can fit in `pid_t`.

src/hotspot/os/linux/os_linux.cpp
jankratochvil: OK, yes, renamed the local variable to `retry`.

src/hotspot/os/linux/os_linux.cpp
jankratochvil: Added: `// Just wait until thaw() is called. No mutex is needed for that.`

src/hotspot/os/linux/os_linux_ifunc.cpp
```diff
@@ -0,0 +1,734 @@
...
+#include <limits.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/mman.h>
+#include <sys/user.h>
+#include <stdlib.h>
+
+// These offsets are for Debian 12 x86_64:
+// gdb -batch /lib64/ld-linux-x86-64.so.2 -ex
+static unsigned   l_scope_offset = 0x3b0;
```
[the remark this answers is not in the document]
jankratochvil: These offsets are not stable, I hope they won't change in one stable OS release but still they can. If there are any problems one can always install `*-debuginfo.rpm` or `*-dbg.deb` which will override all these assumed values. Maybe if there is no debuginfo available there should be a database of `libc.so.`6 binary hashes and refuse to run if none is found.
Still I find this patch only as a proof of concept and a some codebase to ask upstream glibc for inclusion of an IFUNC-reset functionality into glibc mainline. As I stated in Comment 0 if glibc refuses such upstreaming I believe CRaC should rather use `GLIBC_TUNABLES` anyway.
This IFUNC-reset functionality looked simple originally but in the end it got much more complicated than I expected. I would not probably implement it if I could see it will not be simple enough. But when it was already mostly written I have finished it.

src/hotspot/os/linux/os_linux_ifunc.cpp
```diff
@@ -0,0 +1,734 @@
...
+  int err = munmap((void *)p, file_page_size(fd));
+  assert(!err);
+}
+
+struct symtab_lookup {
+  const char *name;
+  const void *start, *end;
+  unsigned sht;
+};
+static int symtab_lookup_iterate_phdr(struct dl_phdr_info *info, size_t size, void *data_voidp) {
```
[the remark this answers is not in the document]
jankratochvil: Already discussed above.

jankratochvil:
> I'm concerned about the maintainability and correctness of the approach though, when running on different systems / glibcs and over time as glibc is recompiled.

Me too. This is why I also called it just "RFC". But it is not up to me whether this temporary and/or a bit fragile solution gets accepted for CRaC or not.

> Ashutosh suggested an alternative approach back in Feb 2022 [0] [1] of providing a single "-XX:[+-]PortableCode" with a common enough baseline set of features

IIUC you can just use `-XX:CPUFeatures=0` with this patch for the same effect. As my employer is all about Java Performance I find useful to provide the option of more optimal runtime JIT features.

> and the use of GLIBC_TUNABLES to control the glibc bindings.

This is what I say in Comment 0:

> If upstream glibc maintainers do not like the IFUNC reset idea [...]. In such case I believe one should switch to using GLIBC_TUNABLES environment variable, [...]

> This is very similar to the approach the OpenJ9 team has taken in their CRIU efforts and have found that it works reasonable well for the use cases they have targeted (mostly restore inside a container).

Could you send some more specific link for that functionality?

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6037,6 +6174,10 @@ static int checkpoint_restore(int *shmid) {
   } while (sig == -1 && errno == EINTR);
   assert(sig == RESTORE_SIGNAL, "got what requested");
 
+  linux_ifunc_reset();
```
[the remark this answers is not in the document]
jankratochvil: You are right, there is a race, I will fix that, thanks.

src/hotspot/os/linux/os_linux.cpp
DanHeidinga: > The number must fit in `pid_t` (otherwise it is a weird Linux kernel). Therefore `int`. But there is no `strtoi`.

If we're boxing to an int, then I think we should check if `l >= INT_MAX` rather than `LONG_MAX`.

jankratochvil: The `int` check is there `ent_tid != (long) l` (as nobody says `pid_t` is really an `int`). The `l >= LONG_MAX` is there to make valid the cast `(long) l`. But this is IMO a [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality).

DanHeidinga:
> > This is very similar to the approach the OpenJ9 team has taken in their CRIU efforts and have found that it works reasonable well for the use cases they have targeted (mostly restore inside a container).
> 
> Could you send some more specific link for that functionality?

Details on the glibc tunables used by OpenJ9 are in: https://github.com/eclipse-openj9/openj9/issues/14253

A script setting them for some of the tests: https://github.com/eclipse-openj9/openj9/commit/172bcf7244116186d4edde019473db69fb5d9559

Open Liberty also sets the glibc tuneables as part of their server scripts for use with CRIU which I think shows consumers need to be aware of the issue: https://github.com/OpenLiberty/open-liberty/pull/22132

> > Ashutosh suggested an alternative approach back in Feb 2022 [0] [1] of providing a single "-XX:[+-]PortableCode" with a common enough baseline set of features
> 
> IIUC you can just use `-XX:CPUFeatures=0` with this patch for the same effect. As my employer is all about Java Performance I find useful to provide the option of more optimal runtime JIT features.

Also a big fan of Java performance and a fan of avoiding foot-guns =)  There are benefits to both approaches, I think most users will want a simple "make it portable given a good baseline system" (aka how most native binaries are built) and those who need the most performance can use the existing Hotspot enable/disable flags to allow particular cpu features.

AntonKozlov:
> I believe one should switch to using GLIBC_TUNABLES environment variable, re-execing OpenJDK after converting the -XX:CPUFeatures OpenJDK format into glibc GLIBC_TUNABLES format. Unfortunately there is a precedent OpenJDK upstream has already rejected such re-exec idea in the past: #31 (comment)

Could you explore that futher? The concerns about #31 were merely about complexity of implementation vs reward. Looking backward, I probably overthougth about documenting the workaround, although unlikely any user would notice the workaround exists.

But here, re-exec seems more simple solution. That would also make `Freeze` mechanism unnecessary, as I understand that is another implementation of safepoint.

AntonKozlov:
> > Ashutosh suggested an alternative approach back in Feb 2022 [0] [1] of providing a single "-XX:[+-]PortableCode" with a common enough baseline set of features
> 
> IIUC you can just use `-XX:CPUFeatures=0` with this patch for the same effect.

That's nice! Since portability affects the final perfomance on restore, the user should have a control between portability and performance. CPUFeatures=0 looks indeed as an impelmentation of some -XX:+PortableCode that was proposed before.

ashu-mehra:
IIUC `CPUFeatures=0` disables all the CPU features, which was never the intention behind PortableCode option. With `PortableCode` we wanted to enable a baseline set of feature which would cater to most commonly used microarchitectures.

Selection of default cpu features is a trade-off between portability and performance. Currently the cpu features are selected based on the native architecture which is one extreme as it makes the generated code less portable (unless you run on a sufficiently old arch). Similarly CPUFeatures=0 is another extreme of that which would make the code portable but is highly likely to degrade the performance. PortableCode lies somewhere in between.

In this context gcc provides a good analogy. It does not target i386 instruction set by default just to make the code more portable. The default target arch used by gcc is x86-64 which corresponds to "A generic CPU with 64-bit extensions". And if you look at the features that are enabled by default, you will find sse and sse2 extended instructions sets in there [0].

Now, I also understand CPUFeatures=<n> can be used to achieve the same affect that PortableCode option would. However, with PortableCode the default set of features would be automatically selected by the JVM; the user doesn't need to figure out the value to pass to CPUFeatures to make the code portable.
And as Dan mentioned earlier, even with PortableCode if the user wants to enable a particular extended instruction set, existing Hotspot options can be used.

[0] Use `gcc -Q --help=target`to see the list of options enabled or disabled

ashu-mehra:
> so are you proposing -XX:+PortableCode to change default CPU features to some conservative set, that still can be extended by existing Hotpost options?

That's right.

> even for CPUFeatures we may want values like generic, native (again, in gcc terms)

Yes, something like CPUFeatures=generic would serve the same purpose as PortableCode.  IMO adding something like `-XX:CPUFeatures=[generic|native]` or `-XX:[+-]PortableCode` is more appropriate. To enable/disable extended instruction sets, hotspot already has options (like UseSSE). Combining those options with `CPUFeatures=[generic|native]` or `PortableCode` would provide an easier way for the user to control the cpu features that hotspot should use.

AntonKozlov:
There is some room anyway for XX:CPUFeatures= to take a specification of a feature set in uniform format, that is clear for user without a need to look in hotspot sources. It may be more simple in implementation, that will be easier to verify. The subsequent -XX: options will likely be still available, if user to prefer them; I assume that it's a rare situation when they need to be altered, but CPUFeatures will likely be more frequently used. CPUFeatures and -XX: flags looks to be on different level of abstractions for users. Even though they are not orthogonal.

jankratochvil:
> To enable/disable extended instruction sets, hotspot already has options (like UseSSE). Combining those options with `CPUFeatures=[generic|native]` or `PortableCode` would provide an easier way for the user to control the cpu features that hotspot should use.

I did not expect anyone would study which CPU model should use which `UseSSE`, `UseAVX`, `UseCLMUL` and other options.

The current code just prints `You have to specify -XX:CPUFeatures=0x421801fcfbd7 during -XX:CRaCCheckpointTo` as described in the Comment 0. So one should just run it once on the oldest farm machine and copy/paste the number to the snapshotting script. Or just to run it on all of the farm machines and do a logical AND of all the numbers if one is unsure which machine is the oldest one in your farm.

I sure do not mind which options should be used. I agree `-XX:CPUFeatures=[generic|native]` would be useful (`native` not so useful as it is the current OpenJDK default).

jankratochvil:
> But here, re-exec seems more simple solution.

OK, if the glibc upstreaming files then there is at least agreement on this way forward. Also maybe OpenJDK should use that anyway to be compatible with older (current) glibc versions.

> That would also make `Freeze` mechanism unnecessary, as I understand that is another implementation of safepoint.

I was originally expecting all threads are at a safepoint during CRaC snapshot. But in practice I have found out they are not. This is why I had to implement the `Freeze` mechanism. Is there some way to stop all the threads at a safepoint?

src/hotspot/os/linux/os_linux.cpp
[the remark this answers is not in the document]
jankratochvil: The race should be fixed now. Although still all threads of OpenJDK should be at a safepoint but in practice they are not, why so?

AntonKozlov:
> This is why I had to implement the Freeze mechanism. Is there some way to stop all the threads at a safepoint?

That should not happen with java threads. Could that be a native thread?

jankratochvil:
> That should not happen with java threads. Could that be a native thread?

It's a pity but I can no longer reproduce the problem (and I no longer have the original core file). So I think I can remove the Freezer for now and if anyone reports it we can discuss whether to reintroduce the Freezer or choose some other solution.

jankratochvil:
I hope I have addressed all the comments.
- `-XX:+ShowCPUFeatures` produces: `This machine's CPU features are: -XX:CPUFeatures=0x7fff9dfcfbf7`
- `-XX:CPUFeatures=native`
- `-XX:CPUFeatures=generic` - the only problem is what features should be the default, the current one is definitely wrong

ashu-mehra:
> One reasonable design point for generic is to match the set of features allowed when compiling the JVM so that a checkpoint is as portable as running the app with the java executable. 

I used `LOG=cmdlines make images` to see the command lines used during compilation. I don't see any `march` or `mtune` option in the command line. As I mentioned earlier we can use `gcc -Q --help=target` to see the extended instructions sets enabled by gcc (would need some filtering I believe). I guess we can use that set as the baseline to begin with.

jankratochvil:
> That should not happen with java threads. Could that be a native thread?

So I have accidentally reproduced it. Maybe it is a native thread. But what else to do with it than the `Freezer`? `checkpoint_restore()` is still executing `linux_ifunc_reset()` that time.

```
Core was generated by `/home/azul/azul/example-jetty/../crac-git/build/linux-x86_64-server-slowdebug/j'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __strlen_evex () at ../sysdeps/x86_64/multiarch/strlen-evex.S:69
69	../sysdeps/x86_64/multiarch/strlen-evex.S: No such file or directory.
[Current thread is 1 (Thread 0x7f092cfff6c0 (LWP 26563))]
(gdb) bt -15
#2920 <signal handler called>
#2921 __memset_evex_unaligned_erms () at ../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:167
#2922 0x00007f0955eb2cfa in GuardedMemory::set_user_bytes(unsigned char) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2923 0x00007f0955eb2c41 in GuardedMemory::release_for_freeing() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2924 0x00007f09564382cf in os::free(void*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2925 0x00007f095592ac4a in ChunkPool::free_all_but(unsigned long) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2926 0x00007f095592af28 in ChunkPool::clean() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2927 0x00007f095592af93 in ChunkPoolCleaner::task() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2928 0x00007f09566ef880 in PeriodicTask::execute_if_pending(int) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2929 0x00007f09566ef3bf in PeriodicTask::real_time_tick(int) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2930 0x00007f09563f0bf5 in WatcherThread::run() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2931 0x00007f09567299cd in Thread::call_run() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2932 0x00007f095643e109 in thread_native_entry(Thread*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2933 0x00007f095763bfd4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#2934 0x00007f09576bc66c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
(gdb) info threads 
  Id   Target Id                         Frame 
* 1    Thread 0x7f092cfff6c0 (LWP 26563) __strlen_evex () at ../sysdeps/x86_64/multiarch/strlen-evex.S:69
  2    Thread 0x7f09553206c0 (LWP 26546) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f0950024108) at ./nptl/futex-internal.c:57
  3    Thread 0x7f0957592b80 (LWP 26545) __futex_abstimed_wait_common64 (private=128, cancel=true, abstime=0x0, op=265, expected=26546, futex_word=0x7f0955320990) at ./nptl/futex-internal.c:57
  4    Thread 0x7f093cad56c0 (LWP 26550) __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f09500f8390) at ./nptl/futex-internal.c:57
  5    Thread 0x7f092d2ff6c0 (LWP 26560) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x7f092d2fec80, op=137, expected=0, futex_word=0x7f09500220c8) at ./nptl/futex-internal.c:57
  6    Thread 0x7f093c6e46c0 (LWP 26552) 0x00007f095644fc4b in strcmp_local(char const*, char const*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
  7    Thread 0x7f093d1df6c0 (LWP 26547) __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f09500729b0) at ./nptl/futex-internal.c:57
  8    Thread 0x7f093d0dd6c0 (LWP 26548) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f0950020db8) at ./nptl/futex-internal.c:57
  9    Thread 0x7f092d0ff6c0 (LWP 26562) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f0950022b88) at ./nptl/futex-internal.c:57
  10   Thread 0x7f093cfdb6c0 (LWP 26549) __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f0950085250) at ./nptl/futex-internal.c:57
  11   Thread 0x7f093c5e26c0 (LWP 26553) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f0950024e38) at ./nptl/futex-internal.c:57
  12   Thread 0x7f093c2e26c0 (LWP 26556) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f0950022a58) at ./nptl/futex-internal.c:57
  13   Thread 0x7f08dfbff6c0 (LWP 26599) __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f09500729b0) at ./nptl/futex-internal.c:57
  14   Thread 0x7f08dfdff6c0 (LWP 26573) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x7f08dfdfe580, op=137, expected=0, futex_word=0x7f0950882d80) at ./nptl/futex-internal.c:57
  15   Thread 0x7f093c9d36c0 (LWP 26551) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f0950020ee8) at ./nptl/futex-internal.c:57
  16   Thread 0x7f092d1ff6c0 (LWP 26561) syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  17   Thread 0x7f093c4e26c0 (LWP 26554) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f09502a4584) at ./nptl/futex-internal.c:57
  18   Thread 0x7f092d7ff6c0 (LWP 26558) syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  19   Thread 0x7f093c3e26c0 (LWP 26555) __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f0950015db0) at ./nptl/futex-internal.c:57
  20   Thread 0x7f092d3ff6c0 (LWP 26559) syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  21   Thread 0x7f08dfcff6c0 (LWP 26598) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f09500244ac) at ./nptl/futex-internal.c:57
  22   Thread 0x7f08dfafd6c0 (LWP 26600) __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f09500729b0) at ./nptl/futex-internal.c:57
  23   Thread 0x7f093c1e26c0 (LWP 26557) __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x7f093c1e1ca0, op=137, expected=0, futex_word=0x7f0950022928) at ./nptl/futex-internal.c:57
  24   Thread 0x7f08df0f56c0 (LWP 26601) __futex_abstimed_wait_common64 (private=<optimized out>, cancel=true, abstime=0x0, op=393, expected=0, futex_word=0x7f09500729b0) at ./nptl/futex-internal.c:57
(gdb) t 6
[Switching to thread 6 (Thread 0x7f093c6e46c0 (LWP 26552))]
#0  0x00007f095644fc4b in strcmp_local(char const*, char const*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
(gdb) bt
#0  0x00007f095644fc4b in strcmp_local(char const*, char const*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#1  0x00007f0956450404 in symtab_lookup_iterate_phdr(dl_phdr_info*, unsigned long, void*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#2  0x00007f095770211d in __GI___dl_iterate_phdr (callback=0x7f09564500a9 <symtab_lookup_iterate_phdr(dl_phdr_info*, unsigned long, void*)>, data=0x7f093c6e3590) at ./elf/dl-iteratephdr.c:74
#3  0x00007f0956450644 in symtab_lookup(char const*, void const**, unsigned int) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#4  0x00007f0956451f2c in reset_glibc() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#5  0x00007f0956452afd in linux_ifunc_reset() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#6  0x00007f095644a599 in os::Linux::checkpoint_restore(int*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#7  0x00007f095644b737 in VM_Crac::doit() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#8  0x00007f09567d269d in VM_Operation::evaluate() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#9  0x00007f0956801122 in VMThread::evaluate_operation(VM_Operation*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#10 0x00007f0956801917 in VMThread::inner_execute(VM_Operation*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#11 0x00007f0956801cc3 in VMThread::loop() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#12 0x00007f0956800cc0 in VMThread::run() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#13 0x00007f09567299cd in Thread::call_run() () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#14 0x00007f095643e109 in thread_native_entry(Thread*) () from /home/azul/azul/crac-git/build/linux-x86_64-server-slowdebug/jdk/lib/server/libjvm.so
#15 0x00007f095763bfd4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#16 0x00007f09576bc66c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```

AntonKozlov:
I think in this PR we can concentrate on CPU features, as CPU core number is a different problem, that can arise even with the same feature set.

As for ifunc change and Watcher Thread problems, this won't happen if we re-execute with GLIBC_TUNABLES. And re-execute will also resolve the maintainability concern.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -615,6 +615,41 @@ class VM_Version_StubGenerator: public StubCodeGenerator {
...
+      | CPU_SSE | CPU_SSE2 // enabled in 'gcc -Q --help=target', used by OpenJDK
+      | CPU_FXSR // enabled in 'gcc -Q --help=target', unused by OpenJDK
+      | CPU_MMX // enabled in 'gcc -Q --help=target', used only by 32-bit x86 OpenJDK
+      | CPU_TSC // not used by gcc, used by OpenJDK
+      | CPU_CX8 // gcc detects it to set cpu "pentium" (=32-bit only), used by OpenJDK
+      | CPU_CMOV // gcc detects it to set cpu "pentiumpro" (=32-bit only), used by OpenJDK
+      | CPU_FLUSH // ="clflush" in cpuinfo, not used by gcc, required by OpenJDK
+#endif
+    ;
+  }
```
DanHeidinga, line 637: +1 to this part!  Getting a meaningful default for "generic" and the rationale for why it was chosen into the code is right overall approach.

jankratochvil:
> I think in this PR we can concentrate on CPU features, as CPU core number is a different problem, that can arise even with the same feature set.

I could split the patch but it is not testable/usable without the CPU count fix/hack.

But I am now preparing the IFUNC patch for glibc upstreaming, whether it will be accepted or not.

As long as you do not want a temporary solution in CRaC we can suspend this patch until glibc upstreaming gets resolved.

rvansa:
Hi @jankratochvil , the checks still show some trouble with whitespaces (tabs). Could you also merge in the current `crac` branch - looks like this does not use the workflow including CRaC tests.

jankratochvil:
Here is a variant which should be compatible with any glibc version using: **GLIBC_TUNABLES=:[glibc.cpu.hwcaps](https://www.gnu.org/software/libc/manual/html_node/Hardware-Capability-Tunables.html)=...**
> `./build/linux-x86_64-server-slowdebug/jdk/bin/java -XX:CPUFeatures=generic -XX:+ShowCPUFeatures Hello.java`

- On older glibcs not supporting macro `CPU_FEATURE_ACTIVE` the disabling of glibc features has no effect (and it may crash the migration even after using `-XX:CPUFeatures=generic`).
- Also newers glibcs than what the OpenJDK/CRaC sources support it may crashes due to some feature not disabled by OpenJDK/CRaC.
- I have setup (for me) tracking of [sysdeps/x86/sys/platform/x86.h](https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=sysdeps/x86/sys/platform/x86.h;hb=HEAD) and [sysdeps/x86/bits/platform/x86.h](https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=sysdeps/x86/bits/platform/x86.h;hb=HEAD).

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -615,6 +622,412 @@ class VM_Version_StubGenerator: public StubCodeGenerator {
...
+            (_glibc_features & GLIBC_CMPXCHG16) &&
+            (_glibc_features & GLIBC_LAHFSAHF)) {
+          // glibc: if (CPU_FEATURE_USABLE_P (cpu_features, AVX)
+          // glibc:     && CPU_FEATURE_USABLE_P (cpu_features, AVX2)
+          // glibc:     && CPU_FEATURE_USABLE_P (cpu_features, BMI1)
+          // glibc:     && CPU_FEATURE_USABLE_P (cpu_features, BMI2)
+          // glibc:     && CPU_FEATURE_USABLE_P (cpu_features, F16C)
+          // glibc:     && CPU_FEATURE_USABLE_P (cpu_features, FMA)
+          // glibc:     && CPU_FEATURE_USABLE_P (cpu_features, LZCNT)
+          // glibc:     && CPU_FEATURE_USABLE_P (cpu_features, MOVBE))
```
AntonKozlov, line 731: SKARA complians on this line

AntonKozlov (commented):
> * On older glibcs not supporting macro `CPU_FEATURE_ACTIVE` the disabling of glibc features has no effect (and it may crash the migration even after using `-XX:CPUFeatures=generic`).

What is GLIBC version supporting the flag? We're used to build JDK on some older platform and assume that will work on every newer platform.

And it turns out on my platform used for the builds the option is not supported.

Since it's required to specify TUNABLES in the text form, can we just define needed options names?

I'll continue reviewing this PR.

src/hotspot/cpu/x86/vm_version_x86.cpp
AntonKozlov: Hmm, I still see an error: https://github.com/openjdk/crac/pull/41/checks?check_run_id=13518791480

src/hotspot/share/runtime/stubCodeGenerator.cpp
```diff
@@ -57,6 +57,11 @@ void StubCodeDesc::freeze() {
   _frozen = true;
 }
 
+void StubCodeDesc::thaw() {
+  assert(_frozen, "repeated thaw operation");
+  _frozen = false;
```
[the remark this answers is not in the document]
jankratochvil: I get during restore:
```
#  Internal Error (../../src/hotspot/share/runtime/stubCodeGenerator.hpp:72), pid=12265, tid=12273
#  assert(!_frozen) failed: no modifications allowed
```
Did you really use a `*debug` build (and not `release` build)? The crash above has been generated on `slowdebug`.

jankratochvil:
> What is GLIBC version supporting the flag? We're used to build JDK on some older platform and assume that will work on every newer platform.
> 
> And it turns out on my platform used for the builds the option is not supported.

That should be fixed now.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -636,12 +1252,50 @@ void VM_Version::get_processor_features() {
...
+        assert(buf[res] == ',', "unexpeced VM_Version::insert_features_names separator instead of ','");
+        buf[res] = '=';
+      }
+      vm_exit_during_initialization(buf);
+    }
+          _features =   CPUFeatures_x64;
+    _glibc_features = GLIBCFeatures_x64;
+  }
+
+  glibc_not_using((CPU_MAX - 1) & ~_features, (GLIBC_MAX - 1) & ~_glibc_features);
```
AntonKozlov, line 1297: Apparently this triggers re-exec even with default command line parameters, which does not look expected, see below.

Also, we need a way to disable CPU feature reduction as a workaround to possible bugs in the implementation.

```
$ ./build/linux-x86_64-server-release/images/jdk/bin/java -XX:+ShowCPUFeatures -version
This machine's CPU features are: -XX:CPUFeatures=0x61805fdfbff,0x1e6
Re-exec of java with new environment variable: GLIBC_TUNABLES=:glibc.cpu.hwcaps=,-RTM,-AVX512F,-AVX512CD,-AVX512BW,-AVX512DQ,-AVX512ER,-AVX512PF,-AVX512VL,-IBT,-FMA4,-SHSTK
This machine's CPU features are: -XX:CPUFeatures=0x61805fdfbff,0x1e6
Environment variable already set, both glibc CPU_FEATURE_ACTIVE and ld.so --list-diagnostics are unavailable - re-exec suppressed: GLIBC_TUNABLES=:glibc.cpu.hwcaps=,-RTM,-AVX512F,-AVX512CD,-AVX512BW,-AVX512DQ,-AVX512ER,-AVX512PF,-AVX512VL,-IBT,-FMA4,-SHSTK
openjdk version "17-internal" 2021-09-14
OpenJDK Runtime Environment (build 17-internal+0-adhoc..crac)
OpenJDK 64-Bit Server VM (build 17-internal+0-adhoc..crac, mixed mode, sharing)
```

jankratochvil: I have remembered that was intentional for more safety of what jdk vs. glibc detect from CPU. I agree it is too expensive to do a jdk re-exec each time. The behavior you have seen can be now achieved by `-XX:CPUFeatures=verify`. Usefulness of this new option may not be too big.

To disable everything around the CPU features is now possible by `-XX:CPUFeatures=ignore`. I agree it may be useful if the glibc interface for these CPU features gets somehow incompatible.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -1292,9 +1313,26 @@ void VM_Version::get_processor_features() {
     }
           _features =   CPUFeatures_x64;
     _glibc_features = GLIBCFeatures_x64;
+
+    if (ShowCPUFeatures) {
+      if (ignore_glibc_not_using) {
+	tty->print_cr("CPU features are being kept intact as requested by -XX:CPUFeatures=ignore");
```
AntonKozlov, line 1319: Could you check whitespace error https://github.com/openjdk/crac/pull/41/checks?check_run_id=13867663789

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -626,10 +626,27 @@ class VM_Version_StubGenerator: public StubCodeGenerator {
...
+# elif INCLUDE_LD_SO_LIST_DIAGNOSTICS
+                  "ld.so --list-diagnostics"
+# endif // INCLUDE_LD_SO_LIST_DIAGNOSTICS
+                  " feature is available");
+#endif // INCLUDE_CPU_FEATURE_ACTIVE || INCLUDE_LD_SO_LIST_DIAGNOSTICS
     return Abstract_VM_Version::features();
   }
+  if (strcmp(ccstr, "ignore") == 0) {
+    return _features;
+  }
```
AntonKozlov, line 648: Should not be here `ignore_glibc_not_using = true` ? I cannot see a place where ignore_glibc_not_using is set to true, could you point?

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -5952,6 +5957,10 @@ static int checkpoint_restore(int *shmid) {
   } while (sig == -1 && errno == EINTR);
   assert(sig == RESTORE_SIGNAL, "got what requested");
 
+  initialize_processor_count();
+  if (_cpu_to_node != NULL)
+    rebuild_cpu_to_node_map();
```
AntonKozlov, line 5962: It seems the only place where number of processors is updated, so it's not clear how safe the operation. Please add a debug option for processor count update, I propose that to be disabled by default.

src/hotspot/cpu/x86/vm_version_x86.cpp
AntonKozlov: > I have remembered that was intentional for more safety of what jdk vs. glibc detect from CPU.

Could you elaborate a bit about the safety? In JDK we should have the same algorithm for determing glibc features. Do you mean the situation when JDK and GLIBC feature detection have diverged, e.g. due to update in the glibc?

src/hotspot/os_cpu/linux_x86/os_linux_x86.cpp
```diff
@@ -215,6 +215,13 @@ bool PosixSignals::pd_hotspot_signal_handler(int sig, siginfo_t* info,
   address pc          = NULL;
 
   //%note os_trap_1
+  if (info != NULL && uc != NULL && thread == NULL && sig == SIGSEGV) {
+    pc = (address) os::Posix::ucontext_get_pc(uc);
+    if ((sig == SIGSEGV) && VM_Version::is_cpuinfo_segv_addr(pc)) {
+      // Verify that OS save/restore AVX registers.
+      stub = VM_Version::cpuinfo_cont_addr();
+    }
+  } else
```
AntonKozlov, line 224: Why do we need this block? Is not this duplicating logic from line 251?

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -636,12 +1271,69 @@ void VM_Version::get_processor_features() {
...
+#if !INCLUDE_CPU_FEATURE_ACTIVE && !INCLUDE_LD_SO_LIST_DIAGNOSTICS
+  if (!verify_glibc_by_reexec) {
+          features_expected =       _features;
+    glibc_features_expected = _glibc_features;
+  }
+#endif
+  if (!ignore_glibc_not_using) {
+    glibc_not_using(      features_expected & ~      _features,
+		    glibc_features_expected & ~_glibc_features);
+  }
```
AntonKozlov, line 1335: I'm trying the code, and getting

```
anton@mercury:~/proj/crac$ ./build/linux-x86_64-server-release/images/jdk/bin/java -XX:+ShowCPUFeatures -XX:CPUFeatures=generic -version
This machine's CPU features are: -XX:CPUFeatures=0x61805fdfbff,0x1e6
CPU features being used are: -XX:CPUFeatures=0x200000080d7,0x0
openjdk version "17-internal" 2021-09-14
OpenJDK Runtime Environment (build 17-internal+0-adhoc..crac)
OpenJDK 64-Bit Server VM (build 17-internal+0-adhoc..crac, mixed mode, sharing)
```

I would expect re-exec, but for some reason it was not perfromed?

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: I have removed the option `-XX:CPUFeatures=verify`. It had some effect only on old glibcs and even there it did only check jdk against jdk's own logic. Not against glibc.

Diversion of JDK and GLIBC feature detection is tested by default but only if either `CPU_FEATURE_ACTIVE` or `ld.so --list-diagnostics` are available

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: Sorry, I have provided some local reconfigurations so that it hopefully should not happen again.

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: That was a regression/bug, thanks, fixed.

src/hotspot/os_cpu/linux_x86/os_linux_x86.cpp
jankratochvil: It was needed for some previous variant of the patch and it is a leftover. Removed. Sorry for not self-reviewing this version of my patch.

jankratochvil: It is in fact needed: https://github.com/openjdk/crac/blob/08c2c0f2b140b1011fec39a1a83037df8e59577d/src/hotspot/os_cpu/linux_x86/os_linux_x86.cpp#L217
During CRaC restore there is somehow no Java thread created yet that time. I do not understand this much so I left it there as working but maybe some Java thread should be initialized earlier?

src/hotspot/os/linux/os_linux.cpp
jankratochvil: OK, moved under `-XX:+CRaCCPUCountInit`. It was crashing (asserting IIRC) for me when moving between 4 and 20 vCPUs but I was unable to reproduce it now.

AntonKozlov (commented):
I meet with the same error as reported by GHA

https://github.com/jankratochvil/crac/actions/runs/5200182180/jobs/9378587527

Replacing vm_exit_during_initialization() with fatal() at least provides the hs_err and the stack trace.

```
Stack: [0x00007f80d6900000,0x00007f80d6a00000],  sp=0x00007f80d69fd760,  free space=1013k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0x191bad4]  VM_Version::glibc_not_using(unsigned long, unsigned long)+0x414
V  [libjvm.so+0x191d3a2]  VM_Version::CPUFeatures_init()+0x1d2
V  [libjvm.so+0x191d67f]  VM_Version::get_processor_features()+0xef
V  [libjvm.so+0x1921f16]  VM_Version::initialize_features()+0x546
V  [libjvm.so+0x1922359]  VM_Version::initialize()+0x9
V  [libjvm.so+0x1916756]  VM_Version_init()+0x26
V  [libjvm.so+0xd30b4c]  init_globals()+0x2c
V  [libjvm.so+0x18110e6]  Threads::create_vm(JavaVMInitArgs*, bool*)+0x326
V  [libjvm.so+0xeb86d9]  JNI_CreateJavaVM+0x69
C  [libjli.so+0x3da4]  JavaMain+0x94
C  [libjli.so+0x7719]  ThreadJavaMain+0x9
```


```
anton@mercury:~/proj/crac$ git diff
diff --git a/src/hotspot/cpu/x86/vm_version_x86.cpp b/src/hotspot/cpu/x86/vm_version_x86.cpp
index a6a04458319..58249612750 100644
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -1225,10 +1225,9 @@ void VM_Version::glibc_not_using(uint64_t excessive_CPU, uint64_t excessive_GLIB
 #ifdef ASSERT
 #define CHECK_KIND(kind) do {                                                                                   \
     if (PASTE_TOKENS(disable_handled_, kind) != PASTE_TOKENS(excessive_handled_, kind)) {                       \
-      jio_snprintf(errbuf, sizeof(errbuf),                                                                      \
+      fatal( \
                    "internal error: Unsupported disabling of " STR(kind) "_* 0x%" PRIx64 " != used 0x%" PRIx64, \
                    PASTE_TOKENS(disable_handled_, kind), PASTE_TOKENS(excessive_handled_, kind));               \
-      vm_exit_during_initialization(errbuf);                                                                    \
     }                                                                                                           \
   } while (0)
   CHECK_KIND(CPU  );
```

I'm not 100% sure fatal() is correct in that state, so I propose a vararg macro/function that expands to fatal(...), which can easily be replaced with something different.

jankratochvil:
> I meet with the same error as reported by GHA

I have fixed that. I am mostly ignoring GHA as there are too many false positives. The problem was I was using only _release_ builds for some time where it did not reproduce.

> Replacing vm_exit_during_initialization() with fatal() at least provides the hs_err and the stack trace.

I did not want a stack trace. Those are normal error messages, stack trace would be rather confusing. Please request it again if you really mean it.

> I'm not 100% sure fatal() is correct in that state, so I propose a vararg macro/function that expands to fatal(...), which can easily be replaced with something different.

I have found there exists **err_msg()** so I have used it. That **jio_snprintf()**+**vm_exit_during_initialization()** are also used at some hotspot places which is why I used that originally like a copycat. Addressed now by: https://github.com/openjdk/crac/pull/41/commits/9d2298ba9d89c1a2f7e5143153c8e280da6dccec

jankratochvil:
> Can we just avoid code-generatation on repeated calls of initialize_features? And drop thaw() completely?

I have done so now. The original idea was calling it so that if a snapshot is done on old CPU and restored on a new CPU the hotspot will still benefit from new CPU for newly compiled JIT code. But it is true with #57 it may become more questionable what is correct to do.

Also the last variant of this patch has no longer been benefiting from this code-regeneration anymore.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -1900,7 +2579,46 @@ void VM_Version::initialize() {
...
+
+  uint64_t       features_saved =       _features;
+  uint64_t glibc_features_saved = _glibc_features;
+
+        _features =   CPUFeatures_x64;
+  _glibc_features = GLIBCFeatures_x64;
+
+  if (ShowCPUFeatures)
+    nonlibc_tty_print_using_features_cr();
```
AntonKozlov, line 2608: Can be a regular printing function call?

AntonKozlov (approved):
I could not spot anything, considering the size of the patch :)

LGTM!

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -1885,6 +2484,86 @@ void VM_Version::check_virtualizations() {
...
+#define SUPPORTS(x)                                           \
+  if (x##_saved != Abstract_VM_Version::x()) {                \
+    supports_exit( #x , Abstract_VM_Version::x(), x##_saved); \
+  }
+  SUPPORTS_SET
+#undef SUPPORTS
+#undef SUPPORTS_SET
+
+  if (ShowCPUFeatures)
+    nonlibc_tty_print_using_features_cr();
```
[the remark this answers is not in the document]
jankratochvil: In fact it was already using a regular printing function (`print_cr()`). It was just a wrongly named function, renamed it to: `print_using_features_cr()`

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: In fact it was already using a regular printing function (`print_cr()`). It was just a wrongly named function, renamed it to: `print_using_features_cr()`

AntonKozlov (approved):

jankratochvil:
It should get a reapproval so that the patch does not remove the `altstack` fix.

AntonKozlov (approved):

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -615,8 +625,585 @@ class VM_Version_StubGenerator: public StubCodeGenerator {
...
+      // FPU is always present on i686+: (_features & CPU_FPU) &&
+      (_features & CPU_SSE2))
+    // These cannot be disabled by GLIBC_TUNABLES.
+    if (excessive_CPU & (CPU_FXSR | CPU_MMX | CPU_SSE)) {
+      assert(!(excessive_CPU & CPU_SSE2), "CPU_SSE2 in both _features and excessive_CPU cannot happen");
+      // FIXME: The choice should be based on glibc impact, not the feature age.
+      // CX8 is i586+, CMOV is i686+ 1995+, SSE2 is 2000+
+      excessive_CPU |= CPU_SSE2;
+    }
+    if ((_features & CPU_FXSR) &&
```
AntonKozlov, line 942: My compiler started to complain on this, and it looks valid:

```
src/hotspot/cpu/x86/vm_version_x86.cpp: In static member function 'static void VM_Version::glibc_not_using(uint64_t, uint64_t)':
src/hotspot/cpu/x86/vm_version_x86.cpp:1118:3: error: this 'if' clause does not guard... [-Werror=misleading-indentation]
   if ((_features & CPU_CMOV) &&
   ^~
src/hotspot/cpu/x86/vm_version_x86.cpp:1129:5: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
     if ((_features & CPU_FXSR) &&
     ^~
cc1plus: all warnings being treated as errors
```

@jankratochvil it indeed looks like a missing bracket, is it a bug?

jankratochvil: Yes, sorry, #94.

PR #42, opened 2023-01-31
RestoreEnvironmentTest refactoring

wkia:
The test was extended with the example to illustrate the scenario when an user don't like to propagate some of environment variables into a restored process (see `RESTORE_ENVIRONMENT_TEST_VAR2` in `RestoreEnvironmentTest.sh`).


DanHeidinga:
I'll quote my response from the (closed) PR #30:
> The current CRaC approach still allows overriding env vars, which is something Java (and developers, and library developers) doesn't expect to happen so they haven't written their code to be aware these values can change. Changing existing env vars on restore will just lead to inconsistent systems.

Full response in https://github.com/openjdk/crac/pull/30#issuecomment-1412545649

PR #38, opened 2022-12-09
Print peristent JAR assumptions

AntonKozlov:
When an application creates URLClassLoader with the JAR, the JAR file will be assumed persistent, that is, it should be available on the restore. To make this more evident, the PR adds reporting for such JARs.

I'm not a big fan of persistent files and the implications, I think it may be done better, but while they are there, better to make them more usable.

---------
### Progress
- [x] Change must not contain extraneous whitespace

### Reviewers
 * [Dan Heidinga](https://openjdk.org/census#heidinga) (@DanHeidinga - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac pull/38/head:pull/38` \
`$ git checkout pull/38`

Update a local copy of the PR: \
`$ git checkout pull/38` \
`$ git pull https://git.openjdk.org/crac pull/38/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 38`

View PR using the GUI difftool: \
`$ git pr show -t 38`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/crac/pull/38.diff">https://git.openjdk.org/crac/pull/38.diff</a>

</details>

PR #43, opened 2023-01-31
Improved open file descriptors tracking

rvansa:
Tracks `java.io.FileDescriptor` instances as CRaC resource; before checkpoint these are reported and if not allow-listed (e.g. as opened as standard descriptors) an exception is thrown. Further investigation can use system property `jdk.crac.collect-fd-stacktraces=true` to record origin of those file descriptors.
File descriptors claimed in Java code are passed to native; native code checks all open file descriptors and reports error if there's an unexpected FD that is not included in the list passed previously.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/43/head:pull/43` \
`$ git checkout pull/43`

`$ git checkout pull/43` \
`$ git pull https://git.openjdk.org/crac.git pull/43/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 43`

`$ git pr show -t 43`

</details>

<a href="https://git.openjdk.org/crac/pull/43.diff">https://git.openjdk.org/crac/pull/43.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/43#issuecomment-1410120990)


rvansa:
@AntonKozlov I've dropped the native FDs tracking.

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -59,15 +61,35 @@ public final class FileDescriptor {
     private List<Closeable> otherParents;
     private boolean closed;
 
+    /**
+     * Called by FileDispatcherImpl when the file descriptor is about to be closed natively.
+     */
+    public void markClosedByNIO() {
```
AntonKozlov, line 67: Can we avoid changing the public interface? This is not a method we'd like users to call

src/java.base/linux/classes/sun/nio/ch/EPollSelectorImpl.java
```diff
@@ -110,7 +111,10 @@ private void initFDs() throws IOException {
 
         try {
             this.eventfd = new EventFD();
-            IOUtil.configureBlocking(IOUtil.newFD(eventfd.efd()), false);
+            FileDescriptor fd = IOUtil.newFD(eventfd.efd());
+            // This FileDescriptor is a one-time use, the actual FD will be closed from EventFD
+            fd.markClosedByNIO();
+            IOUtil.configureBlocking(fd, false);
```
AntonKozlov, line 117: The eventfd was actually closed on line 149, so it's incorrect to mark it is closed here. I think everything was fine before this chunk

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -59,15 +61,35 @@ public final class FileDescriptor {
...
+        resource.closedByNIO = true;
+    }
+
     class Resource implements jdk.internal.crac.JDKResource {
+        private static final boolean COLLECT_FD_STACKTRACES =
+                GetBooleanAction.privilegedGetProperty(JDKContext.COLLECT_FD_STACKTRACES_PROPERTY);
+
+        private boolean closedByNIO;
```
AntonKozlov, line 75: I think it may perfectly reside in FileDesriptor, as it looks like a bug for me that FileDispatcherImpl does not change valid() status of the FileDesriptor. Or even make FileDispatcherImpl to reset FileDescriptor.fd? Are there reasons not to do that?

src/java.base/share/classes/java/io/RandomAccessFile.java
```diff
@@ -93,7 +93,12 @@ private class Resource implements JDKResource {
         @Override
         public void beforeCheckpoint(Context<? extends jdk.crac.Resource> context) throws Exception {
             if (Core.getJDKContext().claimFdWeak(fd, this)) {
-                throw new CheckpointOpenFileException(this.toString());
+                if (Core.getJDKContext().matchClasspath(path)) {
+                    // Files on the classpath are considered persistent, exception is not thrown
+                    return;
+                }
+                throw new CheckpointOpenFileException("RandomAccessFile " + path + " left open." + JDKContext.COLLECT_FD_STACKTRACES_HINT,
```
AntonKozlov, line 100: Stack-traces are good idea! 

Here I would suggest add file desriptor number as well.

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -59,15 +61,35 @@ public final class FileDescriptor {
...
+    /**
+     * Called by FileDispatcherImpl when the file descriptor is about to be closed natively.
+     */
+    public void markClosedByNIO() {
+        resource.closedByNIO = true;
+    }
+
     class Resource implements jdk.internal.crac.JDKResource {
+        private static final boolean COLLECT_FD_STACKTRACES =
+                GetBooleanAction.privilegedGetProperty(JDKContext.COLLECT_FD_STACKTRACES_PROPERTY);
```
AntonKozlov, line 73: Should not be moved to JDKContext or jdk.internal.Core ? To avoid any possible user to get the property value the similar way

src/java.base/share/classes/java/io/RandomAccessFile.java
```diff
@@ -80,6 +85,36 @@ public class RandomAccessFile implements DataOutput, DataInput, Closeable {
...
 
+    private class Resource implements JDKResource {
+        Resource() {
+            Core.getJDKContext().register(this);
+        }
+
+        @Override
+        public void beforeCheckpoint(Context<? extends jdk.crac.Resource> context) throws Exception {
+            if (Core.getJDKContext().claimFdWeak(fd, this)) {
+                if (Core.getJDKContext().matchClasspath(path)) {
```
AntonKozlov, line 96: Actually it would be more natural to match with PersistentJarFile who claimed their FDs already. So we'll be able to handle dynamically loaded JARs that are not a part of the `java.class.path`

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -53,6 +53,8 @@
 #include "runtime/init.hpp"
 #include "runtime/java.hpp"
 #include "runtime/javaCalls.hpp"
+#include "runtime/jniHandles.hpp"
+#include "runtime/jniHandles.inline.hpp"
```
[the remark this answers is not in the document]
rvansa: OK, that comes from your commits, I'll clean it up.

src/java.base/linux/classes/sun/nio/ch/EPollSelectorImpl.java
rvansa: The native descriptor is closed, but the FileDescriptor instance is created and forgotten. If the GC doesn't remove it before checkpoint it may be reported as a FileDescriptor leak. I think I saw that happening.

src/java.base/share/classes/java/io/FileDescriptor.java
rvansa: I haven't changed that to minimize the impact of changes; I still assume at one point it made sense to close the native FD without going through FileDescriptor. The reply I got on nio-dev was this:

> In the java.io APIs, FIS/FOS/RAF define getFD methods so the FileDescriptor can be obtained by applications. This is problematic for a bunch of reasons but for the discussion here, setting it to -1 is needed for the FileDescriptor::valid. Another point is that java.io pre-dates the more robust async close mechanism that is NIO and setting it to -1 helps avoid trying to use a file descriptor that has been closed and/or recycled.
> In the NIO area, FileDescriptor objects do not leak to applications. The only case where it is necessary to set to -1 is on Windows with sockets and this is because there is no equivalent of dup2 there. Another point here is that FileDescriptors aren't really needed as the channels implementation just need the raw file descriptor or handle. 

Having that in FileDescriptor seemed more consolidated but if you prefer it that way I won't object. So, should I do that or risk breaking stuff by setting fd to -1?

src/java.base/share/classes/java/io/FileDescriptor.java
rvansa: I've checked if it would be possible to extend FileDescriptor and put this method there, but there's several non-trivial places where this would have to be changed (e.g. in `sun.nio.ch.InheritedChannel` the FD is constructed using a non-public ctor via reflection). So we can make this package-private and use reflection, or probablz a better way would be through `JavaIOFileDescriptorAccess`.

src/java.base/share/classes/java/io/FileDescriptor.java
rvansa: Not sure what you're trying to prevent, user code reading a property? The field here is private.

src/java.base/share/classes/java/io/RandomAccessFile.java
rvansa: This case is trying to prevent a situation when the FD is **not** opened using `PersistentJarFile`, as Spring Boot uses own implementation.

src/java.base/share/classes/java/io/FileDescriptor.java
AntonKozlov: My note is minor, I'm trying to save future users of COLLECT_FD_STACKTRACES_PROPERTY from readingthe property, instead the property can be used, parsed in JDKContext 
```
class JDKContext {
        static final boolean COLLECT_FD_STACKTRACES =
               GetBooleanAction.privilegedGetProperty(JDKContext.COLLECT_FD_STACKTRACES_PROPERTY);
}

// In this class: just use JDKContext.COLLECT_FD_STACKTRACES without GetProperty
```

src/java.base/share/classes/java/io/FileDescriptor.java
AntonKozlov: From the reply [1] I understand that there was no reason to keep track of the native fd/valid() status in FileDescriptors used by NIO, so that was not done. And for me the status of the FD is closer to FD itself, but I don't have a strong preference.

[1] https://mail.openjdk.org/pipermail/nio-dev/2023-January/013075.html, just for bookkeeping

src/java.base/share/classes/java/io/RandomAccessFile.java
AntonKozlov: Indeed! This is a required transition from the old logic in hotspot, which ignored all files (by path) specified in classpath. 

I was thinking about another use-case, when we have an instance of `PersistentJarFile` (which allows underlying fd), and RandomAccessFile with another underlying fd, but reffering to the same path on file system.

src/hotspot/os/linux/os_linux.cpp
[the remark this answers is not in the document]
rvansa: Turns out this is needed, doesn't compile without.

rvansa:
@AntonKozlov Updated.

rvansa:
@AntonKozlov Bump, I think the comments were handled.

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -125,7 +128,17 @@ private static void checkpointRestore1(long jcmdStream) throws
...
 
-        final Object[] bundle = checkpointRestore0(checkpointException != null, jcmdStream);
+        JDKContext jdkContext = jdk.internal.crac.Core.getJDKContext();
+        List<Map.Entry<Integer, Object>> claimedPairs = jdkContext.getClaimedFds().entrySet().stream().toList();
+        int[] fdArr = new int[claimedPairs.size()];
+        Object[] objArr = new Object[claimedPairs.size()];
+        for (int i = 0; i < claimedPairs.size(); ++i) {
+            fdArr[i] = claimedPairs.get(i).getKey();
+            objArr[i] = claimedPairs.get(i).getValue();
+            System.out.printf("%d %s\n", fdArr[i], objArr[i]);
```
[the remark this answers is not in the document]
rvansa: I am not so sure about that; the `objArr` is later ignored by the `VM_Crac` so I thought this it the place that should report open FDs, even though claimed by Java.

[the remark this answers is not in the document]
AntonKozlov: I see, reporting here could be useful, but then it'd better use a Logger or something like that. BTW this was implemented in ad-hoc manner in https://github.com/openjdk/crac/pull/38/files, these should be unified.

rvansa:
@AntonKozlov I had to add the changes for locking of resources registered during beforeCheckpoint already into this commit; when running the test I've ran into deadlocks.

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -297,6 +368,24 @@ synchronized void close() throws IOException {
...
 
+    private synchronized void beforeCheckpoint() throws CheckpointOpenFileException {
+        if (valid()) {
+            JDKContext ctx = jdk.internal.crac.Core.getJDKContext();
+            if (ctx.claimFdWeak(this, this)) {
+                String msg = "FileDescriptor " + this.fd + " left open. ";
+                if (!JDKContext.Properties.COLLECT_FD_STACKTRACES) {
+                    msg += JDKContext.COLLECT_FD_STACKTRACES_HINT;
+                }
+                throw new CheckpointOpenFileException(msg, resource.stackTraceHolder);
```
AntonKozlov, line 379: Now, the message looks like:

```
	Suppressed: jdk.crac.impl.CheckpointOpenFileException: FileDescriptor 323 left open.
		at java.base/java.io.FileDescriptor.beforeCheckpoint(FileDescriptor.java:379)
		at java.base/java.io.FileDescriptor$Resource.beforeCheckpoint(FileDescriptor.java:80)
		at java.base/jdk.crac.impl.AbstractContextImpl.runBeforeCheckpoint(AbstractContextImpl.java:114)
		at java.base/jdk.crac.impl.AbstractContextImpl.beforeCheckpoint(AbstractContextImpl.java:83)
		at java.base/jdk.internal.crac.JDKContext.beforeCheckpoint(JDKContext.java:85)
		at java.base/jdk.crac.impl.AbstractContextImpl.runBeforeCheckpoint(AbstractContextImpl.java:114)
		at java.base/jdk.crac.impl.AbstractContextImpl.beforeCheckpoint(AbstractContextImpl.java:83)
		at java.base/jdk.crac.Core.checkpointRestore1(Core.java:121)
		... 2 more
	Caused by: java.lang.Exception: This file descriptor was created here
		at java.base/java.io.FileDescriptor$Resource.<init>(FileDescriptor.java:71)
		at java.base/java.io.FileDescriptor.<init>(FileDescriptor.java:100)
		at java.base/sun.nio.ch.IOUtil.newFD(IOUtil.java:544)
		at java.base/sun.nio.ch.Net.socket(Net.java:524)
		at java.base/sun.nio.ch.SocketChannelImpl.<init>(SocketChannelImpl.java:146)
		at java.base/sun.nio.ch.SocketChannelImpl.<init>(SocketChannelImpl.java:129)
		at java.base/sun.nio.ch.SelectorProviderImpl.openSocketChannel(SelectorProviderImpl.java:77)
		at java.base/java.nio.channels.SocketChannel.open(SocketChannel.java:192)
```

While in the past the FD (which is actually a socket, not a file) was reported by the VM like:

```
	Suppressed: jdk.crac.impl.CheckpointOpenSocketException: tcp localAddr 0.0.0.0 localPort 8080 remoteAddr 0.0.0.0 remotePort 0
		at java.base/jdk.crac.Core.translateJVMExceptions(Core.java:92)
		at java.base/jdk.crac.Core.checkpointRestore1(Core.java:157)
```

We miss the fact this is a socket now, as well as details (although the stack trace is very useful!).

We can ask the VM for the FD type and details. To so we'll report report the details as we used to, by reusing the existing code providing the details, that should not be very hard.

src/java.base/share/classes/jdk/crac/LoggerContainer.java
```diff
@@ -0,0 +1,29 @@
+package jdk.crac;
+
+/**
+ * Some classes that could use logging are initialized early during the boot
+ * and keeping the logger in static final field there could cause problems
+ * (e.g. recursion when service-loading logger implementation).
+ * Therefore, we isolate the logger into a subclass and initialize lazily.
+ */
+public class LoggerContainer {
```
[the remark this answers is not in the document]
rvansa: It's not package private as I meant to create one logger for all things CRaC, including logging in one of the implementation packages. I could probably move that to `jdk.crac.impl` or `jdk.internal.crac`.

src/java.base/share/classes/jdk/internal/crac/JDKContext.java
```diff
@@ -26,25 +26,96 @@
...
+                GetBooleanAction.privilegedGetProperty(JDKContext.COLLECT_FD_STACKTRACES_PROPERTY);
+    }
+
+    private WeakHashMap<FileDescriptor, Object> claimedFds;
+
+    public boolean matchClasspath(String path) {
+        Path p = Path.of(path);
+        String classpath = System.getProperty("java.class.path");
+        int index = 0;
+        while (index >= 0) {
```
rvansa, line 62: Bug: looks like the update of `index` was dropped somewhere, this loop may run infinitely...

src/java.base/unix/native/libnet/SocketImpl.c
```diff
@@ -58,3 +59,114 @@ Java_java_net_SocketCleanable_cleanupClose0(JNIEnv *env, jclass c1, jint fd)
...
+    if (getsockname(fd, &local.sa, &llen) != 0) {
+        JNU_ThrowIllegalArgumentException(env, strerror(errno));
+        return NULL;
+    } else {
+        localAddr = create_isa(env, isa_class, isa_ctor, &local);
+        if (localAddr == NULL) {
+           JNU_ThrowOutOfMemoryError(env, "java.net.InetSocketAddres");
+           return NULL;
+        }
+    }
```
AntonKozlov, line 115: No need for `} else {` here and everywhere else since the previous block has anyway terminated the function.

This will make the code more streamlined.

```suggestion
}

localAddr = create_isa(env, isa_class, isa_ctor, &local);
if (localAddr == NULL) {
    JNU_ThrowOutOfMemoryError(env, "java.net.InetSocketAddres");
    return NULL;
}
```

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6145,15 +6079,17 @@ void VM_Crac::doit() {
...
     ok = false;
 
-    char* msg = NEW_C_HEAP_ARRAY(char, strlen(details) + 1, mtInternal);
-    strcpy(msg, details);
+    const int maxinfo = 64;
+    size_t buflen = strlen(details) + maxinfo;
+    char* msg = NEW_C_HEAP_ARRAY(char, buflen, mtInternal);
+    int ilen = snprintf(msg, maxinfo, "FD fd=%d type=%s path=", i, type);
+    ilen = ilen > maxinfo ? maxinfo : ilen;
+    strncpy(msg + ilen, detailsbuf, buflen - ilen);
```
AntonKozlov, line 6092: `ilen >= maxinfo` if the output was truncated [1]

`strncpy` may leave the string unterminated if `buflen-ilen` smaller than details.

So `snprintf(msg, maxinfo, "FD fd=%d type=%s path=%s", i, type, details)` will be a safer option. 

[1] RETURN in https://linux.die.net/man/3/snprintf

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -372,7 +374,31 @@ private synchronized void beforeCheckpoint() throws CheckpointOpenFileException
...
+                        if (addresses[1] != null) {
+                            sb.append(" remote ").append(addresses[1]);
+                        } else {
+                            sb.append(" remote not bound");
+                        }
+                    }
+                    info = sb.toString();
+                } else {
+                    info = (path != null ? path : "unknown path") + " (" + (type != null ? type : "unknown") + ")";
+                }
```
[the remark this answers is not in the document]
rvansa: I am thinking about this also in a context of `newfd-policies`, where we have to record not only the type & path but also things like current offset etc.
Since this is used only in error handling, I wouldn't mind about the native transitions (or performance in general) too much. If you want, I could create a POJO to cross the native border only once. However this goes a bit against the principle that we want to handle as much as we can in Java rather than native code. That's also the reason why I opted for the formatting in Java even though it might be just as simple to do it in the native.

src/java.base/unix/native/libnet/SocketImpl.c
rvansa: The code was supposed to mirror closer the remote part where we can't do that, but OK.

rvansa:
Alright, go for it.

AntonKozlov:
Thanks for making the progress in this! And make things moving :)

PR #45, opened 2023-02-16
Close extraneous file descriptors

rvansa:
Java code cannot use inherited non-standard file descriptors; these can cause trouble when trying to checkpoint the process via CRIU, and hence we will proactively close them on boot.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac pull/45/head:pull/45` \
`$ git checkout pull/45`

`$ git checkout pull/45` \
`$ git pull https://git.openjdk.org/crac pull/45/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 45`

`$ git pr show -t 45`

</details>

<a href="https://git.openjdk.org/crac/pull/45.diff">https://git.openjdk.org/crac/pull/45.diff</a>

</details>

### DIFF
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -5750,6 +5750,9 @@ void os::Linux::vm_create_start() {
   if (!CRaCCheckpointTo) {
     return;
   }
+  if (CRaCCheckpointTo) {
+    close_extra_descriptors();
+  }
   _vm_inited_fds.initialize();
 }
 
@@ -6456,14 +6459,15 @@ void os::Linux::restore() {
   }
 }
 
+static char modules_path[JVM_MAXPATHLEN] = { '\0' };
+
 static bool is_fd_ignored(int fd, const char *path) {
-  int path_len = path ? strlen(path) : -1;
-  if (!strncmp("/proc/", path, 6) && !strcmp("/fd", path + path_len - 3)) {
-    // silently ignore us reading our own descriptors
+  if (!strcmp(modules_path, path)) {
+    // modules directory is opened al the time
     return true;
   }
 
-  const char *list = CRIgnoredFileDescriptors;
+  const char *list = CRaCIgnoredFileDescriptors;
   while (list && *list) {
     const char *end = strchr(list, ',');
     if (!end) {
@@ -6477,6 +6481,7 @@ static bool is_fd_ignored(int fd, const char *path) {
         return true;
       }
     } else { // interpret entry as path
+      int path_len = path ? strlen(path) : -1;
       if (path_len != -1 && path_len == end - list && !strncmp(path, list, end - list)) {
         log_trace(os)("CRaC not closing file descriptor %d (%s) as it is marked as ignored.", fd, path);
         return true;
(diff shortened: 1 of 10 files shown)


rvansa:
I've rebased on top of most recent crac branch but there are no comments.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2100,6 +2100,11 @@ const intx ObjectAlignmentInBytes = 8;
       "-XX:CRaCRestoreFrom and continue initialization if restore is "      \
       "unavailable")                                                        \
                                                                             \
+  product(ccstr, CRaCIgnoredFileDescriptors, NULL, "Comma-separated list "  \
```
AntonKozlov, line 2103: Should not `0,1,2` be the default option then? So it will be also possible to overwrite this default.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6456,6 +6457,66 @@ void os::Linux::restore() {
...
+    }
+  }
+  return false;
+}
+
+void os::Linux::close_extra_descriptors() {
+  if (modules_path[0] == '\0') {
+    const char* fileSep = os::file_separator();
+    jio_snprintf(modules_path, JVM_MAXPATHLEN, "%s%slib%smodules", Arguments::get_java_home(), fileSep, fileSep);
+  }
```
AntonKozlov, line 6501: A comment would be useful here. I cannot suggest a way to avoid this particular check, as modules are opened very early during argument parsing, to get possible additional arguments from the module, so let's document this for future readers of the code.

AntonKozlov (changes requested):
Few minor comments

src/hotspot/share/runtime/globals.hpp
rvansa: I didn't use `0,1,2` as if you want to ignore FD 42 you would have to write `0,1,2,42` rather than just 42. Starting with standard FDs closed is extremely non-standard, and in that case we should do something about `System.out` as well to be consistent.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6456,6 +6457,66 @@ void os::Linux::restore() {
...
+  char path[PATH_MAX];
+  struct dirent *dp;
+
+  DIR *dir = opendir("/proc/self/fd");
+  while (dp = readdir(dir)) {
+    int fd = atoi(dp->d_name);
+    if (fd > 2 && fd != dirfd(dir)) {
+      int r = readfdlink(fd, path, sizeof(path));
+      if (!is_fd_ignored(fd, r != -1 ? path : nullptr)) {
+        tty->print("CRaC closing file descriptor %d: %s\n", fd, path);
```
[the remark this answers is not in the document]
rvansa: It's meant to be more visible than trace-level: for consistency I'll make this `log_warning(os)`

rvansa:
@AntonKozlov Updated!

src/hotspot/share/runtime/globals.hpp
AntonKozlov: Agree, the current implementation makes sense.

AntonKozlov (approved):
LGTM!

PR #46, opened 2023-02-21
CRaC may exit before image dump is completed

wkia:
When running CRaC with docker, java may exit before CRIU is finished dumpring because CRIU kills the original java process, and then docker immediately exits.

It could be reproduced with a simple Java test:
```
import jdk.crac.Core;
public class Test {
    public static void main(String args[]) throws Exception {
        Core.checkpointRestore();
        System.out.println("finish");
    }
}
```
and run with docker:
`docker <docker_options> $JAVA_HOME/java -XX:CRaCCheckpointTo=./cr Test.java`

After the command above finishes, `cr/cppath` is absent in the case of failure. Or/and it will fail on restore:
`docker <docker_options> $JAVA_HOME/java -XX:CRaCRestoreFrom=./cr`

This change fixes the issue by forkin'g the main process in case of PID=1 (pid=1 means it was run with docker), and waiting for children processes are finished. This makes us sure that CRIU finalized the dump, if any. At the same time, there is no conflict with PIDs on restore, since the process being restored has PID not equal to 1, if restoring with the command above..

### DIFF
--- a/src/java.base/share/native/launcher/main.c
+++ b/src/java.base/share/native/launcher/main.c
@@ -93,6 +93,34 @@ WinMain(HINSTANCE inst, HINSTANCE previnst, LPSTR cmdline, int cmdshow)
     __initenv = _environ;
 
 #else /* JAVAW */
+
+#include <sys/wait.h>
+
+static int is_checkpoint = 0;
+
+static void parse_checkpoint(const char *arg) {
+    if (!is_checkpoint) {
+        const char *checkpoint_arg = "-XX:CRaCCheckpointTo";
+        const int len = strlen(checkpoint_arg);
+        if (0 == strncmp(arg, checkpoint_arg, len)) {
+            is_checkpoint = 1;
+        }
+    }
+}
+
+int wait_for_children() {
+    int status = 0;
+    pid_t pid;
+    do {
+        int st = 0;
+        pid = wait(&st);
+        if (!status) {
+            status = st;
+        }
+    } while (!(-1 == pid && ECHILD == errno));
+    return status;
+}
+
 JNIEXPORT int
 main(int argc, char **argv)
 {
@@ -183,6 +211,7 @@ main(int argc, char **argv)
         }
         // Iterate the rest of command line
         for (i = 1; i < argc; i++) {
+            parse_checkpoint(argv[i]);
             JLI_List argsInFile = JLI_PreprocessArg(argv[i], JNI_TRUE);
             if (NULL == argsInFile) {
                 JLI_List_add(args, JLI_StringDup(argv[i]));
@@ -201,6 +230,13 @@ main(int argc, char **argv)
         // add the NULL pointer at argv[argc]
         JLI_List_add(args, NULL);
         margv = args->elements;
+    }
+
+    if (is_checkpoint && 1 == getpid()) {
+        if (0 < fork()) {
+            const int status = wait_for_children();
+            exit(status);
+        }
     }
 #endif /* WIN32 */
     return JLI_Launch(margc, margv,


src/java.base/share/native/launcher/main.c
```diff
@@ -202,6 +231,13 @@ main(int argc, char **argv)
         JLI_List_add(args, NULL);
         margv = args->elements;
     }
+
+    if (is_checkpoint && 1 == getpid()) {
+        if (0 < fork()) {
```
DanHeidinga, line 236: Not an expert in this area, but don't we also need to handle `SIGTERM` appropriately by forwarding it to the child process if the forking receives it?

wkia: @DanHeidinga 
Do you mean "we need to terminate children if the parent process is terminated"?
It seems like termination of a process with PID=1 running inside a container will stop a container run, so I guess children will be terminated also. I'd appreciate any reproducible scenario, in case I'm mistaken.

AntonKozlov: Indeed. Since this process was supposed to be the java process, it should be a proxy a good proxy for the real java process. E.g. it should forward SIGTERM and other signals (SIGQUIT to print stack trace, etc). Although this proxy should not necessarily be a good init process.

We had the same problem in restore, this code probably will be useful https://github.com/openjdk/crac/blob/crac/src/java.base/unix/native/criuengine/criuengine.c#L263

src/java.base/share/native/launcher/main.c
```diff
@@ -93,6 +93,64 @@ WinMain(HINSTANCE inst, HINSTANCE previnst, LPSTR cmdline, int cmdshow)
...
+}
+
+static int wait_for_children() {
+    int status = 0;
+    pid_t pid;
+    do {
+        int st = 0;
+        pid = wait(&st);
+        if (!status && 0 < pid && WIFEXITED(st)) {
+            status = WEXITSTATUS(st);
```
AntonKozlov, line 118: Some `WEXITSTATUS(st)` can end up 0, so the a next exited child can override this status.

I think here we should watch for the status of the g_child_pid, and ignoring statuses of other childs that are not our immediate java process child.

AntonKozlov (changes requested):

src/java.base/share/native/launcher/main.c
DanHeidinga: I think Anton's right - we want to exit with the `g_child_pid`'s status so it's the only one we need to save away.  The rest of the wait() loop looks right in that it ensures we correctly wait on all children (even children of the g_child_pid) like an init process should.

src/java.base/share/native/launcher/main.c
```diff
@@ -93,6 +93,64 @@ WinMain(HINSTANCE inst, HINSTANCE previnst, LPSTR cmdline, int cmdshow)
...
+static pid_t g_child_pid = -1;
+
+static int wait_for_children() {
+    int status = 0;
+    pid_t pid;
+    do {
+        int st = 0;
+        pid = wait(&st);
+        if (pid == g_child_pid && WIFEXITED(st)) {
+            status = WEXITSTATUS(st);
```
AntonKozlov, line 120: Sorry for nit-picking, but now if the java was killed (`WIFEXITED == false`) we won't update status and will return `0`, which does not look correct. `restorewait` in this situation returns `1`, although better, also does not look perfect. Here I suggest be at least consistent with restorewait.

Or we can fix restorewait as well, indicating being killed by returning `128+signal`, as described in the bash manual [1]. How does it sound?

> When a command terminates on a fatal signal N, bash uses the value of 128+N as the exit status.

[1] https://linux.die.net/man/1/bash

wkia: To make things iterative, I suggest to implement wait_for_children() in the same way as restorewait() for now, and then create the next PR to make appropriate changes related to signal handling and returning exit codes.

AntonKozlov: OK, makes sense for me as well.

AntonKozlov (approved):
LGTM, thanks!

PR #47, opened 2023-02-22
Fix failing CRaC tests

rvansa:
Before https://github.com/openjdk/crac/pull/16 the arguments for restored process were ignored; tests were written with this in mind and current behaviour breaks them.

In addition I've added missing `-XX:+UnlockDiagnosticVMOptions` flag and fixed the `recursiveCheckpoint` test runner which got stuck as with `pauseengine` the checkpointed process does not terminate.


AntonKozlov (approved):
LGTM, Thanks!

PR #53, opened 2023-03-23
Correct System.nanotime() value after restore

rvansa:
There are various places both inside JDK and in libraries that rely on monotonicity of `System.nanotime()`. When the process is restored on a different machine the value will likely differ as the implementation provides time since machine boot. This PR records wall clock time before checkpoint and after restore and tries to adjust the value provided by nanotime() to reasonably correct value.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/53/head:pull/53` \
`$ git checkout pull/53`

`$ git checkout pull/53` \
`$ git pull https://git.openjdk.org/crac.git pull/53/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 53`

`$ git pr show -t 53`

</details>

<a href="https://git.openjdk.org/crac/pull/53.diff">https://git.openjdk.org/crac/pull/53.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/53#issuecomment-1481432697)


src/hotspot/share/runtime/os.cpp
```diff
@@ -2023,3 +2030,23 @@ void os::PageSizes::print_on(outputStream* st) const {
...
+}
+
+void os::update_javaTimeNanos_offset() {
+  assert(checkpoint_millis >= 0, "Restore without a checkpoint?");
+  long diff_millis = javaTimeMillis() - checkpoint_millis;
+  // If the wall clock has gone backwards we won't add it to the offset
+  if (diff_millis < 0) {
+    diff_millis = 0;
+  }
+  javaTimeNanos_offset = checkpoint_nanos - javaTimeNanos() + diff_millis * 1000000L;
```
ashu-mehra, line 2051: Can you please explain why isn't `javaTimeNanos_offset = checkpoint_nanos - javaTimeNanos()` sufficient? What am I  missing here?

rvansa: @ashu-mehra Had we ignored the `diff_millis` part, the diff of nanotime before and after checkpoint would indicate that no time elapsed between checkpoint and restore. With this in place the difference does not have the expected accuracy but at least it's monotonic; wall clock time difference gives us the best possible estimate.

AntonKozlov, line 2051: Will `uptime_since_restore` still report the correct time after the change?

https://github.com/openjdk/crac/blob/9ed961106a255145274de777e151577863b013ea/src/hotspot/os/linux/os_linux.cpp#L5697

ashu-mehra:
I understand this change is trying to adjust the return value of any calls made to System.nanoTime() after restore to take into account the elapsed time between checkpoint and restore.
In principle this idea is very similar to CLOCK_BOOTTIME [0] which takes into account the time system has spent in suspend state.
I came across an issue [1] in golang which was suggested to replace CLOCK_MONOTONIC with CLOCK_BOOTTIME but it was considered ill-advised and was closed. There was even a linux kernel patch [2] to make CLOCK_MONOTONIC behave as CLOCK_BOOTTIME which was reverted [3] immediately because it broke many of the existing userland softwares.
Considering this precedent, I think we should also consider the impact of this change on existing frameworks and libraries, that is, can this change break the existing code patterns that use System.nanoTime()?

[0] https://man7.org/linux/man-pages/man3/clock_gettime.3.html
[1] https://github.com/golang/go/issues/24595
[2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d6ed449afdb38f89a7b38ec50e367559e1b8f71f
[3] https://www.spinics.net/lists/linux-tip-commits/msg43709.html

src/hotspot/share/runtime/os.cpp
rvansa: It should, since both times are read after restore.

rvansa:
@ashu-mehra The main point of this change is *not* about whether the time being suspended should be observed or not; I am rather worried about moving the process to another system and getting totally nonsense results from nanotime diffs, and broken code.

I understand that observing the suspended can be a subject to further discussion, though I incline towards the visibility of such interval, as implemented here. Since this fixes some use cases and does not change what wouldn't be broken (on a single system the paused time with system running would be observed anyway unless the whole machine was suspended) I suggest to merge this as-is without considering the topic resolved forever.

The fact that some timers use this as the time source rather than wall clock time is an implementation detail. Applications performing checkpoint and restore will require some tweaks to perform correctly and I intend to work on ways to deal with timers.

src/hotspot/share/runtime/os.cpp
AntonKozlov: restore_start_counter is indeed read on restore, in the process that execs into CREngine, so the value is not adjusted. The value then transfered to the restored VM via SHM [1] and then used for the difference between the javaTimeNanos() which is going to be adjusted. I'm worried the difference between not adjusted and adjusted value may lose sense.

Could you extend the NanoTimeTest to demonstrate the `uptime_since_restore` cannot become negative or "too" big?

[1] https://github.com/openjdk/crac/commit/9ed961106a255145274de777e151577863b013ea#diff-aeec57d804d56002f26a85359fc4ac8b48cfc249d57c656a30a63fc6bf3457adR6385

ashu-mehra:
> I am rather worried about moving the process to another system and getting totally nonsense results from nanotime diffs, and broken code.

@rvansa  I am assuming you are seeing time going backwards with System.nanoTime() calls after restore. Is that correct?
Its interesting if you are seeing such absurd results after restore, because IIUC criu is using time namespace to update the clock offsets in `/proc/<pid>/timens_offsets`, so I wouldn't expect System.nanoTime() to give absurd results on restore.  
I did some tests with C program that call `clock_gettime(CLOCK_MONOTONIC)` before and after restore. In between I restarted my system. I didn't see time going backwards; all I could observe is that the elapsed time between checkpoint and restore was not taken into account.

Can you please provide more details about your observation on nonsense results. Does the system where the absurd behavior is observed support time namepace ? Which version of criu did you use?

rvansa:
@ashu-mehra To be honest I didn't know about this being handled in CRIU - I thought that the offsets can be set only once in the newly created namespace. Does CRIU create another ns for the process?

I have not observed the problem in practice, rather anticipated it as some while ago I ran into similar problem with GraalVM inlining nanotime in a static final variable (caused 100% CPU usage on some machines and normal behaviour on others). I shall rerun the enclosed test without the fix applied but I think it was failing.

ashu-mehra:
> Does CRIU create another ns for the process?

If I am reading it correctly the criu does seem to create new time ns [0].
The complete patch can be seen here [1].

[0] https://github.com/checkpoint-restore/criu/blob/7977416ca821b4895fd5dbc498eb95f7e77f2ede/criu/timens.c#L79
[1] https://github.com/checkpoint-restore/criu/commit/4127ef4ab769dc4417c22d0ce0a4ddaaca4193b4

rvansa:
Thanks for those pointers. I've checked that the NanoTimeTest fails without the 'fix' in place; The dump creates `timens-0.img` (35 bytes) but it looks like this is not used later on; when I run restore with `-v4` I see these log messages:
```
(00.003124) Reading image tree
(00.003138) Add mnt ns 6 pid 8
(00.003139) Add net ns 2 pid 8
(00.003140) Add pid ns 1 pid 8
(00.003141) pstree pid_max=8
(00.003143) Will restore in 0 namespaces
(00.003144) NS mask to use 0
```
and I can't find log entry that would report this:
```
pr_debug("timens: monotonic %ld %ld\n", ts.tv_sec, ts.tv_nsec);
```
Any clues why this is not applied?

ashu-mehra:
what is the criu version you are on?

rvansa:
I am using this one: https://github.com/CRaC/criu/releases/tag/release-1.3
```
Version: 3.17.1-crac
GitID: 8926431
```

src/hotspot/share/runtime/os.cpp
rvansa: You were right, I've applied the correction. I also took the liberty of renaming `_restore_start_counter` (and friends) to `_restore_start_nanos` as the 'counter' meaning might be a bit misleading.

AntonKozlov:
Crac-criu does not use restore timens [1]  since once a bug in kernel or criu caused timedwait to return immediatelly everytime that is called after restore. I don't remember the bug exactly (already fixed), but I believe it was discussed on this maillist.

In general, we should not to depend on very obscure linux abillities, as this reduce chances we'd be able to run on something rather than linux. Having the code in the JVM provides better control for the implementation the java spec.

[1] https://github.com/CRaC/criu/commit/1cb2f4a518a4ae471a1df7a9b540203c1efaf1ba

src/hotspot/share/runtime/os.cpp
```diff
@@ -2023,3 +2030,23 @@ void os::PageSizes::print_on(outputStream* st) const {
...
 }
+
+void os::record_time_before_checkpoint() {
+  // If CRaC supports multiple checkpoint - restore - checkpoint - restore cycles
+  // we want to record the timestamps only on the first checkpoint, but update
+  // the offset after each restore
+  if (checkpoint_millis < 0) {
+      checkpoint_millis = javaTimeMillis();
+      checkpoint_nanos = javaTimeNanos();
+  }
```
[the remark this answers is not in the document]
rvansa: Oh wait a sec, you're partially right - since we always use javaTimeNanos() if the offset calculated after the first restore wouldn't be zero, we wouldn't have this right. I should zero the offset before calculating it again. Too bad I can't create a test for that yet.

src/hotspot/share/runtime/os.cpp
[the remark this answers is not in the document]
rvansa: If the monotonic time on the machine advanced by X, the offset can't be lower than -X (as the millis part is always positive) and therefore any difference between times read before and after will be at least 0. Therefore things snapshot might seem to take no time, but System.nanoTime() is still monotonic.

src/hotspot/share/runtime/os.cpp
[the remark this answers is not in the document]
AntonKozlov: My concern here about this code. This code assumes multiple cycles possible, and does something in not straightforward way. I think we'd better streamline this code (assuming repeated cycles, or not).

Should not we just record checkpoint millis and nanos unconditionally?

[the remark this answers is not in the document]
AntonKozlov: > It estabilishes relation between real time and monotonic time, and it's sufficient to do that just once

I don't belive this is true. Real time can be squezed and extended https://man7.org/linux/man-pages/man3/adjtime.3.html. So it won't be correct to apply a difference in real time to the monotonic clock to calculate the value of monotonic clock in another point of time.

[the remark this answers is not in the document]
rvansa: Alright then, changing wall clock time might be of concern, and I realized it should work even when this is set every time. I'll update it.

src/hotspot/share/runtime/os.cpp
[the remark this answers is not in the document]
rvansa: Looks like there's `/proc/sys/kernel/random/boot_id` so we can set the offset only when the boot ID changes. Since monotonic time is ~time since boot, too, this could work well.

rvansa:
I've updated the PR to change nanotime only if machine `boot_id` changes, and always record the millis + nanos combo before checkpoint. Now I use `ghcr.io/rvansa/crac-test-base` image as the base for tests (we need some native libraries on top of Ubuntu), hopefully we'll set up an org-based repository (`ghcr.io/crac/test-base`) and I'll change update it in the PR.

src/hotspot/share/runtime/os.cpp
```diff
@@ -2032,23 +2033,25 @@ void os::PageSizes::print_on(outputStream* st) const {
...
+  if (!read_bootid(buf, sizeof(buf)) || strncmp(buf, checkpoint_bootid, UUID_LENGTH) != 0) {
+    assert(checkpoint_millis >= 0, "Restore without a checkpoint?");
+    long diff_millis = javaTimeMillis() - checkpoint_millis;
+    // If the wall clock has gone backwards we won't add it to the offset
+    if (diff_millis < 0) {
+      diff_millis = 0;
+    }
+    // Make the javaTimeNanos() on the next line return true monotonic time
+    javaTimeNanos_offset = 0;
+    javaTimeNanos_offset = checkpoint_nanos - javaTimeNanos() + diff_millis * 1000000L;
```
AntonKozlov, line 2055: First assignment does not make effect.

src/hotspot/share/runtime/os.cpp
```diff
@@ -2032,23 +2033,25 @@ void os::PageSizes::print_on(outputStream* st) const {
...
-  // If the wall clock has gone backwards we won't add it to the offset
-  if (diff_millis < 0) {
-    diff_millis = 0;
-  }
-  // Make the javaTimeNanos() on the next line return true monotonic time
-  javaTimeNanos_offset = 0;
-  javaTimeNanos_offset = checkpoint_nanos - javaTimeNanos() + diff_millis * 1000000L;
+  char buf[UUID_LENGTH + 1];
+  // We will change the nanotime offset only if this is not the same boot
+  // to prevent reducing the accuracy of System.nanoTime() unnecessarily
```
AntonKozlov, line 2045: But it would be nice to ensure monotonicity even if it looks like the same boot. Like
```
if (!same_boot)  {
  ...
} else if ((diff = (checkpoint_nanos - javaTimeNanos()) > 0) {
  javaTimeNanos_offset = diff + 1;
}
```

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6595,3 +6600,30 @@ void os::print_memory_mappings(char* addr, size_t bytes, outputStream* st) {
     st->cr();
   }
 }
+
+bool os::read_bootid(char *dest, size_t size) {
+  int fd = ::open("/proc/sys/kernel/random/boot_id", O_RDONLY);
```
AntonKozlov, line 6605: Bood_id looks interesting! But AFAICS it remains the same for each new container, and the boot time may have been adjusted for that container.
```
anton@mercury:~$ docker run -it ubuntu:20.04 cat /proc/sys/kernel/random/boot_id
9b913973-3082-471a-add5-6b802a04a9b2
anton@mercury:~$ docker run -it ubuntu:20.04 cat /proc/sys/kernel/random/boot_id
9b913973-3082-471a-add5-6b802a04a9b2
anton@mercury:~$ cat /proc/sys/kernel/random/boot_id
9b913973-3082-471a-add5-6b802a04a9b2
```

Should not we mix something extra to boot_id, for example, a hostname (which is different for each container)?

src/hotspot/share/runtime/os.cpp
rvansa: It does; `javaTimeNanos()` uses `javaTimeNanos_offset` underhood.

src/hotspot/os/linux/os_linux.cpp
rvansa: The aim of this work is to prevent 'random' readings. New containers will keep the `boot_id` but their monotonic time *might* be adjusted, true. I agree with the comment above that in that case we should still ensure monotonicity, but since the readings are not 'random' I think that we can assume that the user did adjust monotonic time intentionally, and the application should observe that.

AntonKozlov: OK, agree. This may be a comment in the code, explaining that we leave some control of the time in the user's hands.

src/hotspot/share/runtime/os.cpp
AntonKozlov: Indeed! Could you extend the comment, highlightint that javaTimeNanos uses javaTimeNanos_offset

src/java.base/share/classes/java/lang/System.java
```diff
@@ -66,6 +66,7 @@
 import java.util.function.Supplier;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Stream;
+
```
jankratochvil, line 69: Excessive whitespace change.

src/java.base/share/classes/java/lang/System.java
```diff
@@ -2449,4 +2450,5 @@ public void exit(int statusCode) {
             }
         });
     }
+
```
jankratochvil, line 2453: Excessive whitespace change.

test/hotspot/jtreg/testlibrary/jittester/conf/exclude.methods.lst
```diff
@@ -26,7 +26,7 @@ java/lang/System::lineSeparator()
 java/lang/System::load(Ljava/lang/String;)
 java/lang/System::loadLibrary(Ljava/lang/String;)
 java/lang/System::mapLibraryName(Ljava/lang/String;)
-java/lang/System::nanoTime()
+java/lang/System::nanoTime0()
```
jankratochvil, line 29: Is this change really needed?

jankratochvil:
On Fedora 36 x86_64 when I snapshot the image, reboot and restore it with boottime earlier than it was during the snapshot I get a hanging restore:
```
#0  0x00007f43778899b9 in __futex_abstimed_wait_common () from /lib64/libc.so.6
#1  0x00007f437788e983 in __pthread_clockjoin_ex () from /lib64/libc.so.6
#2  0x00007f4377abe655 in CallJavaMainInNewThread (stack_size=1048576, args=0x7ffdd525b5d0) at ../src/java.base/unix/native/libjli/java_md.c:681
#3  0x00007f4377abb7fd in ContinueInNewThread (ifn=0x7ffdd525b6d0, ifn@entry=0x0, threadStackSize=<optimized out>, argc=<optimized out>, argv=0x5641c59c75f8, mode=mode@entry=1, what=0x5641c59c7380 "NanoTime", what@entry=0x0, ret=0) at ../src/java.base/share/native/libjli/java.c:2362
#4  0x00007f4377abe709 in JVMInit (ifn=0x0, ifn@entry=0x7ffdd525b6d0, threadStackSize=<optimized out>, argc=<optimized out>, argv=<optimized out>, mode=mode@entry=1, what=0x0, what@entry=0x5641c59c7380 "NanoTime", ret=<optimized out>) at ../src/java.base/unix/native/libjli/java_md.c:706
#5  0x00007f4377abc4e0 in JLI_Launch (argc=<optimized out>, argc@entry=3, argv=<optimized out>, argv@entry=0x5641c59c72c0, jargc=jargc@entry=0, jargv=jargv@entry=0x5641c3cfa008 <const_jargs>, appclassc=appclassc@entry=0, appclassv=appclassv@entry=0x0, fullversion=0x5641c3cf8088 "17-internal+0-adhoc.azul.crac-git", dotversion=0x5641c3cf8079 "0.0", pname=0x5641c3cf8074 "java", lname=0x5641c3cf806c "openjdk", javaargs=0 '\000', cpwildcard=1 '\001', javaw=0 '\000', ergo=0) at ../src/java.base/share/native/libjli/java.c:342
#6  0x00005641c3cf730a in main (argc=<optimized out>, argv=<optimized out>) at ../src/java.base/share/native/launcher/main.c:282
```

test/jdk/jdk/crac/java/lang/System/NanoTimeTest.java
```diff
@@ -0,0 +1,132 @@
...
+            // the difference in monotonic time is considered "random"
+            assertLT(after, before + TimeUnit.HOURS.toNanos(1), "After too late");
+        } else {
+            assertGT(after, before + TimeUnit.HOURS.toNanos(1), "After too early");
+            assertLT(after, before + TimeUnit.HOURS.toNanos(25), "After too late");
+        }
+        long boottimeAfter = readSystemUptime();
+        assertGTE(boottimeAfter, boottimeBefore + 86_400_000, "Boottime was not changed");
+        RuntimeMXBean runtimeMX = ManagementFactory.getRuntimeMXBean();
+        assertGTE(runtimeMX.getUptime(), 0L,"VM Uptime is negative!");
```
jankratochvil, line 121: whitespace

test/jdk/jdk/crac/java/lang/System/NanoTimeTest.java
```diff
@@ -0,0 +1,132 @@
...
+            builder.withBaseImage("ghcr.io/crac/test-base", "latest")
+                    .dockerOptions("-v", bootIdFile + ":/fake_boot_id")
+                    .inDockerImage(imageName);
+
+            Files.writeString(bootIdFile, "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx");
+            // We need to preload the library before checkpoint
+            builder.doCheckpoint(Container.ENGINE_COMMAND, "exec",
+                    "-e", "LD_PRELOAD=/opt/path-mapping-quiet.so",
+                    "-e", "PATH_MAPPING=/proc/sys/kernel/random/boot_id:/fake_boot_id",
+                    CracBuilder.CONTAINER_NAME, CracBuilder.DOCKER_JAVA);
```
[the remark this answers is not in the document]
rvansa: Are you using your own build of CRIU? Looks like you have CRIU build with libbsd support. When you remove `libbsd-devel` and rebuild CRIU this should work.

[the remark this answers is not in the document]
jankratochvil: Besides discussion of this patch vs. native CRIU timens support I find easier (YMMV) and less error-prone to just use simple `LD_PRELOAD`+`RTLD_NEXT` to simulate the system reboots: https://stackoverflow.com/a/6083624/2995591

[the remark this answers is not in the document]
rvansa: I agree with your line of thinking, and since this test actually uses custom docker image I can add those libraries to the base image. But in general it's a problem of criu packaging with implicit dependencies, officially installed binary should get this handled in dependency management. Since those few functions from libbsd are reimplemented in criu, we could just disable the dependency in our fork.
However we'll run into the same problem with libnftables - I found that with the nftables support the unprivileged restore as root in Docker container does not work anymore. And that got compiled in only because of the -dev library present, there's no compile time switch.

The error you describe can be solved with `git clean -f -x`, regrettably `make clean` does not do its job very well.

Not sure what function are you suggesting to replace?

rvansa:
@jankratochvil After trying to run C/R with fewer privileges in containers I am starting to see the value of moving away from CRIU; it does more than JVM usually needs, but this requires elevated privileges. Leaving it all at once would be a tremendous endeavour, though, so slowly dropping functionality, even through adding some switches into CRaC fork to disable some parts makes sense to me.

rvansa:
@jankratochvil I am going to need some extra info on that hang; I have added some tests involving negative boottime (although bootime should not really change anything) and in the end rebooted my machine to test this out as you desribed but haven't had any issues.

I also need to check what's wrong with the test in CI since I can't reproduce the failure locally.

test/jdk/jdk/crac/java/lang/System/NanoTimeTest.java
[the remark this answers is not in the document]
jankratochvil: I mean to provide normal LD_PRELOAD library and then one no longer needs to deal with any containers with their associated problems discussed above.
[libvirtual.c.txt](https://github.com/openjdk/crac/files/11465377/libvirtual.c.txt)
[clock_gettime.cpp.txt](https://github.com/openjdk/crac/files/11465400/clock_gettime.cpp.txt)
```
rm -rf dir;mkdir dir;LD_PRELOAD=$PWD/libvirtual.so CLOCK_MONOTONIC=10 BOOT_ID=$PWD/fake_boot_id ./clock_gettime &p=$!;sleep 2;sudo ./build/linux-x86_64-server-fastdebug/jdk/lib/criu dump --shell-job -t $p -D dir
CLOCK_MONOTONIC=88547. 87286504 CLOCK_BOOTTIME=88547. 87304536
CLOCK_MONOTONIC=88548. 87433771 CLOCK_BOOTTIME=88548. 87445951
CLOCK_MONOTONIC=88549. 87564393 CLOCK_BOOTTIME=88549. 87573859
Warn  (criu/pages-compress.c:216): decompress: Can't open 'pages-1.comp.img' for dfd=3, errno=2 [--- that is some bug]
CLOCK_MONOTONIC=10.    82447 CLOCK_BOOTTIME=88556.959892251
CLOCK_MONOTONIC=11.   179010 CLOCK_BOOTTIME=88557.959969582
CLOCK_MONOTONIC=12.   264924 CLOCK_BOOTTIME=88558.960056407
```
With original JVM it correctly crashes:
```
#  Internal Error (../../src/hotspot/share/runtime/thread.cpp:2470), pid=74780, tid=74781
#  assert(false) failed: unexpected time moving backwards detected in JavaThread::sleep()
```
With your patch it runs although somehow slowly.
It may need some adjustments, I wrote the library as a proof of concept to prevent the containers.
One cannot use the library to test native CRIU timens as that happens at kernel-level.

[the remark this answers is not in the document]
rvansa: How would you compile that library as part of of the jtreg? Just do `Runtime.getRuntime().exec("gcc", ...)` during the test? Can you create some native libraries as a part of the test library build?

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6391,3 +6396,30 @@ void os::print_memory_mappings(char* addr, size_t bytes, outputStream* st) {
...
+    } else if (r < 0) {
+      if (errno == EINTR) {
+        continue;
+      }
+      success = false;
+      perror("CRaC: Cannot read system boot ID");
+      break;
+    }
+    rd += r;
+  } while (rd < size - 1);
```
jankratochvil, line 6421: I would expect it would be good to check the file really has the expected length and the UUID ends with `'\n'`.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6391,3 +6396,30 @@ void os::print_memory_mappings(char* addr, size_t bytes, outputStream* st) {
...
+  dest[size - 1] = '\0';
```
jankratochvil, line 6422: When the function supports only `UUID_LENGTH` files it does not make much sense to zero-terminate it. It is easier to rather do `memcmp` of `UUID_LENGTH' bytes.

rvansa:
Thanks for the comments on boot ID @jankratochvil , incorporated.

test/jdk/jdk/crac/java/lang/System/NanoTimeTest.java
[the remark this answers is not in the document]
jankratochvil: I do not know much jtreg. One can try `Runtime.getRuntime().exec("gcc", ...)` and Java experts can improve it later.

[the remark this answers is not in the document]
rvansa: I've checked that test sources actually contain various *.c files, but it doesn't seem like this would be compiled from the test. It would make sense to do that from the Makefiles, but haven't found explicit mentions...
Anyway I find packaging stuff in containers quite convenient (and we don't have to add any extra C sources). I can add the released build of CRIU into the test-base image and remove the part mounting it from local system, would that work for you?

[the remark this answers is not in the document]
jankratochvil: I am not sure what it then tests when everything is bundled. But I am fine if the test does PASS.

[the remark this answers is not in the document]
AntonKozlov: JTReg tests has so-called native-image, a part of the test-image, consisting of the native code required for the tests to pass. It's created in during jdk build. There are a few .c files in the test/hotspot/jtreg/, probably they would help...

rvansa:
@jankratochvil The problem with 'hanging' restore is actually not specific to Fedora at all; from the description I thought that it is unresponsive in native part, but `jstack` tells that it was your testcase sleeping in `Thread.sleep()`. Turns out `os::PlatformEvent::park` creates an absolute timestamp on monotonic time and calls `pthread_cond_timedwait` - when the time runs backward during restore this does not get reevaluated. Based on the codepath I think that this would not be limited to `Thread.sleep()` but many other 'waiting' synchronization primitives. 

While I consider this a serious issue (and I'll be thinking about a solution), I think that it's independent from this PR - this handles the value we're reading directly through `System.nanoTime()`. I'll also test whether it happens when timens is used by CRIU.

rvansa:
Hi @AntonKozlov was this only off your radar or is there anything that still needs consideration?

AntonKozlov (approved):
Sorry, LGTM!

PR #54, opened 2023-03-23
Resize ForkJoinPool and some concurrent data structures

rvansa:
Some parts of JDK expect that #cpus is constant; update those places after restore.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/54/head:pull/54` \
`$ git checkout pull/54`

`$ git checkout pull/54` \
`$ git pull https://git.openjdk.org/crac.git pull/54/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 54`

`$ git pr show -t 54`

</details>

<a href="https://git.openjdk.org/crac/pull/54.diff">https://git.openjdk.org/crac/pull/54.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/54#issuecomment-1481491986)

### DIFF
--- a/src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java
+++ b/src/java.base/share/classes/java/util/concurrent/ConcurrentHashMap.java
@@ -593,9 +593,6 @@ public class ConcurrentHashMap<K,V> extends AbstractMap<K,V>
     static final int RESERVED  = -3; // hash for transient reservations
     static final int HASH_BITS = 0x7fffffff; // usable bits of normal node hash
 
-    /** Number of CPUS, to place bounds on some sizings */
-    static final int NCPU = Runtime.getRuntime().availableProcessors();
-
     /**
      * Serialized pseudo-fields, provided only for jdk7 compatibility.
      * @serialField segments Segment[]
@@ -2423,7 +2420,8 @@ public class ConcurrentHashMap<K,V> extends AbstractMap<K,V>
      */
     private final void transfer(Node<K,V>[] tab, Node<K,V>[] nextTab) {
         int n = tab.length, stride;
-        if ((stride = (NCPU > 1) ? (n >>> 3) / NCPU : n) < MIN_TRANSFER_STRIDE)
+        int ncpu = CpuCount.get();
+        if ((stride = (ncpu > 1) ? (n >>> 3) / ncpu : n) < MIN_TRANSFER_STRIDE)
             stride = MIN_TRANSFER_STRIDE; // subdivide range
         if (nextTab == null) {            // initiating
             try {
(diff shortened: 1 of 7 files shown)


rvansa:
@DanHeidinga No, this is my original work. I went through the changes in more recent JDK and those changes were far from trivial, therefore a backport would not be straightforward. If CRaC is rebased on a more recent JDK we should scratch these changes and use the `setParallelism` instead; this PR intends to provide a minimalistic solution in JDK 17.

DanHeidinga:
I agree that the changes Doug made are "far from trivial" which is due to the complexity of the problem space.  Changing the # of cpus at runtime is those algorithms is a tricky problem which makes me hesitant to review a minimal solution.

Have you considered backporting the entire set of j.u.c.* classes that Doug updated in the mainline to jdk17?  That would probably provide a more consistent solution and any bugs would be shared with the mainline

rvansa:
@ashu-mehra that was my intention; I expect these changes to be overwritten when main crac rebases to later JDK, and keep them only in crac-17u.

@DanHeidinga I would expect this kind of reaction for 17 as well, but since I am not so positive in maintaining them in CRaC I went rather for this independent PR. There won't be a perfect solution; let's see what other stakeholders would have to say.

AntonKozlov:
From a purely technical point of view, backporting those changes from upstream provides the CRaC implementation better understood, tested, etc.

Imaging we have a space like crac-17u (jdk17u with CRaC changes), and a policy allowing backporting something extra compared to jdk17u, would we need a special CRaC version to resize FJP or backports would be enough?

AntonKozlov:
From a purely technical point of view, backporting those changes from upstream provides the CRaC implementation better understood, tested, etc.

Imaging we have a space like crac-17u (jdk17u with CRaC changes), and a policy allowing backporting something extra compared to jdk17u, would we need a special CRaC version to resize FJP or backports would be enough?

DanHeidinga:
> From a purely technical point of view, backporting those changes from upstream provides the CRaC implementation better understood, tested, etc.
> 

+1

> Imaging we have a space like crac-17u (jdk17u with CRaC changes), and a policy allowing backporting something extra compared to jdk17u, would we need a special CRaC version to resize FJP or backports would be enough?

I think the backports would cover the tricky changes to FJP though we'd still need a JDKResrouce::{beforeCheckpoint, afterCheckpoint} implementation.

An ::afterCheckpoint method to call ::setParallelism to match the number of cpu's on the restored system.  The ::beforeCheckpoint implementation could be used to resize the pool to an expected number of cpus (default 4?) as an optional optimization.

DanHeidinga:
> From a purely technical point of view, backporting those changes from upstream provides the CRaC implementation better understood, tested, etc.
> 

+1

> Imaging we have a space like crac-17u (jdk17u with CRaC changes), and a policy allowing backporting something extra compared to jdk17u, would we need a special CRaC version to resize FJP or backports would be enough?

I think the backports would cover the tricky changes to FJP though we'd still need a JDKResrouce::{beforeCheckpoint, afterCheckpoint} implementation.

An ::afterCheckpoint method to call ::setParallelism to match the number of cpu's on the restored system.  The ::beforeCheckpoint implementation could be used to resize the pool to an expected number of cpus (default 4?) as an optional optimization.

rvansa:
Given the plans to rebase on a more recent JDK I think this can be closed. When the rebase happens I will check again on the NCPU changes.

PR #57, opened 2023-04-06
Support repeated checkpoint and restore operations

rvansa:
* VM option CRaCCheckpointTo is recognized when restoring the application (destination can be changed)
* The main problem for checkpoint after restore was old checkpoint image mmapped to files (CRaC-specific CRIU optimization for faster boot). Before performing checkpoint we transparently swap this with memory using anonymous mapping.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/57/head:pull/57` \
`$ git checkout pull/57`

`$ git checkout pull/57` \
`$ git pull https://git.openjdk.org/crac.git pull/57/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 57`

`$ git pr show -t 57`

</details>

<a href="https://git.openjdk.org/crac/pull/57.diff">https://git.openjdk.org/crac/pull/57.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/57#issuecomment-1498962564)

### DIFF
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -78,6 +78,7 @@
 #include "utilities/events.hpp"
 #include "utilities/elfFile.hpp"
 #include "utilities/growableArray.hpp"
+#include "utilities/hashtable.hpp"
 #include "utilities/macros.hpp"
 #include "utilities/powerOfTwo.hpp"
 #include "utilities/vmError.hpp"
@@ -288,6 +289,7 @@ class CracRestoreParameters : public CHeapObj<mtInternal> {
   char* _raw_content;
   GrowableArray<const char *>* _properties;
   const char* _args;
+  const char* _next_checkpoint_to;
 
   struct header {
     jlong _restore_time;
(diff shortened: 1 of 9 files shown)


src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6198,6 +6380,8 @@ void VM_Crac::doit() {
   // dry-run fails checkpoint
   bool ok = !_dry_run;
 
+  remap_old_imagedir();
```
AntonKozlov, line 6383: VM was not bothered the way CREngine saved the memory content. The mmaping is an implementation detail of the CR mechnism.

Have you considered switching off the mmaping in CRIU in this repeated checkpoint-restore sequence? Assuming we would be able communicate that to CREngine (in CRIU mmaping is an option).

Semantically, this patch propopses to handle a mapping twice, once in CRIU with mmaping and another time in the VM. There are some benefits of doing everything in the VM and having better control over the process. So it would be cleaner to do a practically big part of the memory management in the VM and leaving bootstraping only to the CRIU.

rvansa: I think that mmaping in CRIU is an important optimization that speeds up boot, so I did not want to force disabling that if you ever want to do the checkpoint again.

You're right that this mixes the abstractions and responsibilities. There's an alternative solution that would not require changes in the VM, but it's technically more complex: we could ptrace VM and replace the mapping for it externally (though I am not sure how exactly should we invoke syscall on behalf of the tracee - maybe a parasite code would be needed?). Since ptracing process needs elevated priviledges we should probably add this as a separate criu command that would be invoked by criuengine.
The advantage is clearer semantics and not relying on SIGSEGV handling here which some consider a sketchy practice. But it's a more complex solution.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6520,6 +6704,11 @@ void os::Linux::close_extra_descriptors() {
   closedir(dir);
 }
 
+// Since putenv does not do its own copy of the strings we need to keep
```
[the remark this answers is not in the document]
rvansa: See the comment in the old code:

> // left this pointer unowned, it is freed when process dies

Without repeated CRs it was fine to leak some memory, but now we have to keep it around until the vars are overwritten and release afterwards. Regrettably `putenv` does not make an internal copy.

PR #58, opened 2023-04-12
RCU Lock - RW lock with very lightweight read- and heavyweight write-locking

rvansa:
This implementation is suitable for uses where the write-locking happens very rarely (if at all), as in the case of CRaC checkpoint, and we don't want to slow down regular access to the protected resource.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/58/head:pull/58` \
`$ git checkout pull/58`

`$ git checkout pull/58` \
`$ git pull https://git.openjdk.org/crac.git pull/58/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 58`

`$ git pr show -t 58`

</details>

<a href="https://git.openjdk.org/crac/pull/58.diff">https://git.openjdk.org/crac/pull/58.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/58#issuecomment-1504938895)

### DIFF
--- a/src/java.base/share/classes/jdk/crac/RCULock.java
+++ b/src/java.base/share/classes/jdk/crac/RCULock.java
@@ -1,0 +1,255 @@
+package jdk.crac;
+
+import jdk.internal.ref.CleanerFactory;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.lang.invoke.*;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/**
+ * This class represents a RW-lock with extremely lightweight read-locking
+ * and heavyweight write locking.
+ * <p>
+ * Readers should use the {@link #readLock()} and {@link #readUnlock()}; when
+ * the writer lock is not locked these are similar to calling no-op methods.
+ * The lock needs a declaration of all read-critical (top-level) methods;
+ * contrary to the regular synchronization pattern we need to
+ * <strong>acquire</strong> the read-lock <strong>inside</strong> the critical
+ * section (at the beginning) and <strong>release</strong> it <strong>outside</strong>,
+ * preferrably in the <code>finally</code> block.
+ * Therefore, the code should be normally separated into wrapper and implementation method:
+ * <hr><blockquote><pre>
+ *   public void wrapper() {
+ *       try {
+ *           implementation();
+ *       } finally {
+ *           lock.readUnlock();
+ *       }
+ *   }
+ *
+ *   &#64;RCULock.Critical
+ *   public void implementation() {
+ *       lock.readLock();
+ *       // critical code ...
+ *   }
+ * </pre></blockquote><hr>
+ * In the example above the {@link RCULock.Critical} annotation would be used
+ * in the {@link #forClasses(Class[])} factory method; alternatively you can
+ * declare the methods (or their signatures) using one of the constructors.
+ * <p>
+ * There should be at most one writer invoking the {@link #synchronizeBegin()}
+ * and {@link #synchronizeEnd()} pair. This operation requires all Java threads
+ * to block in safepoint to inspect their stack; if the stack contains one of
+ * the read-critical methods the thread acquires read-ownership even if
+ * the actual <code>readLock()</code> invocation was a no-op.
+ */
+public class RCULock {
+    private final ReentrantLock lock = new ReentrantLock(true);
+    private final Condition beginCondition = lock.newCondition();
+    private final Condition endCondition = lock.newCondition();
+    private boolean synchronize;
+    @SuppressWarnings("unused") // used in native code
+    private long readerThreadsList;
+    @SuppressWarnings("unused") // used in native code
+    private long readCriticalMethods;
+    private SwitchPoint lockSwitchPoint = new SwitchPoint();
+    private SwitchPoint unlockSwitchPoint = new SwitchPoint();
+    private volatile MethodHandle lockImpl;
+    private volatile MethodHandle unlockImpl;
+
+    static {
+        initFieldOffsets();
+    }
+
+    private static native void initFieldOffsets();
+
+    /**
+     * Inspect the classes for all methods (including private ones) marked with
+     * the {@link RCULock.Critical} annotation and create a lock that can protect
+     * these methods with read lock.
+     *
+     * @param classes List of classes to be inspected.
+     * @return New lock instance.
+     */
+    public static RCULock forClasses(Class<?>... classes) {
+        ArrayList<Method> methods = new ArrayList<>();
+        for (Class<?> cls : classes) {
+            while (cls != null && cls != Object.class) {
+                for (Method m : cls.getDeclaredMethods()) {
+                    if (m.isAnnotationPresent(Critical.class)) {
+                        methods.add(m);
+                    }
+                }
+                cls = cls.getSuperclass();
+            }
+        }
+        return new RCULock(methods);
+    }
+
+    /**
+     * Create a lock that can protect selected methods with read-locking.
+     *
+     * @param readCriticalMethods List of methods with the read-critical section.
+     */
+    public RCULock(Iterable<Method> readCriticalMethods) {
+        this(StreamSupport.stream(readCriticalMethods.spliterator(), false)
+                .map(m -> m.getDeclaringClass().getName() + "." + m.getName() + "(" +
+                Arrays.stream(m.getParameterTypes())
+                        .map(Class::descriptorString).collect(Collectors.joining())
+                + ")" + m.getReturnType().descriptorString()).toArray(String[]::new));
+    }
+
+    /**
+     * Create a lock that can protect selected methods with read-locking.
+     * <p>
+     * The signatures follow the form
+     * <pre>my.package.MyClass.foobar(another/package/SomeClass;Z)V</pre>
+     *
+     * @param readCriticalMethods List of signatures for methods invoked in the read-critical section.
+     */
+    public RCULock(String[] readCriticalMethods) {
+        initSwitchPoints();
+        //noinspection CapturingCleaner
+        CleanerFactory.cleaner().register(this, this::destroy);
+        Arrays.sort(readCriticalMethods);
+        init(readCriticalMethods);
+    }
+
+    private native void init(String[] readCriticalMethods);
+
+    private native void destroy();
+
+    private void initSwitchPoints() {
+        try {
+            MethodType voidType = MethodType.methodType(void.class);
+            MethodHandle noop = MethodHandles.lookup().findSpecial(RCULock.class, "noop", voidType, RCULock.class);
+            MethodHandle readLockImpl = MethodHandles.lookup().findSpecial(RCULock.class, "readLockImpl", voidType, RCULock.class);
+            MethodHandle readUnlockImpl = MethodHandles.lookup().findSpecial(RCULock.class, "readUnlockImpl", voidType, RCULock.class);
+            lockSwitchPoint = new SwitchPoint();
+            lockImpl = lockSwitchPoint.guardWithTest(noop, readLockImpl);
+            unlockSwitchPoint = new SwitchPoint();
+            unlockImpl = lockSwitchPoint.guardWithTest(noop, readUnlockImpl);
+        } catch (NoSuchMethodException | IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private void noop() {
+    }
+
+    /**
+     * Acquire the read-lock. This method <strong>must</strong> be called from
+     * within the read-critical method declared when this lock is created.
+     *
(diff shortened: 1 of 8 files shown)


rvansa:
After implementing this I realized that it could be possible to implement the check using `Thread.getAllStackTraces()` and process the stacks in Java; that's even heavier, though.

I will yet explore extending implementation with more dynamic list of methods; such implementation could let us use only one instance of RCULock in the VM.

src/java.base/share/classes/jdk/crac/RCULock.java
```diff
@@ -0,0 +1,255 @@
...
+    private native void init(String[] readCriticalMethods);
+
+    private native void destroy();
+
+    private void initSwitchPoints() {
+        try {
+            MethodType voidType = MethodType.methodType(void.class);
+            MethodHandle noop = MethodHandles.lookup().findSpecial(RCULock.class, "noop", voidType, RCULock.class);
+            MethodHandle readLockImpl = MethodHandles.lookup().findSpecial(RCULock.class, "readLockImpl", voidType, RCULock.class);
+            MethodHandle readUnlockImpl = MethodHandles.lookup().findSpecial(RCULock.class, "readUnlockImpl", voidType, RCULock.class);
```
DanHeidinga, line 136: Creating a Lookup object is expensive.  Better to cache in a local
```suggestion
            MethodHandles.Lookup lookup = MethodHandles.lookup();
            MethodHandle noop = lookup.findSpecial(RCULock.class, "noop", voidType, RCULock.class);
            MethodHandle readLockImpl = lookup.findSpecial(RCULock.class, "readLockImpl", voidType, RCULock.class);
            MethodHandle readUnlockImpl = lookup.findSpecial(RCULock.class, "readUnlockImpl", voidType, RCULock.class);
```

src/java.base/share/native/libjava/RCULock.c
```diff
@@ -0,0 +1,148 @@
...
+#include "jdk_crac_RCULock.h"
+
+static jfieldID readerThreadsListField = NULL;
+static jfieldID readCriticalMethodsField = NULL;
+
+JNIEXPORT void JNICALL
+Java_jdk_crac_RCULock_initFieldOffsets(JNIEnv *env, jclass cls)
+{
+    readerThreadsListField = (*env)->GetFieldID(env, cls, "readerThreadsList", "J");
+    readCriticalMethodsField = (*env)->GetFieldID(env, cls, "readCriticalMethods", "J");
```
DanHeidinga, line 43: Looking up fields correctly in JNI is a pain to get the error handling correct.  I prefer to do the lookup reflectively and then pass the j.l.reflect.Field objects into the JNI method and convert them using `ToReflectField` [0]

It moves all the tricky error handling into Java and makes for obviously correct code.

If we keep the existing code, we need to null check the `readerThreadsListField` value before attempting a second JNI call as we can't call in with an exception pending.

[0] https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#from_reflected_field

rvansa:
@DanHeidinga Thanks for those hints on error handling, I've corrected them in the last commit.

I've also made the list of methods mutable, implemented binary search for lookup and created a central lock that other components could easily use. This is not used anywhere, yet; the exact way to consume this in user code is up for discussion.

rvansa:
Actually, the *big lock* as I've done it doesn't make much sense; usually we'd need to do some cleanup in the resource as well. So it might make more sense to have a full Context guarded by this lock.

src/java.base/share/classes/jdk/crac/RCULock.java
```diff
@@ -0,0 +1,324 @@
...
+ * This class represents a RW-lock with extremely lightweight read-locking
+ * and heavyweight write locking.
+ * <p>
+ * Readers should use the {@link #readLock()} and {@link #readUnlock()}; when
+ * the writer lock is not locked these are similar to calling no-op methods.
+ * The lock needs a declaration of all read-critical (top-level) methods;
+ * contrary to the regular synchronization pattern we need to
+ * <strong>acquire</strong> the read-lock <strong>inside</strong> the critical
+ * section (at the beginning) and <strong>release</strong> it <strong>outside</strong>,
+ * preferrably in the <code>finally</code> block.
```
AntonKozlov, line 32: This assimetry in the interface does not look nice. Likely because the implementation, we can release the lock only outside a method with the critical code. But this leaks too much implementation details to the interface.

src/java.base/share/classes/jdk/crac/RCULock.java
```diff
@@ -0,0 +1,324 @@
...
+     * <p>
+     * The signatures follow the form
+     * <pre>my.package.MyClass.foobar(another/package/SomeClass;Z)V</pre>
+     *
+     * @apiNote This constructor does not use var-args argument to prevent
+     * accidentally creating a RCULock without declaring any critical methods.
+     *
+     * @param readCriticalMethods List of signatures for methods invoked in the read-critical section.
+     */
+    public RCULock(String[] readCriticalMethods) {
```
AntonKozlov, line 141: It's a big question, why (from the interface point of view) the lock should know methods in which it can be used.

The second tier question, why the method names are String[] and not Method[] at least.

rvansa: There's no point in asking 'why'. Interface is there to specify a contract. You declare the methods, and then the implementation fulfills the contract. And you already know why from the point of implementation.

There is another constructor that accepts `Method[]`, being able to specify the signature is just an option if it's more convenient.

rvansa:
It follows the same principle but its use is not interchangeable (and cannot be made so) - had you replaced existing place that uses `ReadWriteLock` with this one it wouldn't work.

I already made a benchmark this includes a noop baseline (`unsync`), and executes the `quick` method as fast as it can in 8 threads, and `slow` method with 10/100 ms think time (single thread):
```
Benchmark                      (impl)  (pause)   Mode  Cnt           Score           Error  Units
SwitchPointBenchmark.g:quick   unsync       10  thrpt    5  2420777624.146 ± 249641306.573  ops/s
SwitchPointBenchmark.g:slow    unsync       10  thrpt    5          99.248 ±         0.264  ops/s
SwitchPointBenchmark.g:quick   unsync      100  thrpt    5  2244724220.494 ± 328435039.061  ops/s
SwitchPointBenchmark.g:slow    unsync      100  thrpt    5           9.992 ±         0.002  ops/s
SwitchPointBenchmark.g:quick   rwlock       10  thrpt    5     4414608.947 ±   1525681.326  ops/s
SwitchPointBenchmark.g:slow    rwlock       10  thrpt    5          99.191 ±         0.160  ops/s
SwitchPointBenchmark.g:quick   rwlock      100  thrpt    5     4541641.249 ±   3166622.432  ops/s
SwitchPointBenchmark.g:slow    rwlock      100  thrpt    5           9.989 ±         0.003  ops/s
SwitchPointBenchmark.g:quick  rculock       10  thrpt    5   196537498.940 ± 305743615.522  ops/s
SwitchPointBenchmark.g:slow   rculock       10  thrpt    5          94.168 ±         2.159  ops/s
SwitchPointBenchmark.g:quick  rculock      100  thrpt    5   772304327.917 ±  28329265.290  ops/s
SwitchPointBenchmark.g:slow   rculock      100  thrpt    5           9.909 ±         0.025  ops/s
```

In case of 10 ms think time (which is really extremely often) results show more than 20x speedup compared to ReentrantReadWriteLock.readLock().lock()+unlock() combo, and just 10x slowdown vs. noop. With 100 ms think time it's order of magnitude better, > 150x speedup vs. < 3x slowdown.

I've also run benchmark with no pause time to see the maximum frequency of synchronization, and it shows about 4.5k syncs/s (it would be less with more threads and longer stacks for sure).
```
Benchmark                      (impl)  (pause)   Mode  Cnt        Score       Error  Units
SwitchPointBenchmark.g:quick  rculock        0  thrpt    5  1417151.441 ± 72183.322  ops/s
SwitchPointBenchmark.g:slow   rculock        0  thrpt    5     4486.629 ±   201.970  ops/s
```

Note that these results use single fork VM and just few short iterations, but it gives some idea about the order of magnitude.

src/java.base/share/classes/jdk/crac/RCULock.java
rvansa: It would *almost* work except for interpreted mode; in compiled code the safepoint poll is only at method exit (and in uncounted loops). I couldn't find the specific place in code but I've read that in interpreted mode the safepoint may happen on every bytecode instruction, so we would have to somehow disable this. Sounds too fragile to expect that you could jump straight into the critical section.
Alternatively we could try to do flow analysis on the bytecode and check where's the critical section, but that's also too much work just to make it look a tad better. This type of lock is intended for very specific use anyway.

src/java.base/share/classes/jdk/crac/RCULock.java
```diff
@@ -0,0 +1,325 @@
...
+        try {
+            MethodType voidType = MethodType.methodType(void.class);
+            MethodHandles.Lookup lookup = MethodHandles.lookup();
+            MethodHandle noop = lookup.findSpecial(RCULock.class, "noop", voidType, RCULock.class);
+            MethodHandle readLockImpl = lookup.findSpecial(RCULock.class, "readLockImpl", voidType, RCULock.class);
+            MethodHandle readUnlockImpl = lookup.findSpecial(RCULock.class, "readUnlockImpl", voidType, RCULock.class);
+            lockSwitchPoint = new SwitchPoint();
+            lockImpl = lockSwitchPoint.guardWithTest(noop, readLockImpl);
+            unlockSwitchPoint = new SwitchPoint();
+            unlockImpl = lockSwitchPoint.guardWithTest(noop, readUnlockImpl);
```
DanHeidinga, line 209: After spending a lot of time digging through the Hotspot C2 code, I don't think this use of SwitchPoints will be optimized in the way we might want (ie: deopt on SwitchPoint invalidation).

SwitchPoint is implemented as a special MutableCallSite that changes from a known target MH that returns true to a known target MH that returns false.  This is all wrapped into a standard guardWithTest MH (basically a MH "if" statement).

As far as I can tell, to benefit from C2 optimizing the SW, we need to have the MutableCallSite or the GuardWithTest MH rooted in a static final field or in an invokedynamic bytecode. See the code in type.cpp::make_constant_from_field which asserts a Dependency between the CallSite and the target MH for the current method.  Otherwise, we don't have a Dependency to rely on and trigger the eventual deopt.

The above analysis may be wrong so any corrections by C2 experts would be appreciated.

rvansa: Thanks for the insight! So I guess this could be really simplified to a bunch of `if`s - and it would be more clear to the reader, as he wouldn't expect some mysterious magic reason to pick SP. I'll rerun my benchmarks after that and we'll see if there's any difference.

While I an imagine limiting this to a `static` field (coarse-graining this to singleton for the whole VM - it's not too bad anyway since we execute a VM-wide operation), but I can't make this `final` as we need to flip this many times.

rvansa:
Okay, using just a volatile field does not change much. Anton suggested to have a look on the totally uncontended readers case; a single-threaded benchmark results (including the 'new' implementation that fakes an empty read-lock) look like this:
```
Benchmark              (blackhole)   (impl)   Mode  Cnt           Score           Error  Units
InlinedCall.newImpl          false      N/A  thrpt    3  2441232707.809 ± 138092251.983  ops/s
InlinedCall.rcuLocked        false      N/A  thrpt    3  1116925952.834 ± 105484443.430  ops/s
InlinedCall.rwLocked         false      N/A  thrpt    3    59334557.465 ±    267412.636  ops/s
InlinedCall.unsync           false      N/A  thrpt    3  2424408670.896 ± 696453915.897  ops/s
VirtualCall.component        false   unsync  thrpt    3  2439576011.016 ± 194572187.841  ops/s
VirtualCall.component        false   rwlock  thrpt    3    59518346.446 ±     91540.972  ops/s
VirtualCall.component        false  rculock  thrpt    3  1059312172.029 ±  91850879.226  ops/s
VirtualCall.component        false      new  thrpt    3  1845830506.910 ± 256628384.459  ops/s
```
When I run this with 6 threads I get this:
```
Benchmark              (blackhole)   (impl)   Mode  Cnt            Score            Error  Units
InlinedCall.newImpl          false      N/A  thrpt    3  12031688167.608 ± 4218322551.532  ops/s
InlinedCall.rcuLocked        false      N/A  thrpt    3   5660447942.467 ±  237218291.900  ops/s
InlinedCall.rwLocked         false      N/A  thrpt    3    341089781.812 ±   65370977.211  ops/s
InlinedCall.unsync           false      N/A  thrpt    3  11648096223.269 ±  282528592.912  ops/s
VirtualCall.component        false   unsync  thrpt    3  11406362088.019 ± 1207457766.872  ops/s
VirtualCall.component        false   rwlock  thrpt    3    331446602.891 ±   10006346.435  ops/s
VirtualCall.component        false  rculock  thrpt    3   5213272454.145 ±  707041721.132  ops/s
VirtualCall.component        false      new  thrpt    3   8796928884.245 ± 2177441524.351  ops/s
```
I've tried to see why the virtual invocation vs. inlined of `unsync` does not change while `new` has a significant difference, but I can't really tell after looking into `perfasm` results. I've also checked with disabled inlining of the entry method, and the result is quite different. It's hard to tell which version would be the 'right' one - I think that the advantage of RCU vs. RW lock is clear, and having a 2 fold slowdown vs. empty implementation isn't that bad. `perfasm` points most of the weight to those volatile reads.

PR #61, opened 2023-04-28
Support updating MANAGEABLE JVM options during restore

rvansa:
When a JVM option is MANAGEABLE it can be set at any time during runtime, therefore it is safe to change it during the restore operation. Rather than silently ignoring JVM options passed along with -XX:CRaCRestoreFrom we send them to the restored process and either update or print a warning if given option cannot be changed.

---------
### Progress

### Reviewers
 * [Anton Kozlov](https://openjdk.org/census#akozlov) (@AntonKozlov - **Reviewer**) ⚠️ Review applies to [920b3be8](https://git.openjdk.org/crac/pull/61/files/920b3be8de05aba6510f4274d72e878eb3923b10)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/61/head:pull/61` \
`$ git checkout pull/61`

`$ git checkout pull/61` \
`$ git pull https://git.openjdk.org/crac.git pull/61/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 61`

`$ git pr show -t 61`

</details>

<a href="https://git.openjdk.org/crac/pull/61.diff">https://git.openjdk.org/crac/pull/61.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/61#issuecomment-1527130857)


src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6543,6 +6554,33 @@ bool CracRestoreParameters::read_from(int fd) {
...
+        cursor = eq + 1;
+      } else {
+        *eq = '\0';
+        char* value = eq + 1;
+        result = WriteableFlags::set_flag(cursor, value, JVMFlagOrigin::CRAC_RESTORE, err_msg);
+        cursor = value + strlen(value) + 1;
+      }
+    }
+    if (result != JVMFlag::Error::SUCCESS) {
+      warning("VM Option '%s' cannot be changed, ignoring: %s",
```
AntonKozlov, line 6579: A significant set of options cannot be set on restore at the moment, so it will be even better to highlight they don't have effect and produce an error. It may be useful to revert back to warning (with e.g. an option), but by default it should be disabled (leading to the error)

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2100,10 +2103,11 @@ const intx ObjectAlignmentInBytes = 8;
       "-XX:CRaCRestoreFrom and continue initialization if restore is "      \
       "unavailable")                                                        \
                                                                             \
-  product(ccstr, CRaCIgnoredFileDescriptors, NULL, "Comma-separated list "  \
-      "of file descriptor numbers or paths. All file descriptors greater "  \
-      "than 2 (stdin, stdout and stderr are excluded automatically) not "   \
-      "in this list are closed when the VM is started.")                    \
+  product(ccstr, CRaCIgnoredFileDescriptors, NULL, MANAGEABLE,              \
```
AntonKozlov, line 2106: Having this is considered only on init, this also does not fit MANAGEABLE concept (there is no need to change that over management)

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2088,10 +2088,13 @@ const intx ObjectAlignmentInBytes = 8;
...
                                                                             \
-  product(ccstr, CRaCCheckpointTo, NULL, "Path to checkpoint image")        \
+  product(ccstr, CRaCCheckpointTo, NULL, MANAGEABLE,                        \
+        "Path to checkpoint image directory")                               \
                                                                             \
-  product(ccstr, CRaCRestoreFrom, NULL, "Path to image for restore, "       \
-      "replaces the initializing VM on success")                            \
+  /* It does not make sense to change this flag in runtime but we'll tag */ \
+  /* it MANAGEABLE to prevent warnings when setting this on restore. */     \
+  product(ccstr, CRaCRestoreFrom, NULL, MANAGEABLE,                         \
```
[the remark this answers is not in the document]
rvansa: Below we see that `CRaCIgnoredFileDescriptors` is RESTORABLE (I would rather use SET_ON_RESTORE as all flags are restored at its previous value) but not MANAGEABLE, so that's not perfect either. The reason why I rather stayed on MANAGEABLE was to prevent changing every `MANAGEABLE` to `MANAGEABLE | SET_ON_RESTORE`, that would complicate any backport from mainline.

I can do the SET_ON_RESTORE (as superset of MANAGEABLE) but I don't think that the few exceptions that could be rather documented by a few lines of comment really deserve a separate flag.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6543,6 +6554,33 @@ bool CracRestoreParameters::read_from(int fd) {
   ::_restore_start_time = hdr->_restore_time;
   ::_restore_start_counter = hdr->_restore_counter;
 
+  for (int i = 0; i < hdr->_nflags; i++) {
```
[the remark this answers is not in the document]
rvansa: Makes sense, I'll turn those into errors. I should probably also check the presence of `-jar` and `-cp`/`--classpath` and produce a nice explanation; otherwise the code would interpret those as the new main class and its arguments.

src/hotspot/share/runtime/globals.hpp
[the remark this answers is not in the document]
AntonKozlov: > The reason why I rather stayed on MANAGEABLE was to prevent changing every `MANAGEABLE` to `MANAGEABLE | SET_ON_RESTORE`, that would complicate any backport from mainline.

I tottaly support MANAGEABLE => RESTORABLE, as it makes sense.

> I can do the SET_ON_RESTORE (as superset of MANAGEABLE) but I don't think that the few exceptions that could be rather documented by a few lines of comment really deserve a separate flag.

Option are also some form of the documentation for the flags. So the new class of options deserve their own name.

> Below we see that `CRaCIgnoredFileDescriptors` is RESTORABLE but not MANAGEABLE, so that's not perfect either. 

Yes, that is another one. And there are many more, like PrintCompilation. Or any other product flag which handling in the VM allows update on restore. Over time the set of RESTORABLE flags can grow as VM implementation allows. While there is a higher bar to include a flag into MANGEABLE set [1]

[1] https://github.com/openjdk/crac/blob/crac/src/hotspot/share/runtime/globals.hpp#L86

> (I would rather use SET_ON_RESTORE as all flags are restored at its previous value)

I don't like RESTORABLE as well (relates to RESTORE, but how is not clear). The new name better to fit into existing set of: DIAGNOSTIC, EXPERIMENTAL, or MANAGEABLE.

rvansa:
@AntonKozlov Created a separate bit (RESTORE_SETTABLE - adjective as the others), and using non-settable flags fails early. I've also tried to prohibit explicitly setting `-cp` and friends but in the `parse_options_for_restore` I already cannot tell where this comes from so I have not incorporated this in this PR.
There's a few extra changes in `CracBuilder` that would help with testing ^ but I find them useful in general, so I've included those.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -54,8 +54,8 @@
 // only in this file, because the macrology requires single-token type names.
 
 // The optional extra_attrs parameter may have one of the following values:
-// DIAGNOSTIC, EXPERIMENTAL, or MANAGEABLE. Currently extra_attrs can be used
-// only with product/product_pd flags.
+// DIAGNOSTIC, EXPERIMENTAL, MANAGEABLE and RESTORE_SETTABLE. Currently
```
AntonKozlov, line 57: ` and` -> `, or`

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2088,22 +2091,24 @@ const intx ObjectAlignmentInBytes = 8;
   develop(bool, TraceOptimizedUpcallStubs, false,                           \
                 "Trace optimized upcall stub generation")                   \
                                                                             \
-  product(ccstr, CRaCCheckpointTo, NULL, "Path to checkpoint image")        \
+  product(ccstr, CRaCCheckpointTo, NULL, MANAGEABLE,                        \
```
AntonKozlov, line 2094: I see reasons for CRaCCheckpointTo to be MANAGEABLE, but at the moment the flag is assumed to be set in the command line by the implementation, e.g. os::Linux::{vm_create_start,prepare_checkpoint} are called depending on the flag value, and that can happen only during VM initialization.

A set of changes are required before the option can become MANAGEABLE.

The test should also updated once the option is reverted.

src/hotspot/os/linux/os_linux.cpp
AntonKozlov: The place can be `guarantee(result == JVMFlag::Error::SUCCESS, "...")` since the possibility was checked earlier.

AntonKozlov:
Thanks, looks mostly good except few nits.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2125,7 +2126,7 @@ const intx ObjectAlignmentInBytes = 8;
   product(bool, CRDoThrowCheckpointException, true, EXPERIMENTAL,           \
       "Throw CheckpointException if uncheckpointable resource handle found")\
                                                                             \
-  product(bool, CRTrace, true, "Minimal C/R tracing")                       \
+  product(bool, CRTrace, true, MANAGEABLE, "Minimal C/R tracing")           \
```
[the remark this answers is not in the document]
rvansa: Looking for usages (actually only one) of the flag, it qualifies to be set at any time.

AntonKozlov:
Do you have an idea why jdk/crac/recursiveCheckpoint/Test.java has failed?

src/hotspot/share/runtime/globals.hpp
rvansa: Well spotted, I was thinking about changing the path but these two places need to be called in order to prepare for a checkpoint.

rvansa:
@AntonKozlov Updated. The test failed rightfully, I've stopped adding the `CREngine` flag but that breaks the restore with non-default engine. Now made the flag RESTORE_SETTABLE.

rvansa:
I think that we could write the name of the engine used to the checkpoint directory, that way the user won't need to pass the engine on restore.

AntonKozlov (approved):
Thank you! LGTM

AntonKozlov:
Oops, there is a conflict. Could you please update the PR?

rvansa:
@AntonKozlov could you type `/sponsor` once more, please?

PR #62, opened 2023-05-04
Ensure all notifications finish even if only daemon threads remain

AntonKozlov:
If as a result of beforeCheckpoint() no more non-daemon threads remain, it's possible that VM exits prematurely, before one of afterRestore() get a chance to create another non-daemon thread that will keep VM alive. Triggering checkpoint via jcmd (so checkpointRestore() method is executed on daemon attach-listener thread), increases probability to step in the problem.

The change ensures all notifications are done while there is at least one non-daemon thread. The notification methods are still called from the original thread.

---------
### Progress

### Reviewers
 * @rvansa (no known openjdk.org user name / role) ⚠️ Review applies to [a3b1242e](https://git.openjdk.org/crac/pull/62/files/a3b1242ea396681c34dc6f7292f80d8928cf3d53)

### Contributors
 * Radim Vansa `<rvansa@azul.com>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/62/head:pull/62` \
`$ git checkout pull/62`

`$ git checkout pull/62` \
`$ git pull https://git.openjdk.org/crac.git pull/62/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 62`

`$ git pr show -t 62`

</details>

<a href="https://git.openjdk.org/crac/pull/62.diff">https://git.openjdk.org/crac/pull/62.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/62#issuecomment-1534804829)


test/jdk/jdk/crac/DaemonAfterRestore.java
```diff
@@ -0,0 +1,107 @@
...
+            System.out.println("worker thread start");
+            start.countDown();
+            try {
+                finish.await();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+            System.out.println("worker thread finish");
+        });
+        workerThread.setDaemon(false);
```
rvansa, line 82: Unnecessary, thread created from main thread (non-daemon) is not a daemon either.

test/jdk/jdk/crac/DaemonAfterRestore.java
```diff
@@ -0,0 +1,107 @@
...
+        try {
+            start.await();
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+
+        Resource resource = new Resource() {
+            @Override
+            public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+                finish.countDown();
```
rvansa, line 94: It might be worth asserting that here we are running in a daemon thread.

test/jdk/jdk/crac/DaemonAfterRestore.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+            @Override
+            public void afterRestore(Context<? extends Resource> context) throws Exception {
+                Thread.sleep(3000);
```
rvansa, line 98: Could we avoid extending the testsuite run by 3 seconds? I know we're trying to assert that 'nothing happens' rather then replacing waiting for an event by a timed wait. If we can't check that the destroy VM thread had a chance to work, I suggest at least reducing this (say 100ms?).

rvansa (commented):
The code is doing something (starting a thread) before the checkpoint, and after restore (join the thread). Rather than clobbering the 'general' C/R code with a fix for one issue, there is an interface perfectly suited to host this - a Resource.
There's a catch, though - if we register it to the JDKContext this would be run *after* user Resources, and we need to run this *before*. Right now we don't have any means to order things this way.
One solution could be reversing the JDKContext/GlobalContext relationship - JDKContext would be the parent, and would use a specific priority class for GlobalContext (something before `NORMAL`). And there would be one more priority class even before that, and this `KeepaliveResource` would be registered at it.

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -240,18 +241,43 @@ private static void checkpointRestore(long jcmdStream) throws
...
+                throw new CheckpointException("Recursive checkpoint is not allowed");
+            }
+            checkpointInProgress = true;
+            CountDownLatch finishCleanup = null;
+            Thread keepAliveThread = null;
+            if (Thread.currentThread().isDaemon()) {
+                // Create a non-daemon thread while notification is performed.
+                // The notifications are done on the original thread.
+                CountDownLatch start = new CountDownLatch(1);
+                CountDownLatch finish = new CountDownLatch(1);
```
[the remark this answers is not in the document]
AntonKozlov: Since both of suggested options are reusable, having separted events is cleaner and reduce chances to accidentally await when that was intended.

test/jdk/jdk/crac/DaemonAfterRestore.java
AntonKozlov: Otherwise there should be an assert, so it's more straightforward to set the mode.

test/jdk/jdk/crac/DaemonAfterRestore.java
AntonKozlov: 100ms does not look as a good candidate (comparable to a single scheduling period). So 3 sec does not look to bad to avoid interrmittent false positive because sleep() finished before VM has terminated.

AntonKozlov (commented):
Thank you for review.

> Rather than clobbering the 'general' C/R code with a fix for one issue, there is an interface perfectly suited to host this - a Resource.

Having Resource abstraction does not mean that is necessary to use. Resources are suited for objects which may or may not exist and still need to receive notifications. Here we have no problems with doing something directly. So we don't need to rely on some implicit ordering, nor don't need to change Contextes structure.

rvansa:
If not using a Resource, at least refactor that into a standalone (stateful) component, which may be called directly. The way it's written here mixes in some implementation details into a higher-level flow.

test/jdk/jdk/crac/DaemonAfterRestore.java
rvansa: I agree with the assert

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -240,18 +269,40 @@ private static void checkpointRestore(long jcmdStream) throws
...
-                    if (FlagsHolder.TRACE_STARTUP_TIME) {
-                        System.out.println("STARTUPTIME " + System.nanoTime() + " restore-finish");
+
+                // Create a non-daemon thread while notification is performed
+                // to avoid VM exit. The notifications are done on the original
+                // thread.
+                if (Thread.currentThread().isDaemon()) {
+                    try {
+                        keepAlive = new KeepAlive();
+                    } catch (InterruptedException e) {
```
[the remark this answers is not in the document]
rvansa: Also, if you're just rethrowing runtime exception you could move the try-catch into the class.

rvansa:
@AntonKozlov I've addressed the interrupts and moved KeepAlive to separate impl class in https://github.com/rvansa/crac/tree/daemon-after-restore - could you ff-merge into your branch to avoid opening another PR?

AntonKozlov:
Thanks for the follow-up! I've added the commit to this PR.

rvansa (approved):
I am practically approving my own changes, but ok.

PR #63, opened 2023-05-10
Support passing extra options in CREngine

rvansa:
In addition to `-XX:CREngine=program` this adds support to `-XX:CREngine=program,--key,value,--anotherkey` that translates into invoking `program --key value --anotherkey`. Commas support escaping with a backslash.

This generic parameters support is utilized in `criuengine` that accepts `--verbosity` and `--log-file` options and relays them to `criu`.

---------
### Progress

### Reviewers
 * [Anton Kozlov](https://openjdk.org/census#akozlov) (@AntonKozlov - **Reviewer**) ⚠️ Review applies to [945e95ed](https://git.openjdk.org/crac/pull/63/files/945e95ed19ecb93c13094bc70d075c071160097a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/63/head:pull/63` \
`$ git checkout pull/63`

`$ git checkout pull/63` \
`$ git pull https://git.openjdk.org/crac.git pull/63/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 63`

`$ git pr show -t 63`

</details>

<a href="https://git.openjdk.org/crac/pull/63.diff">https://git.openjdk.org/crac/pull/63.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/63#issuecomment-1541629819)


src/hotspot/share/runtime/globals.hpp
```diff
@@ -2094,7 +2094,10 @@ const intx ObjectAlignmentInBytes = 8;
       "replaces the initializing VM on success")                            \
                                                                             \
   product(ccstr, CREngine, "criuengine", "Path or name of a program "       \
-      "implementing checkpoint/restore")                                    \
+      "implementing checkpoint/restore and optional extra parameters "      \
+      "as a comma-separated list of key[=value] pairs; "                    \
+      "-XX:CREngine=program,key=value,anotherkey results in calling "       \
+      "'program --key value --anotherkey'")                                 \
```
AntonKozlov, line 2100: There is CRAC_CRIU_OPTS that is used to pass additional options to CRIU. I see that remains in place, and this turns out to be alternative?

Not every CREngine may want to follow this convention of argument passing. Usually, `program,--key,value,--anotherkey` is specified to call `program --key value --anotherkey`.

This probably needs to be RESTORE_SETTABLE.

src/java.base/unix/native/criuengine/criuengine.c
```diff
@@ -102,9 +107,13 @@ static int checkpoint(pid_t jvm,
             "-t", jvmpidchar,
             "-D", imagedir,
             "--shell-job",
-            "-v4", "-o", "dump4.log", // -D without -W makes criu cd to image dir for logs
```
AntonKozlov, line 105: The dump4.log is a great debugging aid for checkpoint and restore failures, please keep it.

I think a part of the reason for this change is to provide some logging to the console. I think the reporting should be implemented separately https://github.com/openjdk/crac/pull/64#discussion_r1221827050

rvansa: By default the file is still used, see lines 113 - 116. The args just support overriding those.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -5872,30 +5874,83 @@ static int cr_util_path(char* path, int len) {
...
+    *target = '\0';
+    _crengine_args[next_arg++] = arg;
+    _crengine_args[next_arg] = NULL;
+  }
   return true;
 }
 
+static void add_crengine_arg(const char *arg) {
+  for (size_t i = 2; i < ARRAY_SIZE(_crengine_args) - 1; ++i) {
+    if (_crengine_args[i] == NULL) {
```
AntonKozlov, line 5945: The last empty position can be remembered along _crengine_args

AntonKozlov (commented):
The update made the code simpler, thank you!

LGTM with small nits.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -436,6 +436,8 @@ static int clock_tics_per_sec = 100;
 
 // CRaC
 static const char* _crengine = NULL;
+static char* _crengine_arg_str = NULL;
+static const char* _crengine_args[32] = { NULL, NULL, NULL };
```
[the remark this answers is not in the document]
rvansa: Yes. First is in place of the binary; second is the operation (`checkpoint`/`restore`) and we want the third position initialized to NULL - when add_crengine_arg is called it should find a NULL and paste the value in there, initializing the 'tail' to NULL.

src/hotspot/os/linux/os_linux.cpp
rvansa: The fewer static vars the better IMO, but I can certainly use an explicit index var and drop the initialization to three NULLs if you think it's better.

AntonKozlov: In general I agree, but since _crengine_args is already stateful, I don't see any troubles adding more statics that is in fact a part of the same state. But this is minor, so up to you. OK for me as it is.

AntonKozlov (approved):

rvansa:
Updated using the index var.

wkia:
It seems like `jdk/crac/recursiveCheckpoint/Test.java` starts failing with timeout after this PR. Have you tested it locally?

As far as I see, the test is run with a comma added to CR engine name ("`-XX:CREngine=pauseengine,`")
And then, `pauseengine `cannot recognize `imagedir` name, because `arg[2]` is empty.

Can you reproduce this locally?

test/lib/jdk/test/lib/crac/CracBuilder.java
```diff
@@ -331,7 +335,9 @@ private List<String> prepareCommand(List<String> javaPrefix, boolean isRestore)
         }
         cmd.add("-ea");
         if (engine != null) {
-            cmd.add("-XX:CREngine=" + engine.engine);
+            String engArgs = engineArgs == null ? "" : "," + Arrays.stream(engineArgs)
```
wkia, line 338: I suppose `engineArgs == null` should be replaced with something like `engineArgs == null || 0 == engineArgs.length`

rvansa: You're right, I must have had something cached because when I re-ran the test it was passing... until I added some debug logs to confirm what you're seeing. I'll file a fix in a sec.

PR #64, opened 2023-05-10
Make CheckpointException/RestoreException aggregate-only

rvansa:
CheckpointException and RestoreException (both jdk.crac and javax.crac) cannot carry any message, cause and don't collect stack trace. The sole purpose of these is to aggregate other exceptions in suppressed exceptions.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/64/head:pull/64` \
`$ git checkout pull/64`

`$ git checkout pull/64` \
`$ git pull https://git.openjdk.org/crac.git pull/64/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 64`

`$ git pr show -t 64`

</details>

<a href="https://git.openjdk.org/crac/pull/64.diff">https://git.openjdk.org/crac/pull/64.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/64#issuecomment-1542119604)


src/java.base/share/classes/javax/crac/CheckpointException.java
```diff
@@ -36,6 +36,18 @@ public class CheckpointException extends Exception {
...
+        super(null, null, true, false);
+    }
+
+    /**
+     * Constructs a {@code CheckpointException} with the specified
+     * detail message.
+     *
+     * @param message the detail message.
+     */
+    public CheckpointException(String message) {
```
AntonKozlov, line 50: What if we remove this constructor and hide the other one? https://github.com/openjdk/crac/pull/60/files#r1190070472

rvansa: First of all, I think that `javax.crac` should mirror `jdk.crac` API- and docs-wise. It will be much easier when everyone will be able to just change the imports.

About the constructor with message: I find a bit confusing when an exception is thrown because of some problem with `criu` but there's no actionable message. I have added a simple 'Native checkpoint failed' but we should probably point user to the dump4.log file. (`criuengine` should also make some sanity checks on permissions but that's another thing). I wouldn't object to hiding it, though.

About the one without: `Context` narrows the `throws` to CE/RE and since we expect users to implement Context this would give them no chance to throw checked exceptions, not even the aggregating one. Hiding that won't work.

AntonKozlov: Yes, that is a bug javax.crac does not follow jdk.crac.

Agree about the constructor without args.

And still want to delete the a constructor with message. We can introduce a jdk.crac.impl.CheckpointMessageException (along j.c.i.CheckpointOpenResourceException) which we use to communicate different reason(s) checkpoint is not successful.

rvansa: What about turning the inheritance the other way: final CheckpointAggregateException extends CheckpointException? This would be easily distinguishable, no need to change anything on Context interface. The no-arg constructor in CheckpointException would be protected.
It's kind of natural that exceptions carry messages.

AntonKozlov: What will be the point of CheckpointException then? A more specific exception will also be preferable instead of that, isn't it?

rvansa: `Core` would throw a generic exception when something failed in the native part (`criuengine` returned exit code 1...). A custom implementation of `Context` would throw it when it can't call its children for some reason (but it's not a failure in the Resource itself).

AntonKozlov: But that does not take away the need for the base CheckpointException to be able aggregate exceptions from Resources, so CheckpointException and Combined are mostly overlaping. Moreover,  so we'll need to be able to re-aggregate exceptions of two types. In the current form, Combined and CheckpointException are both fine to be thrown from a user context, and there is no guadance which one to use.

Let's keep it simple and use CheckpointException for aggregations as it was supposed to be. The message constructor should be removed, and every constructor-with-the-message call should be replaced with an exception of different type. Especially since we have examples for this [1].

[1] https://github.com/openjdk/crac/blob/6f403eaf7655f9ce4d10da25082a836d0ad1574c/src/java.base/share/classes/jdk/crac/Core.java#L164

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -155,18 +142,14 @@ private static void checkpointRestore1(long jcmdStream) throws
         if (retCode != JVM_CHECKPOINT_OK) {
             if (checkpointException == null) {
                 checkpointException = new CheckpointException();
+                if (messages.length == 0) {
+                    checkpointException.addSuppressed(new RuntimeException("Native checkpoint failed"));
+                }
```
AntonKozlov, line 147: codes.length == messages.length, and they are used to communicate native descriptors problems (handled in translateJVMExceptions). So a problem should be reported based on the retCode below, if that is not reported already (?)

AntonKozlov (commented):
Few minor comments

src/java.base/share/classes/javax/crac/CheckpointException.java
AntonKozlov: Thank you! I think the PR is mostly good, see the comments.

src/java.base/share/classes/jdk/crac/Core.java
rvansa: When the checkpoint does not fail in JVM code (checking open FDs) but the engine returns non-zero status, the retCode is `JVM_CHECKPOINT_ERROR` but the messages are empty. In this case the exception would look just like CheckpointException without any explanation nor stack trace, and the user would be rather clueless.
We could add to this message a hint to check out `dump4.log`, but I am not entirely sure about this as it breaks encapsulation of CR engine (alternative implementations... though we might make CRIU use 'standardized' location) and it might not exist at all, e.g. when the CRIU binary does not exist at all.

AntonKozlov: Regarding providing more details, in general, that should be handled by the engine. The criuengine (as the entity controlling the dump4.log name and the path) has a chance to print a few lines from the log. E.g. by calling `system("tail ...")`, although not very pretty, for that will be better than nothing.

src/java.base/share/classes/jdk/crac/CheckpointException.java
```diff
@@ -36,16 +36,8 @@ public class CheckpointException extends Exception {
      * Creates a {@code CheckpointException}.
```
AntonKozlov, line 36: I probably misclicked "Add comment",

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Throwable.html#%3Cinit%3E(java.lang.String,java.lang.Throwable,boolean,boolean):
> Subclasses of Throwable should document any conditions under which suppression is disabled and document conditions under which the stack trace is not writable.

The javadoc should specify the stack trace is not writeable, here and for other exceptions

AntonKozlov (commented):
Could you please update the PR titile to match the new content?

AntonKozlov (approved):
Thank you! LGTM

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -81,25 +81,20 @@ private Core() {
     private static void translateJVMExceptions(int[] codes, String[] messages,
                                                CheckpointException exception) {
         assert codes.length == messages.length;
-        final int length = codes.length;
+        // When the CR engine fails (e.g. due to permissions, missing binaries...)
+        // there are no messages recorded, but the user would end up with an empty
+        // CheckpointException without stack trace nor message.
+        if (codes.length == 0) {
+            exception.addSuppressed(new RuntimeException("Native checkpoint failed."));
+        }
```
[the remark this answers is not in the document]
rvansa: Shouldn't a dry-run return `JVM_CHECKPOINT_OK`? Which test is that, `DryRunTest` works for me?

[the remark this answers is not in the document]
AntonKozlov: The test does not cover this case: when a Resource throws something and we go to dry run JVM inspection, but the inspection does not find anything. 

Turns out, ContextOrderTest fails bacuse the same reason.

JVM dry-run now always returns CHECKPOINT_ERROR by implementation. But in general the the code should depend on the inspection result: OK if no offending descriptors found, ERROR (better to be FD_ERROR) when there are open file descriptors.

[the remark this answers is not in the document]
rvansa: Fixed in https://github.com/openjdk/crac/pull/80

PR #66, opened 2023-05-12
Register CallSite cleaners with higher priority

rvansa:
Cleaners for CallSites cannot be registered with regular CLEANERS priority as this would hang/throw exceptions any time a lambda or method reference is used during or after processing this resource priority class; therefore we postpone as the last priority.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/66/head:pull/66` \
`$ git checkout pull/66`

`$ git checkout pull/66` \
`$ git pull https://git.openjdk.org/crac.git pull/66/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 66`

`$ git pr show -t 66`

</details>

<a href="https://git.openjdk.org/crac/pull/66.diff">https://git.openjdk.org/crac/pull/66.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/66#issuecomment-1545561779)

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
@@ -27,10 +27,13 @@ package java.lang.invoke;
 
 import jdk.internal.access.JavaLangAccess;
 import jdk.internal.access.SharedSecrets;
+import jdk.internal.crac.JDKResource;
 import jdk.internal.ref.CleanerFactory;
+import jdk.internal.ref.CleanerImpl;
 import sun.invoke.util.Wrapper;
 
 import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.ref.Cleaner;
 import java.lang.reflect.Field;
 
 import static java.lang.invoke.MethodHandleNatives.Constants.*;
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
```diff
@@ -87,7 +90,8 @@ static CallSiteContext make(CallSite cs) {
             // become unreachable, its Context is retained by the Cleanable instance
             // (which is referenced from Cleaner instance which is referenced from
             // CleanerFactory class) until cleanup is performed.
-            CleanerFactory.cleaner().register(cs, newContext);
+            new CleanerImpl.PhantomCleanableRef(cs, CleanerFactory.cleaner(),
+                    newContext, JDKResource.Priority.CALL_SITES);
```
AntonKozlov, line 94: This is based on impl details. Is it possible to use e.g. a `CleanerFactory.criticalCleaner().register(...)` that will use different priorities?

AntonKozlov (approved):
LGTM as a workaround, but indeed we need to generalize the problem and think about possible solutions.

src/java.base/share/classes/java/lang/invoke/MethodHandleNatives.java
rvansa: 1) `CleanerFactory` could get a new method since it is in `jdk.internal.ref`, but such alternative implementation would spin up another thread.
2) `Cleaner` is a public interface, so I cannot add an alternative method. Regrettably it's also a final class, so I can't make CleanerFactory wrap the common cleaner and override just the `register()` method.

I find that the info that we need to do something different already leaks to MethodHandleNatives and since your suggested solution has impact on memory usage, I would rather choose breaking the encapsulation.

PR #69, opened 2023-05-12
Handle open file descriptors with configurable policies

rvansa:
When the application does not close some file descriptors through Resources we can use `jdk.crac.resource-policies` pointing to a configuration file that might adjust the behaviour. The file uses a simple YAML-conformant format with individual rules separated by a line with `---`:

```
type: file
path: /path/to/*.txt
action: close
---
type: socket
localAddress: 127.0.0.1
localPort: 8080
action: ignore
```

Available types:
* `file`: supports `path` with 'glob' pattern matching (see FileSystem.getPathMatcher() for details)
* `pipe`: anonymous pipes (named pipe is handled as `file`)
* `socket`: can be refined using `family`, `localAddress`, `localPort`, `localPath` (in case of Unix sockets), `remoteAddress`, `remotePort` and `remotePath`

Actions depend on each resource, defaulting to `error`, with common options `close` and `ignore`. Files have `reopen` action, too.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/69/head:pull/69` \
`$ git checkout pull/69`

`$ git checkout pull/69` \
`$ git pull https://git.openjdk.org/crac.git pull/69/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 69`

`$ git pr show -t 69`

</details>

<a href="https://git.openjdk.org/crac/pull/69.diff">https://git.openjdk.org/crac/pull/69.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/69#issuecomment-1545565907)


jankratochvil:
CRaC now started complaining to me when I try to snapshot: https://github.com/CRaC/example-jetty
```
Suppressed: jdk.crac.impl.CheckpointOpenFileException: FileDescriptor 12 left open: tcp6 local /[0:0:0:0:0:0:0:0]:8080 remote not bound 
```

1. I haven't found how to specify a property. I have tried below the normal `-D` command line parameter, is that correct?
2. But specifying the command line parameter has no effect. It still no longer works:
```
$ (cd ../example-jetty/;. JAVA_HOME-crac-git-slowdebug;(sleep 4;sudo rm -f /tmp/coredump.*;sudo jcmd target/example-jetty-1.0-SNAPSHOT.jar JDK.checkpoint) & sudo rm -rf cr-host2/;sudo bash -c 'ulimit -c unlimited;. JAVA_HOME-crac-git-slowdebug;$JAVA_HOME/bin/java -XX:CRaCCheckpointTo=cr-host2 -Djdk.crac.collect-fd-stacktraces=true -Djdk.crac.fd-policy.checkpoint=CLOSE -jar target/example-jetty-1.0-SNAPSHOT.jar';echo restore;sudo $JAVA_HOME/bin/java -XX:CRaCRestoreFrom=cr-host2)
2023-06-02 13:43:57.835:INFO::main: Logging initialized @2712ms to org.eclipse.jetty.util.log.StdErrLog
2023-06-02 13:43:58.647:INFO:oejs.Server:main: jetty-9.4.30.v20200611; built: 2020-06-11T12:34:51.929Z; git: 271836e4c1f4612f12b7bb13ef5a92a927634b0d; jvm 17-internal+0-adhoc.azul.crac-git
1519:
2023-06-02 13:43:59.647:INFO:oejs.AbstractConnector:main: Started ServerConnector@55a1c291{HTTP/1.1, (http/1.1)}{0.0.0.0:8080}
2023-06-02 13:43:59.735:INFO:oejs.Server:main: Started @4744ms
Jun 02, 2023 1:44:00 PM jdk.internal.crac.LoggerContainer info
INFO: /home/azul/azul/example-jetty/target/dependency/jetty-io-9.4.30.v20200611.jar is recorded as always available on restore
Jun 02, 2023 1:44:00 PM jdk.internal.crac.LoggerContainer info
INFO: /home/azul/azul/example-jetty/target/dependency/jetty-util-9.4.30.v20200611.jar is recorded as always available on restore
Jun 02, 2023 1:44:00 PM jdk.internal.crac.LoggerContainer info
INFO: /home/azul/azul/example-jetty/target/dependency/jetty-http-9.4.30.v20200611.jar is recorded as always available on restore
Jun 02, 2023 1:44:00 PM jdk.internal.crac.LoggerContainer info
INFO: /home/azul/azul/example-jetty/target/dependency/javax.servlet-api-3.1.0.jar is recorded as always available on restore
Jun 02, 2023 1:44:00 PM jdk.internal.crac.LoggerContainer info
INFO: /home/azul/azul/example-jetty/target/dependency/jetty-server-9.4.30.v20200611.jar is recorded as always available on restore
Jun 02, 2023 1:44:00 PM jdk.internal.crac.LoggerContainer info
INFO: /home/azul/azul/example-jetty/target/example-jetty-1.0-SNAPSHOT.jar is recorded as always available on restore
An exception during a checkpoint operation:
jdk.crac.CheckpointException
        at java.base/jdk.crac.Core.checkpointRestore1(Core.java:129)
        at java.base/jdk.crac.Core.checkpointRestore(Core.java:264)
        at java.base/jdk.crac.Core.checkpointRestoreInternal(Core.java:280)
        Suppressed: jdk.crac.impl.CheckpointOpenFileException: FileDescriptor 10 left open: tcp6 local /[0:0:0:0:0:0:0:0]:8080 remote not bound
                at java.base/java.io.FileDescriptor.beforeCheckpoint(FileDescriptor.java:391)
                at java.base/java.io.FileDescriptor$Resource.beforeCheckpoint(FileDescriptor.java:84)
                at java.base/jdk.crac.impl.PriorityContext$SubContext.invokeBeforeCheckpoint(PriorityContext.java:107)
                at java.base/jdk.crac.impl.OrderedContext.runBeforeCheckpoint(OrderedContext.java:70)
                at java.base/jdk.crac.impl.AbstractContextImpl.beforeCheckpoint(AbstractContextImpl.java:81)
                at java.base/jdk.crac.impl.AbstractContextImpl.invokeBeforeCheckpoint(AbstractContextImpl.java:41)
                at java.base/jdk.crac.impl.PriorityContext.runBeforeCheckpoint(PriorityContext.java:70)
                at java.base/jdk.crac.impl.AbstractContextImpl.beforeCheckpoint(AbstractContextImpl.java:81)
                at java.base/jdk.internal.crac.JDKContext.beforeCheckpoint(JDKContext.java:97)
                at java.base/jdk.crac.impl.AbstractContextImpl.invokeBeforeCheckpoint(AbstractContextImpl.java:41)
                at java.base/jdk.crac.impl.OrderedContext.runBeforeCheckpoint(OrderedContext.java:70)
                at java.base/jdk.crac.impl.AbstractContextImpl.beforeCheckpoint(AbstractContextImpl.java:81)
                at java.base/jdk.crac.Core.checkpointRestore1(Core.java:127)
                ... 2 more
        Caused by: java.lang.Exception: This file descriptor was created by main at epoch:1685706239263 here
        at java.base/java.io.FileDescriptor$Resource.<init>(FileDescriptor.java:75)
                at java.base/java.io.FileDescriptor.<init>(FileDescriptor.java:104)
                at java.base/sun.nio.ch.IOUtil.newFD(IOUtil.java:544)
                at java.base/sun.nio.ch.Net.serverSocket(Net.java:534)
                at java.base/sun.nio.ch.ServerSocketChannelImpl.<init>(ServerSocketChannelImpl.java:128)
                at java.base/sun.nio.ch.ServerSocketChannelImpl.<init>(ServerSocketChannelImpl.java:109)
                at java.base/sun.nio.ch.SelectorProviderImpl.openServerSocketChannel(SelectorProviderImpl.java:72)
                at java.base/java.nio.channels.ServerSocketChannel.open(ServerSocketChannel.java:145)
                at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:339)
                at org.eclipse.jetty.server.ServerConnector.open(ServerConnector.java:310)
                at org.eclipse.jetty.server.AbstractNetworkConnector.doStart(AbstractNetworkConnector.java:80)
                at org.eclipse.jetty.server.ServerConnector.doStart(ServerConnector.java:234)
                at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
                at org.eclipse.jetty.server.Server.doStart(Server.java:386)
                at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
                at com.example.ServerManager.<init>(App.java:27)
                at com.example.App.main(App.java:73)
restore
open cppath: No such file or directory
```
I am curious it says `Suppressed:` there but the snapshot still crashes due to `An exception during a checkpoint operation:`.
3. The setting `-Djdk.crac.fd-policy.checkpoint=CLOSE` would be wrong anyway. I do not want to silently close some ongoing TCP client-server communication. But in this my case a socket in LISTEN state could be safely automatically restored without any coordination from the Java application: `tcp6 local /[0:0:0:0:0:0:0:0]:8080 remote not bound`

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -397,9 +422,62 @@ private synchronized void beforeCheckpoint() throws CheckpointOpenFileException
...
+                } else if (resource.originalType.equals("socket")) {
+                    throw new RestoreException("Cannot reopen file descriptor " +
+                            resource.originalFd + ": cannot restore socket");
+                }
+                path = resource.originalPath;
+            }
+            // We will attempt to open at the original offset even if the path changed;
+            // this is used probably as the file moved on the filesystem but the contents
+            // are the same.
+            if (!reopen(resource.originalFd, path, resource.originalFlags, resource.originalOffset)) {
```
AntonKozlov, line 456: Another option is a log file in the append mode, which may grow larger -- in that case we'd like to have that in append mode with position at the end. Probably that is handled by setting proper flags, but at least this would contradict with the comment.

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -61,6 +60,11 @@ public final class FileDescriptor {
     private boolean closed;
 
     class Resource implements jdk.internal.crac.JDKResource {
+        private int originalFd = -1;
+        private String originalType;
+        private String originalPath;
+        private int originalFlags;
+        private long originalOffset;
```
AntonKozlov, line 67: Regarding the design, path and offset are features of a file desriptor referring to the filesystem, but they do not make sense to socket FD.

src/java.base/share/classes/jdk/crac/impl/OpenFDPolicies.java
```diff
@@ -0,0 +1,307 @@
...
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.*;
+import java.util.regex.Pattern;
+
+import static jdk.crac.impl.OpenFDPolicies.AfterRestore.OPEN_OTHER;
+
+public class OpenFDPolicies<P> {
+    public static final String CHECKPOINT_PROPERTY = "jdk.crac.fd-policy.checkpoint";
+    public static final String RESTORE_PROPERTY = "jdk.crac.fd-policy.restore";
```
AntonKozlov, line 23: Having separated policies for checkpoint and restore enables some weird configuration, when e.g. checkpoint specifing CLOSE and restore -- REOPEN. It would be better to have a combined, consisted checkpoint-restore policy that specifes both parts at once.

src/java.base/share/classes/jdk/crac/impl/OpenFDPolicies.java
```diff
@@ -0,0 +1,307 @@
...
+            for (var entry : pathPolicies) {
+                if (entry.getKey().matches(p)) {
+                    return entry.getValue();
+                }
+            }
+        }
+        if (type.equals("fifo") && fd > 2) {
+            return fifoPolicy;
+        } else if (type.equals("socket")) {
+            return socketPolicy;
```
AntonKozlov, line 228: Obviously, only a single policy is possible for all sockets. In general sockets are much harder than files, and in not so many cases we can automatically handle them. I think some connection-less sockets _may_ be covered, and some listening sockets.

If we continue this implementation, FileDescriptor (a rather simple object intially) will grow larger and larger, knowing about all possible uses in the JDK.

src/java.base/share/classes/java/io/FileDescriptor.java
rvansa: Yes, you could add special classes for Sockets vs. Files. It's easier, though, to treat this in single POJO and keep the path `null` if it does not make sense in this case.

src/java.base/share/classes/jdk/crac/impl/OpenFDPolicies.java
rvansa: I've started with a single policy enum but it turned out the inlined cross product of behaviours was rather a long and repetitive.
In fact CLOSE + REOPEN is the combination that makes a perfect sense. You close the file on checkpoint (rather than error out and fail the checkpoint) and then you reopen the same file (rather than opening something else).

src/java.base/share/classes/jdk/crac/impl/OpenFDPolicies.java
rvansa: Someone already asked for automatic reopening of listening sockets. The policy could then e.g. filter specific ports (similar to files). Yes, you can't to anything but KEEP_CLOSED for sockets related to connection. I assume that the restore policy will be set during restore, rather than during checkpoint, so there's no chance to handle a non-sense configuration during checkpoint (because the configuration is not present yet).

Growing FileDescriptor has a point, any complex implementation for sockets should be hosted next to other Socket related stuff. For files, I don't know if there's a better place, but we could refactor it to an util class.

src/java.base/share/classes/jdk/crac/impl/OpenFDPolicies.java
rvansa: And note that the restore configuration should be set on restore, because in case of opening other files you don't know where these are in the deployment (restore) environment during checkpoint. Had you specified the policy together you would be overriding the behaviour on checkpoint which was already executed, which is even more confusing.

AntonKozlov: > I've started with a single policy enum but it turned out the inlined cross product of behaviours was rather a long and repetitive.

Do you have artifacts of that? Because it may mean that cross-product may have a few invalid / unsafe behaviors. It would be interesting to look at the list.

> In fact CLOSE + REOPEN is the combination that makes a perfect sense. 

Indeed, bad example. What about ERROR+OPEN_OTHER. Or for the sockets, for the current implementation we should not allow OPEN_OTHER, as the implementation can only throw RestoreException in this case.

AntonKozlov: > And note that the restore configuration should be set on restore, because in case of opening other files you don't know where these are in the deployment (restore) environment during checkpoint. Had you specified the policy together you would be overriding the behaviour on checkpoint which was already executed, which is even more confusing.

The combined policy can have additional parameters settable on restore. But the combined policy should be able to limit policies or parameters for restore part of the policy.

src/java.base/share/classes/java/io/FileDescriptor.java
rvansa: What I was trying to say that it's not about reading the 'append' flag when the file was opened, because that was just saying "let's open and then seek to the end of the file". We shouldn't conclude that we want to start at the 'new end' (provided that the file was modified) on restore.
Having a policy that will reopen at end, regardless of original offset, is certainly a valid option.

src/java.base/share/classes/jdk/crac/impl/OpenFDPolicies.java
rvansa: When the behaviour is ERROR, you won't have a chance to use OPEN_OTHER, because you never get to restore.

If you set policy OPEN_OTHER on FD 42, the restore will throw an exception. We could completely fail the restore and exit, too. So yes, we do not allow that.
We could fail when already reading policies that look like SOCKET=OPEN_OTHER:..., why not add an extra check. But failing when we actually try to do that is only a fraction of second later.

src/java.base/share/classes/jdk/crac/impl/OpenFDPolicies.java
AntonKozlov: > Someone already asked for automatic reopening of listening sockets.

It was me :) 

> The policy could then e.g. filter specific ports (similar to files). Yes, you can't to anything but KEEP_CLOSED for sockets related to connection.

Connection-less sockets may be fine as well. Maybe even stream sockets, with a big disclaimer that the policy administrator is fully responsible for breaking communication logic. By default, we should do the most safest thing. 

> I assume that the restore policy will be set during restore, rather than during checkpoint, so there's no chance to handle a non-sense configuration during checkpoint (because the configuration is not present yet).

I think this can be implemented https://github.com/openjdk/crac/pull/69#discussion_r1224461715

> Growing FileDescriptor has a point, any complex implementation for sockets should be hosted next to other Socket related stuff. For files, I don't know if there's a better place, but we could refactor it to an util class.

#79 does that with JDKFIleResource, which provides a limited hard-coded poilcy for files: exceptions are thrown only for files not in classpath.

rvansa:
Updated:
* merged-in last changes from `crac` branch
* added `REOPEN_AT_END` and `OPEN_OTHER_AT_END` policies
* refactored all the policy handling code into FileDescriptorResource (akin to JDKFileResource, name is not too important)

As for the checkpoint policy, the ERROR one probably won't be used (it's the default), so until we support some form of not closing the FD at all (replication use case) anything the user would set would boil down to CLOSE/WARN_CLOSE. Therefore, if you want to set it in a one big policy this would probably mean brain-dead duplication of the options.

If you want to present a few examples of how you'd like the user-facing options to look like, I am listening, but I don't see a better way myself.

rvansa:
Merged in `crac` branch, and applied requested changes:
* The detection and policies now live in JDKFileResource and JDKSocketResource
* Policies for files and sockets are split

One thing I am not so sure about is location of some native methods - I've left what's used in `JDKFileResource` in `FileDispatcher_md.c`, and `JDKSocketResource` uses methods both in `UnixDomainSockets.c` and some are in `Net.c`. This might deserve own files, but I've reused what was in those files.

src/java.base/share/classes/jdk/crac/impl/OpenFilePolicies.java
```diff
@@ -0,0 +1,245 @@
...
+         * declaration is opened instead; the policy name should be followed
+         * by an equal sign '=' and the new path.
+         * If the other file cannot be open the checkpoint throws an exception.
+         */
+        OPEN_OTHER,
+        /**
+         * Similar to {@link #OPEN_OTHER} but ignored the previous offset
+         * and opens the file at the end.
+         */
+        OPEN_OTHER_AT_END,
```
AntonKozlov, line 220: Semantically, these modes are the same, the second is for FD opened in APPEND mode. But having both modes allows changing the FD mode for the application. Such change will break app assumptions about the file descriptor.

src/java.base/share/classes/java/io/FileOutputStream.java
```diff
@@ -457,15 +459,20 @@ public FileChannel getChannel() {
...
+    private final JDKFileResource resource = new JDKFileResource(this) {
         @Override
         protected FileDescriptor getFD() {
             return fd;
         }
 
         @Override
-        protected String getPath() {
-            return path;
+        protected Supplier<Exception> claimException(int fdNum, String path) {
```
AntonKozlov, line 470: Here and everywhere else, please use platform independent `FileDescriptor`.

src/java.base/share/classes/java/io/FileInputStream.java
```diff
@@ -516,6 +520,23 @@ public FileChannel getChannel() {
...
+        @Override
+        protected Supplier<Exception> claimException(int fdNum, String path) {
+            if (fd == FileDescriptor.in) {
+                return null;
+            } else {
+                return super.claimException(fdNum, path);
+            }
+        }
+    };
+
```
AntonKozlov, line 539: Providing full range of policies for FileInputStreams seems inherently unsafe, e.g. *_AT_END. I also think we need to distinguish FileInputStreams created with a path or with fd.

AntonKozlov (commented):
Some random notes from first reading, I'll do another pass. This is not a request to change anything.

src/java.base/share/classes/java/net/DatagramSocketImpl.java
```diff
@@ -72,6 +74,10 @@ public DatagramSocketImpl() {}
      */
     protected FileDescriptor fd;
 
+    // We don't know the protocol family when this socket is created and FD allocated, but it's not UNIX
+    @SuppressWarnings("unused")
+    private final JDKSocketResource resource = new JDKSocketResource(this, StandardProtocolFamily.INET, () -> fd);
```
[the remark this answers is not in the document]
rvansa: The family here is used to distinguish network sockets vs. unix sockets; IPv4 and IPv6 is handled identically. I have considered making the arg just boolean, but using `family` seemed more future proof. POSIX defines many more types...
In fact the IPv4/IPv6 is decided based on a static call on IPv6 support, so I could use a static helper method (would have to expose IPv6 availability, `Net.isIPv6Available` is package-private).

src/java.base/share/classes/jdk/crac/impl/OpenFilePolicies.java
rvansa: There's many ways that you could break an application invariant if the file is changed/removed when the application thinks it has exclusive write access to it, e.g. it might not expect to write files bigger than X because it streams at most X bytes to that, but if you change the position during reopening you might violate that expectation. These are different options to handle that case: if you think the simplicity of use outweighs the choice I can make OPEN_OTHER use the _AT_END semantics when the file was opened for appending.

src/java.base/share/classes/java/io/RandomAccessFile.java
```diff
@@ -81,16 +81,11 @@ public class RandomAccessFile implements DataOutput, DataInput, Closeable {
...
         @Override
         protected FileDescriptor getFD() {
             return fd;
         }
-
-        @Override
-        protected String getPath() {
-            return path;
-        }
     };
```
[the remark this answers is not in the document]
rvansa: The symlink has a point, from the 'application intention' view. On the other hand if the file is changed; there's a higher chance that something unexpected will be read.
In any case user has the chance to override the target in policies.

AntonKozlov (commented):
First batch of the comments

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -67,14 +67,15 @@ public void beforeCheckpoint(Context<? extends jdk.crac.Resource> context) throw
...
-                    return new CheckpointOpenResourceException(
-                        FileDescriptor.class.getSimpleName() + " " + fd + ": " + nativeDescription0(),
-                        getStackTraceHolder());
-                });
+                // Normally the claiming should be overridden by FileInputStream/FileOutputStream
+                // but in case these are collected we handle FDs 0..2 here as well.
+                if (self == in || self == out || self == err) {
+                    claimedFDs.claimFd(self, self, null);
+                }
+
```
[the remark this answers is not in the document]
rvansa: How would you expect to specify *which* FD should not throw, without fetching the details through native code? Would a numeric identifier (or full wildcard) be sufficient?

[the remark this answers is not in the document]
rvansa: Alternatively, we could change the claiming process to create a FD -> higher resource reference when the FD is accepted in the resource (constructor or setter), preventing this altogether from happening. Regrettably one FD would have to be able to reference multiple resources this way, though.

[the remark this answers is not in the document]
AntonKozlov: I'd propose a numeric id, and ideally a regex for the native description, without trying to interpret that on the FD level.

Another use-case for these FD policies is to work around a forgotten claim from a higher-level Java object, which should be considered as a JDK bug. It will be uncommon and will be discouraged, but anyway, worth to have as a workaround.

src/java.base/share/classes/jdk/internal/crac/JDKFileResource.java
```diff
@@ -34,20 +42,44 @@ public void beforeCheckpoint(Context<? extends Resource> context) throws Excepti
...
-        }
-
+        OpenResourcePolicies.Policy policy = OpenResourcePolicies.findForPath(false, path);
+        String action = policy != null ? policy.action.toLowerCase() : "error";
+        Supplier<Exception> exceptionSupplier = switch (action) {
+            case "error":
+                error = true;
+                // Files on the classpath are considered persistent, exception is not thrown
+                yield matchClasspath(path) ? NO_EXCEPTION :
+                        () -> new CheckpointOpenFileException(path, getStackTraceHolder());
```
AntonKozlov, line 52: So error + no exception is a valid combination? Should not we choose `ignore` as the default for files matching classpath instead?

src/java.base/share/classes/jdk/internal/crac/JDKSocketResourceBase.java
```diff
@@ -0,0 +1,160 @@
...
+                return true;
+            }
+        };
+    }
+
+    @SuppressWarnings("fallthrough")
+    @Override
+    public void afterRestore(Context<? extends Resource> context) throws Exception {
+        // Don't do anything when we've already failed
+        if (originalFd < 0 || error) {
```
AntonKozlov, line 138: The orignalFd seems to be a leftover, it's not clear why is it stored as a part of the state. Should not we just consider `error`?

src/java.base/share/classes/jdk/internal/crac/JDKFileResource.java
```diff
@@ -34,20 +42,44 @@ public void beforeCheckpoint(Context<? extends Resource> context) throws Excepti
...
-        Supplier<Exception> exceptionSupplier;
-        if (matchClasspath(path)) {
-            // Files on the classpath are considered persistent, exception is not thrown
-            exceptionSupplier = () -> null;
-        } else {
-            exceptionSupplier = () -> new CheckpointOpenFileException(path, getStackTraceHolder());
-        }
-
+        OpenResourcePolicies.Policy policy = OpenResourcePolicies.findForPath(false, path);
+        String action = policy != null ? policy.action.toLowerCase() : "error";
```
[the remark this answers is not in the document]
rvansa: The policies are meant to be resource-specific, therefore I didn't want to use one big enum in `OpenResourcePolicies`. It's already a stretch from the concept that `FILE`, `PIPE` and `SOCKET` are defined there.
So we could have them as constants in the resource parent, but these would be mostly used just once.
FYI, I am moving OpenResourcePolicies.findForPath to JDKFIleResource, it's not used elsewhere anyway.

src/java.base/share/classes/jdk/internal/crac/JDKSocketResourceBase.java
rvansa: Functionally it's used to mark the socket as invalid, and I've kept the FD number for debugging, but I can remove that...

src/java.base/unix/classes/sun/nio/ch/FileDispatcherImpl.java
```diff
@@ -58,7 +58,10 @@ public void afterRestore(Context<? extends Resource> context)
     static {
         IOUtil.load();
         init();
-        Core.Priority.NORMAL.getContext().register(resourceProxy);
+        // We cannot register using normal priority because other JDK resources
+        // might read configuration files with this or later priority.
+        // It's difficult to trigger static initialization outside the package.
+        Core.Priority.PRE_FILE_DESCRIPTORS.getContext().register(resourceProxy);
```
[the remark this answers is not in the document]
rvansa: The problem with registration at NORMAL level is that this is the same level as OpenResourcePolicies registers to, in order to fetch the configuration just before the checkpoint (to allow dynamic configuration from within the app). When ORP.beforeCheckpoint opens the configuration file and this is the first moment FDI is used (triggering the static initialization) this results in FDI registering with the priority that is already being executed - depending on the context impl this either blocks or the FDI resourceProxy is not called.
Ideally I would trigger clinit for FDI from FileDescriptor and ensure that it's already registered before checkpoint, but haven't found any simple way to do that.

src/java.base/unix/classes/sun/nio/ch/FileDispatcherImpl.java
```diff
@@ -203,8 +206,10 @@ static void afterRestore() throws IOException {
     // Shared with SocketDispatcher and DatagramDispatcher but
     // NOT used by FileDispatcherImpl
     static void closeAndMark(FileDescriptor fd) throws IOException {
-        fdAccess.markClosed(fd);
-        close0(fd);
+        // Originally this used fdAccess.markClosed() and close0() but leaving
+        // the FD value set breaks JDKSocketResource (we don't want the extra
+        // test if the FD resource has been marked).
+        fdAccess.closeNoCleanup(fd);
```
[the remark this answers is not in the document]
rvansa: Do you mean this thread? https://mail.openjdk.org/pipermail/nio-dev/2023-January/013075.html

Well this was meant to change the behaviour as little as possible. All the ways the FD can be closed looks pretty spaghetti to me; I can give the regular close a go and see what breaks...

src/java.base/share/classes/java/io/FileOutputStream.java
```diff
@@ -467,5 +472,24 @@ protected FileDescriptor getFD() {
...
+
+        @Override
+        protected void reopenAfterRestore(OpenResourcePolicies.Policy policy) throws IOException {
+            synchronized (closeLock) {
+                // We have been writing to a file, but it disappeared during checkpoint
+                if (!Files.exists(Path.of(path))) {
+                    throw new IOException("File " + path + " is not present during restore");
+                }
+                // FileOutputStream writes only forward, so it makes sense to reopen it appending
+                open(path, true);
```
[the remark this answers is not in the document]
rvansa: This does not work; when creating the FileChannel (lazy) as an attempt to read the position I would create another JDKFileResource, and try to register it => this would get blocked as we're already in the FDs priority.

I could probably add in some exceptions to not create it on this codepath, but is it really worth it? I thought that the idea was to work we already have in the resource. And there's no difference in behaviour unless someone appends to that file during checkpoint.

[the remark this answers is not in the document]
AntonKozlov: Implicit change of append mode from false to true will surprise users of the channel. E.g. after automatic re-open, the position setting won't take effect anymore, while it was possible if the FOS was created in non-appending mode.

I'm fine to implement automatic re-opening only FOS originally opened with append. In general, re-opening a file in non-append will also require (optional) verification the file has not been changed, which may be pretty complex, so I don't think it should be implemented now for FOS or other Files.

src/java.base/unix/classes/sun/nio/ch/FileDispatcherImpl.java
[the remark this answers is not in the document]
AntonKozlov: Oh, sorry, I did not mean to make it more dangerous, just in the previous version it was rather inconsistent (the method is called closeAndMark, but apparently no marking was going to happen).

Thank you for the link. Does that mean that NIO does not reset FD just because that is not necessary?

And could you elaborate on why we don't want the extra test for the mark? Since we need more accurate tracking of the FD state, including close() from NIO, but we don't want to change the existing FD behavior, adding the check to JDKSocketResource would be pretty straightforward.

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -61,21 +61,63 @@ public final class FileDescriptor {
...
+                OpenResourcePolicies.Policy policy = findPolicy(nativeDescription);
+                String action = "error";
+                String warn = "false";
+                Supplier<Exception> supplier = null;
+                // Normally the claiming should be overridden by FileInputStream/FileOutputStream
+                // but in case these are collected we handle FDs 0..2 here as well.
+                if (policy != null) {
+                    action = policy.action;
+                    warn = policy.params.getOrDefault("warn", "false");
```
AntonKozlov, line 80: Please make the warning enabled by default.

src/java.base/unix/classes/sun/nio/ch/FileDispatcherImpl.java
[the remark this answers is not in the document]
rvansa: > Thank you for the link. Does that mean that NIO does not reset FD just because that is not necessary?

I think so. It looks like a bit of stretch to me (writing a non-volatile field is not that costly) but I would think that someone did a benchmark and observed a difference.

> And could you elaborate on why we don't want the extra test for the mark?

To put it a bit into perspective, I have used the marking when only FileDescriptor's JDKResource was involved - one extra mark that would be set and read only by the resource seemed simple enough to do the job without changing the other code paths. But if the JDKSocketResource needs to break the encapsulation and check if (an independent) FD's resource has a mark set in order to treat it differently, the complexity builds up and isn't really worth of skipping the single field write.

src/java.base/share/classes/java/io/FileOutputStream.java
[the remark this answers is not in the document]
rvansa: I was not aware that `lseek` wouldn't work in case the file is re-opened for appending; that really forces us to record the position. Since non-appending mode is the default, I'll try to use another resource with lower priority to create the FileChannel so that it can be used later on by the actual FOS resource.

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -61,21 +61,63 @@ public final class FileDescriptor {
...
                     }
-                    return new CheckpointOpenResourceException(
-                        FileDescriptor.class.getSimpleName() + " " + fd + ": " + nativeDescription0(),
-                        getStackTraceHolder());
-                });
-            }
+                }
+                String regex = params.get("regex");
+                if (regex != null) {
+                    return Pattern.compile(regex).matcher(nativeDescription).matches();
```
AntonKozlov, line 117: I think `.find()` instead `.matches()` will fit better. The matches() looks for the complete match, while find() will allow substring match specification. Complete match specification will still be possible with `^` and/or `$`.

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -61,21 +61,63 @@ public final class FileDescriptor {
...
+                    } catch (NumberFormatException e) {
+                        throw new IllegalArgumentException("Cannot parse file descriptor value '" + value + "'");
                     }
-                    return new CheckpointOpenResourceException(
-                        FileDescriptor.class.getSimpleName() + " " + fd + ": " + nativeDescription0(),
-                        getStackTraceHolder());
-                });
-            }
+                }
+                String regex = params.get("regex");
```
AntonKozlov, line 115: Should it be `"descriptionRegex"`? as now it may appear appear the FD value will be matched as well.

rvansa: FD value is a number, would you think of applying a regex on that?

AntonKozlov: It was not clear what part of the `FileDescriptor 2: character: /dev/pts/43` is matched. If that is not evident for me only, OK.

src/java.base/share/classes/java/io/FileDescriptor.java
```diff
@@ -61,21 +61,63 @@ public final class FileDescriptor {
...
+                supplier = switch (action.toLowerCase()) {
+                    case "error":
+                        yield () -> new CheckpointOpenResourceException(
+                            FileDescriptor.class.getSimpleName() + " " + fd + ": " + nativeDescription,
+                            getStackTraceHolder());
+                    case "close":
+                        close();
+                    case "ignore":
+                        if (Boolean.parseBoolean(warn)) {
+                            LoggerContainer.warn("File descriptor {0} was not closed by the application. Use 'warn: false' in the policy to suppress this message.", fd);
```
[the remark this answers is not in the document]
AntonKozlov: Thank you for removing the code duplication. Could you also print the hint only once? Now it's still
```
Jul 19, 2023 2:50:54 PM jdk.internal.crac.LoggerContainer warn
WARNING: File descriptor 2 was not closed by the application. Use warn: false in the policy to suppress this message.
Jul 19, 2023 2:50:54 PM jdk.internal.crac.LoggerContainer warn
WARNING: File descriptor 1 was not closed by the application. Use warn: false in the policy to suppress this message.
Jul 19, 2023 2:50:54 PM jdk.internal.crac.LoggerContainer warn
WARNING: File descriptor 0 was not closed by the application. Use warn: false in the policy to suppress this message.
```

AntonKozlov (approved):
LGTM! Thank you very much!

PR #70, opened 2023-05-12
Synchronize concurrent clean() in PhantomCleanableRef

rvansa:
Fixes failures in RefQueueTest and JarFileFactoryCacheTest.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/70/head:pull/70` \
`$ git checkout pull/70`

`$ git checkout pull/70` \
`$ git pull https://git.openjdk.org/crac.git pull/70/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 70`

`$ git pr show -t 70`

</details>

<a href="https://git.openjdk.org/crac/pull/70.diff">https://git.openjdk.org/crac/pull/70.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/70#issuecomment-1545791516)

### DIFF
--- a/src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
+++ b/src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
@@ -156,7 +156,9 @@ public final class CleanerImpl implements Runnable {
      * Perform cleaning on an unreachable PhantomReference.
      */
     public static final class PhantomCleanableRef extends PhantomCleanable<Object> implements JDKResource {
-        private final Runnable action;
+        // We are setting this to null when the object is cleaned - using a boolean instead would
+        // increase object size by 8 bytes.
+        private Runnable action;
         private final JDKResource.Priority priority;
 
         /**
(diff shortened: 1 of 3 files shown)


AntonKozlov:
> I think that you won't avoid waiting for the cleanup thread - there would have to be a point in time where we'd switch a global (or per-resource?) flag to enter the blocking behavior (cannot be on all the time, because the checkpoint may never happen), and then wait for the in-flight resources to complete. It would be possible to achieve the synchronization through switching the flag on in the cleaner thread, but that only means waiting for it as well.

Agree, syncronization is required. What about adding a special synchronizing Ref added to the cleaner queue, so checkpoint can only proceed once the ref "cleanup" has been started? And which will block the cleaner until restore complete.

> All in all it seems much more complicated. Are you aware of any pros? 

Asking becauses synchronizing a single cleaner looks more efficient compared to syncronizing each PhantomCleanableRef. The performCleanup() is proposed to have an additional synchronized section, which is required only for CRaC.

rvansa:
It seems that the general approach in #73 would be favored, though it's a more complicated one. Feel free to reopen.

PR #72, opened 2023-05-14
Linux file system watcher support

joeyleeeeeee97:
inotify monitors changes on filesystem, this support automatic restore for LinuxFileWatcher.

FileWatcherAfterRestoreTest verifies watcher service works after restore.
FileWatcherTest verifies automatic closing inotiify fd

The watcher keys are still managed by user, so exception will be thrown if no watcher keys are leaked, as in FileWatcherWithOpenKeysTest

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/72/head:pull/72` \
`$ git checkout pull/72`

`$ git checkout pull/72` \
`$ git pull https://git.openjdk.org/crac.git pull/72/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 72`

`$ git pr show -t 72`

</details>

<a href="https://git.openjdk.org/crac/pull/72.diff">https://git.openjdk.org/crac/pull/72.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/72#issuecomment-1546861808)

### DIFF
--- a/src/java.base/linux/classes/sun/nio/fs/LinuxWatchService.java
+++ b/src/java.base/linux/classes/sun/nio/fs/LinuxWatchService.java
@@ -502,6 +502,8 @@ class LinuxWatchService
                     try {
                         this.wait();
                     } catch (InterruptedException e) {
+                        checkpointState = CheckpointRestoreState.CHECKPOINT_ERROR;
+                        throw e;
                     }
                 }
                 if (checkpointState == CheckpointRestoreState.CHECKPOINT_ERROR) {
@@ -522,6 +524,8 @@ class LinuxWatchService
                     try {
                         this.wait();
                     } catch (InterruptedException e) {
+                        checkpointState = CheckpointRestoreState.CHECKPOINT_ERROR;
+                        throw e;
                     }
                 }
                 if (checkpointState != CheckpointRestoreState.NORMAL_OPERATION) {
--- a/src/java.base/linux/classes/sun/nio/fs/LinuxWatchService.java
+++ b/src/java.base/linux/classes/sun/nio/fs/LinuxWatchService.java
@@ -28,6 +28,11 @@ package sun.nio.fs;
 import java.nio.file.*;
 import java.util.*;
 import java.io.IOException;
+
+import jdk.crac.CheckpointException;
+import jdk.crac.Context;
+import jdk.crac.Resource;
+import jdk.internal.crac.JDKResource;
 import jdk.internal.misc.Unsafe;
 
 import static sun.nio.fs.UnixNativeDispatcher.*;
@@ -51,30 +56,7 @@ class LinuxWatchService
     private final Poller poller;
 
     LinuxWatchService(UnixFileSystem fs) throws IOException {
-        // initialize inotify
-        int ifd = - 1;
-        try {
-            ifd = inotifyInit();
-        } catch (UnixException x) {
-            String msg = (x.errno() == EMFILE) ?
-                "User limit of inotify instances reached or too many open files" :
-                x.errorString();
-            throw new IOException(msg);
-        }
-
-        // configure inotify to be non-blocking
-        // create socketpair used in the close mechanism
-        int sp[] = new int[2];
-        try {
-            configureBlocking(ifd, false);
-            socketpair(sp);
-            configureBlocking(sp[0], false);
-        } catch (UnixException x) {
-            UnixNativeDispatcher.close(ifd);
-            throw new IOException(x.errorString());
-        }
-
-        this.poller = new Poller(fs, this, ifd, sp);
+        this.poller = new Poller(fs, this);
         this.poller.start();
     }
 
@@ -141,7 +123,7 @@ class LinuxWatchService
     /**
      * Background thread to read from inotify
      */
-    private static class Poller extends AbstractPoller {
+    private static class Poller extends AbstractPoller implements JDKResource {
         /**
          * struct inotify_event {
          *     int          wd;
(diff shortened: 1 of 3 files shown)


test/jdk/jdk/crac/fileDescriptors/FileWatcherAfterRestoreTest.java
```diff
@@ -0,0 +1,80 @@
...
+public class FileWatcherAfterRestoreTest implements CracTest {
+    @Override
+    public void test() throws Exception {
+        CracBuilder builder = new CracBuilder();
+        builder.doCheckpointAndRestore();
+    }
+
+    @Override
+    public void exec() throws Exception {
+        WatchService watchService = FileSystems.getDefault().newWatchService();
```
rvansa, line 48: Could you call `close()` on this service at the end of the test to check that it works as it should after C/R?

rvansa (commented):
Thanks for the contribution!

I wonder if you have an example of code where it's useful to automatically suspend the service but leave `WatchKey` management up to the user. To me this would be either fully transparent or fully managed, breaking the cycle that polls the events and closing the service as well.

joeyleeeeeee97:
@rvansa  Thanks for your comments, I working on my patch based on these recommendations.

I noticed a occasional crash in FileWatcherAfterRestoreTest when restore the Process from dump.

It crashes on where main thread is exiting, this is the full log. 
https://gist.github.com/joeyleeeeeee97/42711c35e02b0d2c530bad22aa2c5dd0

I could reproduce it locally with `hile ~/jtreg/jtreg/bin/jtreg -v  ./test/jdk/jdk/crac/fileDescriptors/FileWatcherAfterRestoreTest.java; do :; done` and `/home/ubuntu/jdk/build/linux-x86_64-server-release/images/jdk/bin/java  -XX:CRaCRestoreFrom=cr`.

Could I get some inputs on where my updates might caused the crash? Or could it be a separate issue?

rvansa:
Can't tell out of top of my head - the crash.log has the VM error log printed just after calling `ls` rather than `java -ea -XX:CRaCCheckpointTo=cr jdk.test.lib.crac.CracTest __run_test__ FileWatcherAfterRestoreTest` - have you edited it?

Could you try to reproduce with a `CONF=fastdebug` build?

joeyleeeeeee97:
@rvansa , sorry, looks like I accidentally paste some of the bash histories in gist, the error is caused by `java -ea -XX:CRaCCheckpointTo=cr jdk.test.lib.crac.CracTest __run_test__ FileWatcherAfterRestoreTest` .

And here is the `fastdebug` level [crash log ](https://gist.github.com/joeyleeeeeee97/42711c35e02b0d2c530bad22aa2c5dd0).

And I am not editing any related logic, this happens after the restored process finished execution and start exiting.
```
V  [libjvm.so+0x8d4275]  oop RawAccessBarrier<594020ul>::oop_load<oop>(void*)+0x15
V  [libjvm.so+0x8d4905]  AccessInternal::PostRuntimeDispatch<G1BarrierSet::AccessBarrier<594020ul, G1BarrierSet>, (AccessInternal::BarrierType)2, 594020ul>::oop_access_barrier(void*)+0x15
V  [libjvm.so+0x137d20b]  ObjectMonitor::object_peek() const+0x1b
V  [libjvm.so+0x1798354]  ObjectSynchronizer::release_monitors_owned_by_thread(JavaThread*)+0xf4
V  [libjvm.so+0x180f447]  JavaThread::exit(bool, JavaThread::ExitType)+0x8d7
V  [libjvm.so+0xe91a4d]  jni_DetachCurrentThread+0x8d
C  [libjli.so+0x4c71]  JavaMain+0xc41
C  [libjli.so+0x7a19]  ThreadJavaMain+0x9
```

rvansa:
I was hoping that a fastdebug build with assertions on would fail somewhere earlier but it is not the case. I would suspect that since this NULL dereference is happening when the thread already exits there's an earlier event that goes wrong.

joeyleeeeeee97:
@rvansa  Yes, it seems a obj monitor in in use list is wrong. This is my [gdb debug history](https://gist.github.com/joeyleeeeeee97/96213d4dfd655894d86f7e36531f03f2)

test/jdk/jdk/crac/fileDescriptors/FileWatcherAfterRestoreTest.java
joeyleeeeeee97: ok

src/java.base/linux/classes/sun/nio/fs/LinuxWatchService.java
```diff
@@ -431,6 +489,49 @@ private void processEvent(int wd, int mask, final UnixPath name) {
...
+                }
+            }
+        }
+
+        @Override
+        public void afterRestore(Context<? extends Resource> context) throws Exception {
+            if (!watcher.isOpen()) {
+                return;
+            }
+            synchronized (checkpointLock) {
```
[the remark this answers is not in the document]
joeyleeeeeee97: I added a check at after restore to check if initFD succeed.
```
                if (checkpointState != CheckpointRestoreState.NORMAL_OPERATION) {
                    throw new CheckpointException("LinuxWatchService restore exception");
                }
```

test/jdk/jdk/crac/fileDescriptors/FileWatcherTest.java
```diff
@@ -0,0 +1,48 @@
...
+import jdk.crac.Core;
+import jdk.test.lib.crac.CracBuilder;
+
+/**
+ * @test
+ * @library /test/lib
+ * @build FileWatcherTest
+ * @run driver jdk.test.lib.crac.CracTest
+ */
+public class FileWatcherTest implements CracTest {
```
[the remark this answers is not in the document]
joeyleeeeeee97: removed duplicate test

src/java.base/linux/classes/sun/nio/fs/LinuxWatchService.java
```diff
@@ -311,7 +363,13 @@ public void run() {
                     int nReady, bytesRead;
 
                     // wait for close or inotify event
-                    nReady = poll(ifd, socketpair[0]);
+                    try {
+                        do {
+                            nReady = poll(ifd, socketpair[0]);
```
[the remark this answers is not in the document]
rvansa: Let's explain the design first, please, then we can think about the lost wakeup. I wrote the comment actually before I realized that watch keys are not handled.

I understand the hesitation to manage watch keys automatically. What I am missing, though, is an example of idiomatic code where the user actually manages the watch keys during C/R - you only provide a test that checks that C/R fails when the key is left open. Maybe  There I would like to demonstrate that closing the notify service automatically simplifies the code, rather than just adding a close/reopen to the watch key management.

rvansa (commented):
Thanks for fixing the nitpicks; I'll wait for the test that demonstrates code managing WatchKey with this concurrent checkpoint.

joeyleeeeeee97:
Hi @rvansa , thanks for your review. I have updated FileWatcherAfterRestoreTest with a minimal case where the monitor key could be different after restore, and this test will manually  re-register alternative path.


```
When closing the notify service automatically simplifies the code, rather than just adding a close/reopen to the watch key management.
```
Where this patch could help simplifies code:
1. When users are manually managing all keys, this patch could only save them from  a `close()` and `open()` call before and  after checkpoint.
2. But some cases close and reopen a watch service could be very troublesome, like this case:
Notice the `watchservice` could be final for some 3rd party libraries, in that case we can not reopen and assign a new value for watch service. This patch allows user to focus only on keys, which might change after restore.
```java

import jdk.crac.Core;

import java.io.IOException;
import java.nio.file.FileSystems;
import java.nio.file.WatchService;

public class Case {
    private final WatchService watcher;

    Case() {
        watcher = createWatcher();
    }

    private static WatchService createWatcher() {
        try {
            return FileSystems.getDefault().newWatchService();
        } catch (IOException e) {
            return null;
        }
    }
}



```

test/jdk/jdk/crac/fileDescriptors/FileWatcherAfterRestoreTest.java
```diff
@@ -0,0 +1,104 @@
...
+        Path directory;
+        WatchKey key;
+
+        directory = Paths.get(System.getProperty("user.dir"), "workdir");
+        directory.toFile().mkdir();
+        Files.createTempFile(directory, "temp", ".txt");
```
rvansa, line 62: You're creating the file inside `isMatchFound`, too.

src/java.base/linux/classes/sun/nio/fs/LinuxWatchService.java
```diff
@@ -431,6 +488,52 @@ private void processEvent(int wd, int mask, final UnixPath name) {
...
+            if (!watcher.isOpen()) {
+                return;
+            }
+            synchronized (this) {
+                checkpointState = CheckpointRestoreState.RESTORE_TRANSITION;
+                this.notifyAll();
+                while (checkpointState == CheckpointRestoreState.RESTORE_TRANSITION) {
+                    try {
+                        this.wait();
+                    } catch (InterruptedException e) {
```
rvansa, line 524: Same here.

rvansa (commented):
Alright, the test does not really show what I was asking for (an independent thread using the watch service and **reacting** to checkpoint triggered elsewhere, e.g. through `jcmd <pid> JDK.checkpoint` but your comment explains that this change *might* be useful if the watch service is provided as static from a library beyond our control.

Therefore I think that once the technicalities are resolved this could be integrated @AntonKozlov .

rvansa (approved):
LGTM. Sorry for taking a while, I missed your last update. Please type `/integrate` so that @AntonKozlov can `/sponsor` and get this merged if he's OK with these changes (I am not a committer myself yet).

AntonKozlov (approved):
LGTM, thank you!

src/java.base/linux/classes/sun/nio/fs/LinuxWatchService.java
```diff
@@ -171,25 +153,94 @@ private static class Poller extends AbstractPoller {
...
+
+        private boolean processCheckpointRestore() throws IOException {
+            if (checkpointState != CheckpointRestoreState.CHECKPOINT_TRANSITION) {
+                return false;
+            }
+
+            synchronized (this) {
+                CheckpointRestoreState thisState;
+                if (wdToKey.size() == 0) {
+                    unsafe.freeMemory(address);
```
AntonKozlov, line 219: Although wait, where is the address is reinitialized? That is created in the constructor, and initFDs() reinitializes only file descriptors. This is very likely to be the reason behind those crashes, I can reproduce one locally.

I don't see any reasons to free the memory, it can perfectly be saved to the image. After removing this line, the test has passed 10 of 10 local runs.

AntonKozlov (changes requested):

rvansa:
Hi @joeyleeeeeee97 , Anton found an issue with memory allocation - could you address/react on that one?

PR #73, opened 2023-05-16
Active cleanup by CleanerImpl on checkpoint

rvansa:
Rather than registering PhantomCleanableRef as a resource that gets cleaned up in the C/R thread force cleanup in the CleanerImpl as this already keeps a list of all eligible references.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/73/head:pull/73` \
`$ git checkout pull/73`

`$ git checkout pull/73` \
`$ git pull https://git.openjdk.org/crac.git pull/73/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 73`

`$ git pr show -t 73`

</details>

<a href="https://git.openjdk.org/crac/pull/73.diff">https://git.openjdk.org/crac/pull/73.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/73#issuecomment-1549526207)


rvansa:
This is an alternative to https://github.com/openjdk/crac/pull/70

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
```diff
@@ -152,6 +169,33 @@ public void run() {
...
 
+    @Override
+    public synchronized void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+        // We block the cleaner thread to prevent race conditions between this
+        // thread and checkpointing thread invoking clean().
+        // When the cleanup starts in cleaner thread the checkpoint will skip
+        // it, but without waiting for the cleanup to finish (which might be
+        // critical for the checkpoint, e.g. closing FDs).
+        // The limitation is that code performing C/R must not wait on any task
+        // completed by the cleaner.
```
AntonKozlov, line 180: Could you elaborate why there is this limitation?

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
```diff
@@ -152,6 +169,33 @@ public void run() {
...
+        blockForCheckpoint = true;
+        thread.interrupt();
```
AntonKozlov, line 182: Why it has to be interrupt and not notify(), for example?

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
```diff
@@ -138,6 +141,20 @@ public void run() {
...
                 mlThread.eraseThreadLocals();
             }
+            if (blockForCheckpoint) {
+                try {
+                    synchronized (this) {
+                        waitingForCheckpoint = true;
+                        notify();
+                        while (blockForCheckpoint) {
+                            wait();
+                        }
```
AntonKozlov, line 151: Once we've got here, is it possible to ensure Cleaners has been called, and drop separate registration of Cleaners?

A concurrent cleaner registration is not a problem, as that depends on GC which is not predictable. I.e. if that happens, a slight race may also cause the cleaner to run after restore.

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
rvansa: The interrupt wakes up cleaner in `queue.remove()` (line 161) in case it's blocking for next task.

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
rvansa: The thread is blocked until `afterRestore`. PhantomCleanableRefs are cleaned independently within its own priority class (after the thread is stopped), but any other cleanup tasks simply won't happen. Therefore if C/R after this point waits for the cleanup, it will deadlock.

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
rvansa: I don't follow. PhantomCleanableRefs are cleaned strictly after this point (these have later priority), other cleaners will be queued up by GC but not called until restore.

AntonKozlov: I mean, once we synchronized cleaner thread from picking up PhantomCleanableRefs, there is apparently no reason to go to the next priority and discover cleanups through registrations in the CRaC -- instead, all cleanups are discoverable thorugh the list maintained by the cleaner. It seems we can go through the list right from this Resource and run cleanups.

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
AntonKozlov: OK, thanks. But what if the interrupt arrives to the thread during a cleaning while that runs, it may interrupt the cleaning, right?

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
rvansa: Alright, you mean that PCRs won't be Resources anymore and we'll clean them up whether are these already in the `queue` or not, without waiting for the GC to move them into that.

AntonKozlov: Something like that, but indeed the interaction between RefQueue and the list maintained by the cleaner should reviewed in the context is it safe and correct to do that.

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
AntonKozlov: Hmm, this is a problem. ReferenceQueue may have a JDK-internal interface via SharedSecretes to interrupt the wait. Or the interface to remove entries without any wait.

rvansa:
@AntonKozlov Updated.

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
```diff
@@ -138,10 +143,30 @@ public void run() {
...
+                            next = next.cleanIfNull();
+                        } while (next != phantomCleanableList);
+                    }
+                    synchronized (this) {
+                        cleanupComplete = true;
+                        notify();
+                        while (forceCleanup) {
+                            wait();
+                        }
+                        cleanupComplete = false;
```
AntonKozlov, line 160: AFIAU this synchronizes with beforeCheckpoint to verfiy cleanupComplete==true is noticied. An alternative would be to drop this sync, releasing cleaner sooner, and then just reset cleanupComplete in beforeCheckpoint(). But this is not a big deal.

AntonKozlov (approved):
LGTM, thank you!

src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java
rvansa: This has a point, the waiting is not needed in here. I've moved the synchronization from previous version rather mindlessly.

AntonKozlov: In the current state, at initialization `forceCleanup == cleanupComplete == false`. But after synchronization has been started, `forceCleanup == !cleanupComplete`, always. It seems possible to drop cleanupComplete, right?

rvansa: Right. I also realized that I didn't handle thread termination well; had the thread exited just after setting `forceCleanup`, the checkpoint thread might end up waiting for a notify that never comes. I am not sure if this can happen in practice, because one of the refs in the list is the cleaner itself, but let's be on the safe side.

PR #74, opened 2023-05-17
Introduce per-Priority Context with different policies

AntonKozlov:
A follow-up work for #60:

* Each priority now has a dedicated context, so contextes may provide different policies. CALL_SITE now uses new CriticalUnorderedContext that runs beforeCheckpoint on concurrent registration, fixes [1]. Whether or not CALL_SITE needs to be registered to at all is an open question and out of scope of this PR.
* the Global Context reverted from BlockingOrderedContext to OrderedContext, as that may have a huge impact on users. Probably we'll want to expose blocking/criticalUnorderd context along the global one, or at some point expose an implementation. But this is also out of scope of the PR.
* hierachy of the Context implementations are cleaned up a bit [2]

The JDKContext is now just a holder of ClaimedFDs, I'll address this in a follow-up that depends on this Context follow-up.

[1] https://github.com/openjdk/crac/pull/60#issuecomment-1545588281
[2] https://github.com/openjdk/crac/pull/60#discussion_r1185510445


src/java.base/share/classes/jdk/crac/impl/CriticalUnorderedContext.java
```diff
@@ -0,0 +1,90 @@
...
+        restoreException.throwIfAny();
+    }
+
+    @Override
+    public void register(R resource) {
+        synchronized (this) {
+            resources.put(resource, null);
+            if (concurrentCheckpointException != null) {
+                try {
+                    invokeBeforeCheckpoint(resource);
```
rvansa, line 83: This context is blocking as well - since you hold the monitor on this object if the code tries to indirectly (via operation in another thread) register other resource, it will deadlock, and cannot be woken up via interrupt.
If the only usage is call sites we know that there won't be a recursive registration, so this not such a big flaw, though.

test/jdk/jdk/crac/ContextOrderTest.java
```diff
@@ -94,26 +97,26 @@ private static void testOrder() throws Exception {
...
-        System.gc();
     }
 
     private static void testRegisterBlocks() throws Exception {
         var recorder = new LinkedList<String>();
+
         // blocks register into the same OrderedContext
-        getGlobalContext().register(new CreatingResource<>(recorder, null, "regular",
-                getGlobalContext(), null));
+        Context<Resource> context = new BlockingOrderedContext<>();
```
rvansa, line 106: Why does this create a children context rather than using global one directly?

test/jdk/jdk/crac/ContextOrderTest.java
```diff
@@ -294,12 +283,18 @@ public String toString() {
         private final Context<R> childContext;
         private boolean first = true;
 
-        private CreatingResource(List<String> recorder, Priority priority, String id, Context<R> childContext, Priority childPriority) {
+        private CreatingResource(List<String> recorder, Priority priority, String id, Priority childPriority) {
             super(recorder, priority, id);
-            this.childContext = childContext;
+            this.childContext = (Context<R>) childPriority.getContext(); // XXX
```
rvansa, line 288: `// XXX`

test/jdk/jdk/crac/ContextOrderTest.java
```diff
@@ -51,21 +57,18 @@ public class ContextOrderTest {
...
+        m.invoke(null);
     }
 
     private static void testOrder() throws Exception {
         var recorder = new LinkedList<String>();
         getGlobalContext().register(new MockResource(recorder, null, "regular1"));
-        getJDKContext().register(new MockResource(recorder, NORMAL, "jdk-normal"));
-        getJDKContext().register(new MockResource(recorder, SECURE_RANDOM, "jdk-later"));
-        getGlobalContext().register(new CreatingResource<>(recorder, null, "regular2", getJDKContext(), NORMAL));
+        JDKResource resource2 = new MockResource(recorder, NORMAL, "jdk-normal");
```
rvansa, line 67: Looks like the local vars are not used?

test/jdk/jdk/crac/ContextOrderTest.java
```diff
@@ -253,31 +251,22 @@ private MockResource(List<String> recorder, Priority priority, String id) {
             this.recorder = recorder;
             this.priority = priority;
             this.id = id;
-        }
 
-        @Override
-        public Priority getPriority() {
-            return priority;
+            if (priority != null) {
+                priority.getContext().register(this);
```
rvansa, line 256: Registering to global context explicitly and to local one implicitly is confusing; if the mock does not require the `priority` field just remove the constructor arg and register it directly as well.

src/java.base/share/classes/jdk/crac/impl/BlockingOrderedContext.java
```diff
@@ -0,0 +1,60 @@
...
+
+    // This method has particularly verbose name to stick out in thread dumps
+    // when the registration leads to a deadlock.
+    private void waitWhileCheckpointIsInProgress(R resource) {
+        if (Thread.currentThread().isInterrupted()) {
+            LoggerContainer.debug(Thread.currentThread().getName() + " not waiting in " + this +
+                " to register " + resource + "; the thread has already been interrupted.");
+            // We won't cause IllegalStateException because this is not an unexpected state
+            // from the point of CRaC - it probably tried to register some code before.
+            throw new RuntimeException("Interrupted thread tried to block in registration of " + resource + " in " + this);
```
rvansa, line 20: The use of `this` in the exception relies on naming the context and the `toString()` method for easy identification. Since you've removed these it will show only class type rather than Global Context/custom name/JDK resource priority.

src/java.base/share/classes/jdk/internal/crac/JDKResource.java
```diff
@@ -29,72 +29,4 @@
 import jdk.crac.Resource;
 
 public interface JDKResource extends Resource {
```
rvansa, line 31: Could we drop the interface completely?

rvansa (changes requested):
My main objection is allowing the checkpoint to proceed even if a resource in the `CriticalUnorderedContext` throws an exception.

However, when I started the review I was hoping that you'll be able to remove the special priority for call sites and broken encapsulation in `MethodHandleNatives`. You've made this change a explicitly out of scope, but without this and no extra test that would demonstrate a previously flawed behavior it is unclear what is the actual benefit this PR brings.

src/java.base/share/classes/jdk/crac/impl/CriticalUnorderedContext.java
```diff
@@ -0,0 +1,90 @@
...
+        try {
+            super.beforeCheckpoint(context);
+        } catch (CheckpointException e) {
+            synchronized (this) {
+                concurrentCheckpointException.handle(e);
+            }
+        }
+        synchronized (this) {
+            ExceptionHolder<CheckpointException> e = concurrentCheckpointException;
+            concurrentCheckpointException = new ExceptionHolder<>(CheckpointException::new);
```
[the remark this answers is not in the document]
AntonKozlov: Exception holder is inteded to be a short-living object. Doing that is possible, but will also create a temptation to reuse-the object, that will complicate its lifecycle without a need.

src/java.base/share/classes/jdk/crac/impl/ExceptionHolder.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public void handle(Exception e) {
+        if (e == null) {
+            return;
+        }
+
+        E exception = get();
+        if (exception.getClass() == e.getClass()) {
+            for (Throwable t : e.getSuppressed()) {
```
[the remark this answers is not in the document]
AntonKozlov: To avoid this PR grow too much, I basically agree with the comments, but still propose an aggregate exception, replied to the comment.

src/java.base/share/classes/jdk/crac/impl/CriticalUnorderedContext.java
AntonKozlov: Right, this context implementation something that should suit JDK needs, and the point is to verify the user is capable to do something similar with existing API. So, indeed, there is a possiblity of the deadlock with other thread. Fixing this does not seem trivial, so let's assume JDK will have to do something about this, if the problem become possible.

test/jdk/jdk/crac/ContextOrderTest.java
AntonKozlov: Global Context was changed to non-blocking OrderedContext as stated in the PR description, but the test needs BlockingOrderedContext.

test/jdk/jdk/crac/ContextOrderTest.java
AntonKozlov: Indeed, I feel something wrong with the resulting types in the test, but not sure how to fix.

test/jdk/jdk/crac/ContextOrderTest.java
AntonKozlov: Overlooked Resources are creating strong refs in their constructors.

test/jdk/jdk/crac/ContextOrderTest.java
rvansa: Yes, but it does not explain much about the **reason** but "... as that may have a huge impact on users". I don't think there's much concern about backwards compatibility at this point; it's more important to have the best UX even in case that users do an error.

If you want to conserve certain behaviour, write a test that will validate what happens when you attempt to register into global context in one of those notifications.

AntonKozlov: Apparently you mean why the global context implementation has been changed, not why the test has to use an explicit implementation.

Honestly I did not realize the global context properties were changed in #60, so I just want to revert that for a while. This was rather big change. And that deserved a separate line in the javadoc, which is btw the specification, not the tests (although they are definetely useful). I lean toward using BlockingContext, but want to think a bit more about that, with the spec change.

src/java.base/share/classes/jdk/internal/crac/JDKResource.java
AntonKozlov: Probably, I propose to do this a bit later. E.g. it may be useful to override beforeCheckpoint here declaring no Exception is thrown.

test/jdk/jdk/crac/ContextOrderTest.java
rvansa: Yes, the behaviour should be specified through javadoc but that does not say anything about registration when the checkpoint is proceeded. Per your logic that behaviour is unspecified and hence the change doesn't matter. Moreover, you've explicitly asked to postpone javadoc changes into #65 which did not get review in 2 weeks.

Rather than changing implementation forth and back you could explain why certain behaviour is more useful. Now you've removed a test for certain behaviour - creating resource in global context - if you think it should behave in a different way you should keep it and assert different outcome.

AntonKozlov: Having that we've met so many self-deadlocks after the blocking was introduced, it seems it will happen with the users who will run into the same problem with the global context, which we suggest by default. I'm still trying to untagle with consequences of #60, with this PR in particular. You can see the policy for the global context is going to be specified by a single line of code. Anyway sorry for reviews taking so long. You can simplify reviewer job by separating functional and non-functional changes. Rephrasing javadoc for readability is non functional, changing the meaning is functional.

src/java.base/share/classes/jdk/crac/impl/CriticalUnorderedContext.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public void register(R resource) {
+        synchronized (this) {
+            resources.put(resource, null);
+            if (concurrentCheckpointException != null) {
+                try {
+                    invokeBeforeCheckpoint(resource);
+                } catch (Exception e) {
+                    concurrentCheckpointException.handle(e);
```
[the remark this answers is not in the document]
AntonKozlov: This makes sense. As a thought excercise, it is possible to have add a last Resource, that will query all pending exceptions in all previous contextes, or make this context block once all beforeCheckpoint are done and there is nowhere to report exceptions. Both are also possible for user contextes, outside of the system classes.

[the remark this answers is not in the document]
rvansa: If you're speaking about querying Contexts, that means extending its API for traversing and fetching those. Alternatively, should the contexts report exceptions to a well-known resource this resource would become a part of the API.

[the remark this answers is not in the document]
AntonKozlov: It is possible to add query method to JDK contextes without changing public context interface. The same way, the alternative won't require changes in the public API (it will be a convention between JDK contextexes and the well-known resource).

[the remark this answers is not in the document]
rvansa: Sure, if we're talking only about JDKContext and this is used nowhere except in JDKContext the options are wider, and your proposed solution would work.

test/jdk/jdk/crac/ContextOrderTest.java
AntonKozlov: It was burried along other things that PR was doing. After some thought [1] Resources are used for two different purposes: fix-up the state to ensure checkpoint won't fail later (close file descriptors, which otherwise will be detected later); and later we discovered some resources are ensuring properties of the image (e.g. cleaning secrets). The blocking behavior makes sense for the latter, and for real concurrent registration it uses the race to delay registration and thus the problem is resolved without user intervention. But it brings significant usability drawback with the deadlock with registration from the same thread when notification has been started. And this is IMO more frequent that a registration from another thread, which is waited for by the thread executing e.g. beforeCheckpoint. For the clean-up, blocking 

We should not impose such a big inconvenience for the more frequent use case, regardless how the error is detected or reported. Applications, in their turn, can implement a blocking context with necessary guarantees (or use some implementation, which we eventually expose, once we got this contextes shaked out).

My take on blocking is preffered to an exception was in context of JDK implementation: we do not expose the problem via API, but auto-deadlocks is a bug in the JDK implementaiton.

[1] https://github.com/openjdk/crac/pull/60#issuecomment-1545853147

rvansa: You won't know whether the resource registered in global context is to be invoked optionally or if it's mandatory; any checks like those for FDs is only a last line of defense and debugging help. Therefore not being executed before C/R is a problem the platform should detect, preventing C/R from happening, to be on the safe side. We should not choose the unsafe for default, because users likely won't find out that something fishy is happening without any 'inconvenience'. As you say, users can change the semantics in their own context once they know.

I agree that deadlock is not the best way to report this, if you want to choice different means for the global context I am okay with that. We've talked about `register` throwing might not be the best option as the exception could be swallowed, exception from `Core.checkpointRestore` along with stack trace of the registration would be preferable.

src/java.base/share/classes/jdk/crac/impl/OrderedContext.java
```diff
@@ -28,55 +28,43 @@
...
-    private final String name;
-    private boolean checkpointing = false;
-    protected long order = 0;
-    protected final WeakHashMap<R, Long> resources = new WeakHashMap<>();
+/**
+ * Context performing Checkpoint notification in reverse order of registration.
+ * Concurrent registration along notification is silently ignored.
+ * @param <R>
+ */
+public class OrderedContext<R extends Resource> extends AbstractContext<R> {
```
rvansa, line 38: This class should not be used anywhere directly, and is not exposed to users => should be abstract.

src/java.base/share/classes/jdk/internal/crac/JDKContext.java
```diff
@@ -40,23 +39,14 @@
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.WeakHashMap;
 
-public class JDKContext extends PriorityContext<JDKResource.Priority, JDKResource> {
+public class JDKContext implements JDKResource {
```
rvansa, line 46: Since this is not a context anymore I believe that in the current form the class would deserve a rename (`ClaimedFileDescriptors`?)

src/java.base/share/classes/jdk/internal/crac/Core.java
```diff
@@ -26,15 +26,45 @@
...
+import jdk.crac.impl.BlockingOrderedContext;
+
 public class Core {
-    private static JDKContext JDKContext;
+    private static JDKContext jdkContext = new JDKContext();
 
-    static {
-        JDKContext = new JDKContext();
-        jdk.crac.Core.getGlobalContext().register(JDKContext);
+    public static JDKContext getJDKContext() {
```
rvansa, line 35: Please rename the method along with `JDKContext` rename.

test/jdk/jdk/crac/ContextOrderTest.java
```diff
@@ -94,26 +95,26 @@ private static void testOrder() throws Exception {
...
         assertTrue(recorder.stream().anyMatch("jdk-create-child2-before"::equals));
         assertTrue(recorder.stream().anyMatch("regular2-child2-before"::equals));
-
-        rememberMe.clear();
-        System.gc();
     }
 
     private static void testRegisterBlocks() throws Exception {
         var recorder = new LinkedList<String>();
+        BlockingOrderedContext<Resource> blockingCtx = new BlockingOrderedContext();
```
rvansa, line 102: We don't need separate contexts now that global one is blocking.

test/jdk/jdk/crac/ContextOrderTest.java
```diff
@@ -146,6 +144,12 @@ private static void assertWaits(Runnable runnable, String cls, String method) th
                         // clears the flag without rethrowing it is a bug.
                         thread.interrupt();
                         thread.join(TimeUnit.NANOSECONDS.toMillis(deadline - System.nanoTime()));
+                        System.out.println(thread.getState() + " " + thread.isAlive());
```
rvansa, line 147: Looks like forgotten debug logging

src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -123,6 +121,7 @@ private static void checkpointRestore1(long jcmdStream) throws
         LoggerContainer.debug("Starting checkpoint at epoch:{0}", System.currentTimeMillis());
 
         try {
+            jdk.internal.crac.Core.getJDKContext().beforeCheckpoint(null);
```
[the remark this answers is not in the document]
AntonKozlov: This part was intentionally omitted. JDKContext does not throw.

> The JDKContext is now just a holder of ClaimedFDs, I'll address this in a follow-up that depends on this Context follow-up.

src/java.base/share/classes/jdk/crac/impl/OrderedContext.java
AntonKozlov: By it's own it's a valid Contex implementation.

src/java.base/share/classes/jdk/internal/crac/JDKContext.java
AntonKozlov: See other comment https://github.com/openjdk/crac/pull/74#discussion_r1211309286

test/jdk/jdk/crac/ContextOrderTest.java
AntonKozlov: The test checks blocking properties, while blocking behavior of the global context  is not specified.

src/java.base/share/classes/jdk/crac/impl/BlockingOrderedContext.java
AntonKozlov: The name complicates the interface and from the stack trace the context will be completely evident.

test/jdk/jdk/crac/ContextOrderTest.java
```diff
@@ -146,6 +144,11 @@ private static void assertWaits(Runnable runnable, String cls, String method) th
                         // clears the flag without rethrowing it is a bug.
                         thread.interrupt();
                         thread.join(TimeUnit.NANOSECONDS.toMillis(deadline - System.nanoTime()));
+                        if (thread.getState() == Thread.State.WAITING) {
```
rvansa, line 147: I think you wanted to remove printing the stack traces here, too.

rvansa (approved):
I don't agree with some of your conclusions and handling some deficiencies in a follow-ups, but I guess this is as far as we can get here. Therefore approved, let's move on.

I hope that https://github.com/openjdk/crac/pull/64 can be addressed soon to fix on losing stack traces and messages from CheckpointException.

test/jdk/jdk/crac/ContextOrderTest.java
AntonKozlov: Let's have it there. This is on the failure path, so this won't appear until the failure. And everytime the failure happens, it's  really required to know what is the stack trace of the thread that does not respond to the interruption.

AntonKozlov:
You get me right. Meanwhile I had change to use contextes with the blocking policies, and in the current form the auto-deadlock fires too often. I was assuming deadlock is fine, once it a rare edge case in the JDK internals. But as the Global Context implementation, blocking context with auto-deadlock possibility is not acceptable IMO. We unlikely want to offer such to implementation to users. So definetely, this is not the last change in the area :)

PR #78, opened 2023-05-30
Support C/R engines loaded as dynamic library

rvansa:
Existing C/R engines assume external process to perform the C/R; this loads the code from a dynamic library.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/78/head:pull/78` \
`$ git checkout pull/78`

`$ git checkout pull/78` \
`$ git pull https://git.openjdk.org/crac.git pull/78/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 78`

`$ git pr show -t 78`

</details>

<a href="https://git.openjdk.org/crac/pull/78.diff">https://git.openjdk.org/crac/pull/78.diff</a>

</details>


src/hotspot/share/runtime/crac.cpp
```diff
@@ -488,10 +488,16 @@ void crac::restore() {
...
+    if (is_dynamic_library(_crengine)) {
+      // This function should not return on success
+      call_crengine_library(false, CRaCRestoreFrom);
+      warning("cannot restore from dynamic library %s", _crengine);
+    } else {
+      _crengine_args[1] = "restore";
+      add_crengine_arg(CRaCRestoreFrom);
+      os::execv(_crengine, _crengine_args);
+      warning("cannot execute \"%s restore ...\" (%s)", _crengine, os::strerror(errno));
+    }
```
AntonKozlov, line 500: It looks like we can have a single interface with dynamic libs. The lib may always spawn a process if required.

And to simplify the implementation even more, we can open the lib earlier that the checkpoint.

rvansa: > It looks like we can have a single interface with dynamic libs. The lib may always spawn a process if required.

Are you suggesting that we should turn criuengine/pauseengine/simengine to dynamic libraries and use only that?

> And to simplify the implementation even more, we can open the lib earlier that the checkpoint.

In other words, open it and keep open = leak it?

rvansa: > From the implementation, it will be enough to dlopen the engine on the VM start.

OK, in fact it is probably better to fail early.

PR #82, opened 2023-06-15
Draft: Win/Mac build for CRaC

wkia:


src/hotspot/share/runtime/os.cpp
```diff
@@ -91,6 +100,11 @@ julong os::free_bytes = 0;          // # of bytes freed
 
 static size_t cur_malloc_words = 0;  // current size for MallocMaxTestWords
 
+// CRaC
+static const char* _crengine = NULL;
```
rvansa, line 104: Looks like you unintentionally reverted recent changes - search for `_crengine_args` and handling code.

rvansa (commented):
I would probably have to do a more careful function-by-function comparison to check if any other changes were accidentally reverted.

wkia:
@rvansa Hi,
You can ignore this PR for now. Currently it is for testing purposes.
I will remove "draft" mark when it is ready for review.

src/java.base/share/native/launcher/main.c
```diff
@@ -151,7 +152,7 @@ static void setup_sighandler() {
     sigact.sa_flags = SA_SIGINFO;
     sigact.sa_sigaction = sighandler;
 
-    for (int sig = 1; sig < __SIGRTMIN; ++sig) {
+    for (int sig = 1; sig <= 31; ++sig) {
```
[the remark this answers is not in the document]
wkia: Yes, for different platforms there are no named constants with the same meaning. We could create one especially for CRaC.

src/hotspot/share/runtime/os.cpp
wkia: Thanks, I will check this

wkia:
@jan @rvansa 
Hi,
Could you review this, please?

src/hotspot/share/services/diagnosticCommand.cpp
```diff
@@ -1045,16 +1047,15 @@ void CheckpointDCmd::execute(DCmdSource source, TRAPS) {
...
                          vmSymbols::checkpointRestoreInternal_name(),
                          vmSymbols::checkpointRestoreInternal_signature(), &args, CHECK);
   oop str = result.get_oop();
   if (str != NULL) {
     char* out = java_lang_String::as_utf8_string(str);
     if (out[0] != '\0') {
-      current_op = LinuxAttachListener::get_current_op();
-      outputStream* stream = current_op->is_effectively_completed() ? tty : output();
+      outputStream* stream = output();
+      LINUX_ONLY(if (LinuxAttachListener::get_current_op() && LinuxAttachListener::get_current_op()->is_effectively_completed()) { stream = tty; })
```
AntonKozlov, line 1058: A multiline comment would look better
```
#ifdef LINUX
...
```

The original code did not check LinuxAttachListener::get_current_op() == NULL, as at this point the operation should exist. I propose omit the check or replace that with assert() at most.

src/java.base/windows/native/pauseengine/pauseengine.c
AntonKozlov, line 1: Are there strong reasons to duplicate the complete file? And not to introduce ifdef's to the common code? A lot of logic was duplicated with this.

wkia: Done

AntonKozlov (approved):
LGTM, thank you!

PR #85, opened 2023-06-19
Wake up all TIMED_WAITING threads after restore

rvansa:
This is a fix for an issue found by @jankratochvil when testing #53: Threads that enter sleep or timed parking use absolute monotonic time for pthread_cond_timedwait(). When the monotonic time changes during C/R we need to wake all threads to readjust the timeout to the new absolute time.

This introduces effectively a spurious wakeup; this is permitted for all the uses of pthread_cond_timedwait. Implementation either handles that transparently or propagates the wakeup to Java.

This commit does not handle timed waiting in non-Java threads other than WatcherThread.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/85/head:pull/85` \
`$ git checkout pull/85`

`$ git checkout pull/85` \
`$ git pull https://git.openjdk.org/crac.git pull/85/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 85`

`$ git pr show -t 85`

</details>

<a href="https://git.openjdk.org/crac/pull/85.diff">https://git.openjdk.org/crac/pull/85.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/85#issuecomment-1597328533)


src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6116,6 +6116,26 @@ bool VM_Crac::is_claimed_fd(int fd) {
   return false;
 }
 
+static void wakeup_threads_in_timedwait() {
+  SafeThreadsListPtr listPtr(Thread::current(), true);
+  ThreadsList *list = listPtr.list();
+  assert(list != NULL, "Thread list is null");
+  for (uint i = 0; i < list->length(); ++i) {
+    JavaThread* t = list->thread_at(i);
```
AntonKozlov, line 6124: Please use Threads::threads_do, that includes non-Java threads as well.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6116,6 +6116,26 @@ bool VM_Crac::is_claimed_fd(int fd) {
...
+    assert(t != NULL, "Thread is null");
+    ThreadState state = t->osthread()->get_state();
+    // ThreadState::SLEEPING is not used
+    if (state == ThreadState::CONDVAR_WAIT || state == ThreadState::OBJECT_WAIT) {
+      // We want to cause a wakeup but not interrupted exception
+      t->interrupt();
+      t->osthread()->set_interrupted(false);
+    }
```
AntonKozlov, line 6132: I think the problem of the native implementation of the timed wait is being fixed on the wrong level of abstraction. The interrupt() is much higher leve operation than the pthread_cond_signal, that we are actually targeting. And the latter would be also much cheaper. Could you replace the code with signaling every cond?

AntonKozlov:
> Threads that enter sleep or timed parking use absolute monotonic time for pthread_cond_wait(). 

Probably pthread_cond_timedwait() ? Should not be other *_timedwait be handled, like sem_timedwait()?

> When the monotonic time changes during C/R we need to wake all threads to readjust the timeout to the new absolute time.

I assume this happens in the implementation of pthread_cond_timedwait and not the JVM caller function? Otherwise I can't see the code doing the recalculation.

> This introduces effectively a spurious wakeup; this is permitted for all the uses of pthread_timed_wait.
> Implementation either handles that transparently or propagates the wakeup to Java.

OS-level spurious wake-ups are fine, but propogating that to java level seems dangerous. I could not spot the example of the propogation. E.g. can that happen to Thread.sleep()?

src/hotspot/os/linux/os_linux.cpp
rvansa: We are not triggering the Java InterruptedException; when you look into the implementation of `JavaThread::interrupt` you'll see that unparking from those 3 types of conditions is all it's actually doing, setting and immediately unsetting the interrupted flag is a noop on posix systems.
So I don't think this can be any cheaper. We are operating on a bit higher level only to reuse code that we would otherwise inline in here.

src/hotspot/os/linux/os_linux.cpp
rvansa: Good point, I'll use `Threads::java_threads_do`. Not sure what I could do with non-Java threads, I haven't found any use of `pthread_cond_timedwait` from non-Java threads except the WatcherThread (and we can't fix non-JVM threads, we need to track any condition to signal).

rvansa:
> Probably pthread_cond_timedwait() ? Should not be other *_timedwait be handled, like sem_timedwait()?

Right, sorry for those typos. `sem_timedwait` is fine since it uses wall-clock time, while we're concerned about monotonic time shifts.

> I assume this happens in the implementation of pthread_cond_timedwait and not the JVM caller function? Otherwise I can't see the code doing the recalculation.

No, I remember that happened somewhere higher up in JVM (I have to insert some debug logs to see that), will check it out again.

rvansa:
> OS-level spurious wake-ups are fine, but propogating that to java level seems dangerous. I could not spot the example of the propogation. E.g. can that happen to Thread.sleep()?

I've added checks for the actual duration to the test (on my machine the C/R takes about 800 ms, but I've tried with a longer wait time to give it a chance to fail) and the methods that are supposed to guarantee waiting (Thread.sleep/join, tryLock...) work correctly. `Object.wait` and `LockSupport.parkUntil` return spuriously after restore, but spurious return is explicitly allowed in javadoc. `Condition.await` is allowed to return but the implementation tracks the deadline internally and does not return before that.

src/hotspot/os/linux/os_linux.cpp
AntonKozlov: t->interrupt() triggers unparks of two ParkEvents and of a Parker. They do some bookkeeping in addition to just pthread_cond_timedwait(). But OK, I see we need to actually unpark the thread for park() callers to recalculate the abs time. Anyway, Thread::interrupt looks like a hack. AFAICS unpark()s can be called without Java thread state check, which also will be more straightforward.

rvansa: The thread state check might be more of an optimization, even though unparking a non-parked thread is a noop, it still means some work that's unnecessary.

AntonKozlov: https://github.com/openjdk/crac/blob/crac/src/hotspot/share/runtime/osThread.hpp#L39:
```
// The thread states represented by the ThreadState values are platform-specific
// and are likely to be only approximate, because most OSes don't give you access
// to precise thread state information.

// Note: the ThreadState is legacy code and is not correctly implemented.
// Uses of ThreadState need to be replaced by the state in the JavaThread.
```

Relying on the ThreadState creates a risk of missing unparking, which does not pay off, since the operation is perfromed once on the restore. It's possible that ThreadState will become more broken over time as the rest of Hotspot implementation evolve.

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -6116,6 +6116,30 @@ bool VM_Crac::is_claimed_fd(int fd) {
...
+class WakeupClosure: public ThreadClosure {
+  void do_thread(Thread* thread) {
+    ThreadState state = thread->osthread()->get_state();
+    // Note: ThreadState::SLEEPING is not used
+    if (state != ThreadState::CONDVAR_WAIT && state != ThreadState::OBJECT_WAIT) {
+      return; // No need to unpark running threads
+    }
+
+    assert(thread->is_Java_thread(), "must be called from java_threads_do");
+    JavaThread *jt = (JavaThread *) thread;
```
AntonKozlov, line 6128: A nit, just to mention:

```suggestion
    JavaThread* jt = thread->as_Java_thread(); 
```

rvansa:
@AntonKozlov Updated, removing the OS thread state and using `as_Java_thread()`

AntonKozlov (approved):
LGTM, thank you!

PR #86, opened 2023-06-20
PID adjustment on checkpoint

wkia:
On restore, there might be PID value conflicts because of small PID values, if it was checkpoint'ed in a container. Therefore, when checkpointing in a container, we need to move PID value for new processes to a particular value to avoid conflicts on restore.

See https://github.com/CRaC/example-lambda/blob/master/checkpoint.cmd.sh#L8 for example.

This PR contains implemented functionality similar to the example above, making this work out of the box. By default, if checkpointing, PID is adjusted only if Java's PID is 1 that means Java is run in a container. To adjust PID manually for a checkpoint'ed process, `-XX:CRaCMinPid=<value>` option should be used along with `CRaCCheckpointTo`. Min `CRaCMinPid` value is 1, max `CRaCMinPid` value is `UINT_MAX`, but it is actually limited by OS's pid_max.

There are the following possible scenarios for CRaC running in a container:
```
    // getpid   CRaCMinPid  |   set_last_pid      fork
    // ------------------------------------------------
    //   1         -        |    yes (default)    yes
    //   1         1        |    no               yes
    //   1        >1        |    yes              yes
    //   >1        -        |    no               no
    //   >1      <=getpid   |    no               no
    //   >1       getpid<   |    yes              yes
```

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/86/head:pull/86` \
`$ git checkout pull/86`

`$ git checkout pull/86` \
`$ git pull https://git.openjdk.org/crac.git pull/86/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 86`

`$ git pr show -t 86`

</details>

<a href="https://git.openjdk.org/crac/pull/86.diff">https://git.openjdk.org/crac/pull/86.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/86#issuecomment-1599253925)


src/java.base/share/native/launcher/main.c
```diff
@@ -167,6 +167,49 @@ static void setup_sighandler() {
...
+        last_pid = -1;
+    }
+    fclose(last_pid_file);
+    return last_pid;
+}
+
+static void set_last_pid(int pid) {
+    FILE *last_pid_file = fopen(last_pid_filename, "w");
+    if (!last_pid_file) {
+        perror("last_pid_file fopen");
```
rvansa, line 191: Should we have this message printed, even if we can achieve it through spinning? Looks like a unnecessary noise when we're still 'fine'.

src/java.base/share/native/launcher/main.c
```diff
@@ -278,15 +321,32 @@ main(int argc, char **argv)
...
-    if (is_checkpoint && 1 == getpid()) {
-        g_child_pid = fork();
-        if (0 < g_child_pid) {
-            // The main process should forward signals to the child.
-            setup_sighandler();
-            const int status = wait_for_children();
-            exit(status);
+    if (is_checkpoint) {
+        const int crac_min_pid_default = 128;
+        const char *env_min_pid_str = getenv("CRAC_MIN_PID");
```
rvansa, line 326: Why using an environment variable rather than JVM option? Env vars cannot be 'listed', so the user can't know about them without finding a specific place in documentation.

wkia: This is because of PID adjustment  happens before JVM in its, so to use JVM option we need to have JVM option parsing implemented in main.c (at least for one option), more complicated than it was done in parse_checkpoint(). That is why I've decided to use env var instead. OTOH maybe it worth considering moving PID adjustment & forking implementation to java.c, closer to JVM_Init call.

src/java.base/share/native/launcher/main.c
```diff
@@ -167,6 +167,49 @@ static void setup_sighandler() {
...
+        return;
+    }
+    if (0 > fprintf(last_pid_file, "%d", pid)) {
+        perror("last_pid_file fprintf");
+    }
+    fclose(last_pid_file);
+}
+
+static void spin_last_pid(int pid) {
+    for (pid_t child = fork(); child < (pid_t)pid; child = fork()) {
```
AntonKozlov, line 201: Although this part is not peformance critical, in my previous experiments, fork() was considerably more expensive compared to some minimalistic thread or vfork(). The best I found was `syscall(SYS_clone, SIGCHLD, NULL, NULL, 0)`. Could you check can that be made not too ugly? Maybe pthread_create?

src/java.base/share/native/launcher/main.c
```diff
@@ -167,6 +167,49 @@ static void setup_sighandler() {
...
+        if (0 == child) {
+            exit(0);
+        }
+        int status;
+        if (0 > waitpid(child, &status, 0)) {
+            perror("waitpid last pid");
+            break;
+        }
```
AntonKozlov, line 209: Can we drop this completely and rely on wait_for_children() below in the control flow?

src/java.base/share/native/launcher/main.c
```diff
@@ -278,15 +321,32 @@ main(int argc, char **argv)
...
+                set_last_pid(crac_min_pid);
+                if (get_last_pid() < crac_min_pid) {
+                    spin_last_pid(crac_min_pid);
+                }
+            }
+
+            // Avoid unexpected process completion when checkpointing under docker container run
+            // by creating the main process waiting for children before exit.
+            g_child_pid = fork();
+            if (0 < g_child_pid) {
```
AntonKozlov, line 344: Does it make sense to check if `g_child_pid < crac_min_pid`? As the ns_last_pid is a subject for races, and instead of guessing what the next pid will be, we'll have the ability to explicity check if the child has the right pid. The child will have to compare its pid with the expected crac_min_pid.

src/java.base/share/native/launcher/main.c
```diff
@@ -167,6 +167,49 @@ static void setup_sighandler() {
...
+    return last_pid;
+}
+
+static void set_last_pid(int pid) {
+    FILE *last_pid_file = fopen(last_pid_filename, "w");
+    if (!last_pid_file) {
+        perror("last_pid_file fopen");
+        return;
+    }
+    if (0 > fprintf(last_pid_file, "%d", pid)) {
```
AntonKozlov, line 194: Honestly I'd prefer write() to fprintf, to eliminate any possibility of buffering to write partial value to the kernel-exposed file.

src/java.base/share/native/launcher/main.c
```diff
@@ -278,15 +321,32 @@ main(int argc, char **argv)
...
-            const int status = wait_for_children();
-            exit(status);
+    if (is_checkpoint) {
+        const int crac_min_pid_default = 128;
+        const char *env_min_pid_str = getenv("CRAC_MIN_PID");
+        const int env_min_pid = env_min_pid_str ? atoi(env_min_pid_str) : 0;
+        // TODO: should it be checked for max pid overflow?
+        const int crac_min_pid = 0 < env_min_pid ? env_min_pid : crac_min_pid_default;
+
+        if (getpid() <= crac_min_pid) {
```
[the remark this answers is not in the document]
wkia: This should work for pid==1 as well. 
The possible values for crac_min_pid are 1, 2, 3, ...
In case of `getpid() < crac_min_pid`, it needs to have an additional check.

src/java.base/share/native/launcher/main.c
wkia: Sure, however, will it be OK if we create hundreds or thousands of child processes at the same time, even if they exist soon?

src/java.base/share/native/launcher/main.c
```diff
@@ -278,15 +321,32 @@ main(int argc, char **argv)
...
-        if (0 < g_child_pid) {
-            // The main process should forward signals to the child.
-            setup_sighandler();
-            const int status = wait_for_children();
-            exit(status);
+    if (is_checkpoint) {
+        const int crac_min_pid_default = 128;
+        const char *env_min_pid_str = getenv("CRAC_MIN_PID");
+        const int env_min_pid = env_min_pid_str ? atoi(env_min_pid_str) : 0;
+        // TODO: should it be checked for max pid overflow?
```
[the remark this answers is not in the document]
wkia: As far as I understand the max PID is not INT_MAX, it is 2^22. In case of CRAC_MIN_PID is greater than 2^22, we'd have a possible infinite loop while spinning the last PID.

src/java.base/share/native/launcher/main.c
```diff
@@ -278,15 +321,32 @@ main(int argc, char **argv)
...
+
+        if (getpid() <= crac_min_pid) {
+            // Move PID value for new processes to a desired value
+            // to avoid PID conflicts on restore.
+            if (get_last_pid() < crac_min_pid) {
+                set_last_pid(crac_min_pid);
+                if (get_last_pid() < crac_min_pid) {
+                    spin_last_pid(crac_min_pid);
+                }
+            }
```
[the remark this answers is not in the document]
wkia: The last PID may be greater than the current PID, so there might be the last PID satisfying CRaC requirements. If so, there is no need to set the last PID at all in such case.

src/java.base/share/native/launcher/main.c
wkia: In non-privileged containers, ns_last_pid cannot be even open for writing. So this message may be considered as a warning, or info, or suppressed at all as Radim suggested.

src/java.base/share/native/launcher/main.c
wkia: ok, I will check

src/java.base/share/native/launcher/main.c
wkia: I'm not sure I understand your thoughts. Could you explain what do you mean, please?

`if (0 < g_child_pid)` here is just a detector for child/parent code to execute, nothing more. At this line, there is no need for child to check its PID.

src/java.base/share/native/launcher/main.c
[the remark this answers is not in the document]
rvansa: I wouldn't test what's in `/proc/sys/kernel/pid_max` explicitly, but if PID #N+1 is lower than PID #N we have probably overflowed and that's what we should detect, and fail gracefully.

rvansa:
I wonder, given your other work, what will be the path to support other platforms? If that's going to be just `#ifdef LINUX` you could already guard platform-specific code with that.

src/java.base/share/native/launcher/main.c
AntonKozlov: Oh, of course. In case we cannot even open that, and that is expected, we should not print anything.

src/java.base/share/native/launcher/main.c
```diff
@@ -167,6 +167,49 @@ static void setup_sighandler() {
...
+}
+
+static void set_last_pid(int pid) {
+    FILE *last_pid_file = fopen(last_pid_filename, "w");
+    if (!last_pid_file) {
+        perror("last_pid_file fopen");
+        return;
+    }
+    if (0 > fprintf(last_pid_file, "%d", pid)) {
+        perror("last_pid_file fprintf");
```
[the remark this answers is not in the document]
AntonKozlov: I was wrong, the error is expected on open().

src/java.base/share/native/launcher/main.c
AntonKozlov: Performance-wise, it looks even better to collect them after java has started execution. But you're right, that will create a burst of short living processes. OK, probably this is a safer option.

src/java.base/share/native/launcher/main.c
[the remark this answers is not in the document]
AntonKozlov: Got it, in case user has provided some big value. For spin_last_pid, yes, it makes a lot of sense to build some safety, like Radim's suggestion. Or to limit number of attempts e.g. by the value itself (it should not take more than min_pid attempts to get pid>min_pid).

src/java.base/share/native/launcher/main.c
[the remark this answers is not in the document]
AntonKozlov: But to understand that, we need an extra read before write, and that read alone is of the same cost as the unconditional write.

src/java.base/share/native/launcher/main.c
AntonKozlov: I mean to check child pid (that is `g_check_pid), does it satisfy our min pid requirements. 

Right now, we have to check last_ns_pid and _assume_ what will be the child pid. Instead, we may create the child and based on the pid it gets, decide if the requirement is satifised, or we need another fork. Althought this will require one extra fork, if last_ns_pid is writeable.

This actually relates to https://github.com/openjdk/crac/pull/86#discussion_r1236956565.
OK, please disregard this.

src/java.base/share/native/launcher/main.c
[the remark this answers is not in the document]
wkia: Should Java fails if PID cannot be moved to a desired PID value? Or just warn and go on?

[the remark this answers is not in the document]
rvansa: If the value was explicitly set, I think it would be better to fail. When it's trying to get to PID 128 by default I think it is sufficient to warn the user **and** tell him that he could switch off the warning setting `-XX:CRMinPid=1`.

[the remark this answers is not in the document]
wkia: I did some experiments with PID spinning and a desired PID value that exceeds max_pid. It takes too long to spin PID until PID overflows. In case of a wrong value set by an user, this may seem like java hangs, so the user cannot wait so long to see the error message. This is also true for a valid desired PID value which is pretty big, e.g. 2_000_000.

We could remove `waitpid()` call to speed up PID spinning, but by removing this, we can easily reach container's resource limits (I tested this), so we cannot remove `waitpid` easily.

To avoid reading from `kernel/pid_max` and to avoid hanging on PID spinning, we could introduce max number of spin tries, say 10_000. If we reach this limit while spinning PIDs, we'd stop spinning and continue run Java with the currently reached PID. It actually seems doubtful that users want to move PID to 2M starting with PID=1 or 8 in a container. If users have some processes running in their container, on checkpoint they'd adjust desirable PID value in accordance with the state of the container, limited by a max try count we introducing. This solution seems portable for POSIX-like platforms.

Or, since things're becoming so complicated, it'd be easier to read max_pid, only for Linux though.

Please note that I'm talking about PID spinning, i.e. a case when writing to `ns_last_pid` haven't worked for some reasons.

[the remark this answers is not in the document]
rvansa: I agree that setting min pid to 2M is unlikely to be practical. Spinning fixed number of times is simpler, but if anything like this happens it would make sense to print a warning message (e.g. We are cycling PIDs due to -XX:CRMinPID=200000) after certain elapsed time (3 seconds?) rather than just number of iterations. Then the user can decide to cancel. Please make sure the stream gets flushed (in case this is written to a file or so...).

src/java.base/share/native/launcher/main.c
wkia: It seems like I've just met a case when `fprintf` returns ok (i.e. the correct number of bytes written), but the file content isn't changed. `strace`'ing shows that underlying `write` call returns `EPERM`, which probably wasn't forward to a caller.

src/java.base/share/native/launcher/main.c
wkia: Implemented -XX:CRaCMinPid option

src/java.base/share/native/launcher/main.c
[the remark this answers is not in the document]
wkia: Implemented PID spinning up to 10000 times. The 10000 is hardcoded now, we'd introduce an option if needed in future.

src/java.base/share/native/launcher/main.c
```diff
@@ -278,15 +343,33 @@ main(int argc, char **argv)
...
+        if (getpid() <= min_pid) {
+            // Move PID value for new processes to a desired value
+            // to avoid PID conflicts on restore.
+            {
+                const int res = set_last_pid(min_pid);
+                if (EPERM == res || EACCES == res || EROFS == res) {
+                    spin_last_pid(min_pid);
+                } else if (0 != res) {
+                    errno = res;
+                    perror("set_last_pid");
```
AntonKozlov, line 359: This apparently prints a textual version of the error. Better to avoid modifying the global state:

```suggestion
                    fprintf(stderr, "set_last_pid: %s\n", strerror(res))
```

src/java.base/share/native/launcher/main.c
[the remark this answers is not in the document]
wkia: As we discussed this offline, now PID adjustment is not applied by default, it works only if CracMinPid option is set by an user. So 'fork & wait' now is applied for pid==1 or for PID adjustment.

src/java.base/share/native/launcher/main.c
```diff
@@ -167,6 +184,56 @@ static void setup_sighandler() {
...
+    if (0 > len || sizeof(buf) < (size_t)len) {
+        return EINVAL;
+    }
+    const char *last_pid_filename = "/proc/sys/kernel/ns_last_pid";
+    const int last_pid_file = open(last_pid_filename, O_WRONLY|O_CREAT|O_TRUNC, 0666);
+    if (0 > last_pid_file) {
+        return errno;
+    }
+    int res = 0;
+    if (0 > write(last_pid_file, buf, len)) {
```
snazarkin, line 200: I'd compare with len, just to handle all  "write" return values

snazarkin (changes requested):

src/java.base/share/native/launcher/main.c
```diff
@@ -167,6 +184,56 @@ static void setup_sighandler() {
...
 
+static int set_last_pid(int pid) {
+#ifdef LINUX
+    char buf[11]; // enough for int32
+    const int len = snprintf(buf, sizeof(buf), "%d", pid);
+    if (0 > len || sizeof(buf) < (size_t)len) {
+        return EINVAL;
+    }
+    const char *last_pid_filename = "/proc/sys/kernel/ns_last_pid";
+    const int last_pid_file = open(last_pid_filename, O_WRONLY|O_CREAT|O_TRUNC, 0666);
```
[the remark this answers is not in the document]
wkia: Agreed.

src/java.base/share/native/launcher/main.c
```diff
@@ -106,6 +115,14 @@ static void parse_checkpoint(const char *arg) {
             is_checkpoint = 1;
         }
     }
+    if (!is_min_pid_set) {
+        const char *checkpoint_arg = "-XX:CRaCMinPid=";
+        const int len = strlen(checkpoint_arg);
+        if (0 == strncmp(arg, checkpoint_arg, len)) {
+            crac_min_pid = atoi(arg + len);
```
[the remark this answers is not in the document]
wkia: You're right, but in this implementation we don't care about errno. It's enough to get 0 as a parsed wrong value, because CRaCMinPid is introduced as JVM option with the min value =1, so in case of wrong value, JVM prints an error message and stops.

test/jdk/jdk/crac/ContainerPidAdjustmentTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @CracTestArg(4)
+    long expectedLastPid;
+
+    @CracTestArg(value = 5, optional = true)
+    String lastPidSetup;
+
+    final private String CURRENT_PID_MESSAGE = "Current PID = ";
+
+    @Override
+    // FIXME: need to add a test for default values, for Java's PID==1.
```
AntonKozlov, line 81: Is it still valid?

AntonKozlov (commented):
Code changes looks almost good, see the comment.

src/java.base/share/native/launcher/main.c
```diff
@@ -167,6 +185,56 @@ static void setup_sighandler() {
...
+    close(last_pid_file);
+    return res;
+#else
+    return EPERM;
+#endif
+}
+
+static void spin_last_pid(int pid) {
+    const int MaxSpinCount = pid < 1000 ? 1000 : pid;
+    for (int child = fork(), prev = 0, cnt = MaxSpinCount; child < pid; child = fork(), --cnt) {
```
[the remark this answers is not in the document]
AntonKozlov: Without wait, the child will occupy the PID number forever, without any need for that. So that's wrong. May be we can simplify the loop, so we won't leak the PID?

AntonKozlov (approved):
Thanks, LGTM!

src/java.base/share/native/launcher/main.c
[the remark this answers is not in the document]
wkia: Done

PR #87, opened 2023-06-26
Selectable Global Context implementation

AntonKozlov:
A number of problems were caused by switching to BlockingOrderedContext in the JDK code. To prevent the same in the potential users of EA builds, I temporarily switch the default implementation to the previous OrderedContext. The BlockingContext is still available via a command line option. After an EA build and transition of CRaC development to track newer version openjdk/jdk, the default implementation will be changed.

To make properties available, it was required to delay Reference resource registration. Otherwise, the GlobalContext implementation decision had to be done too early, in the Reference initialization during JDK bootstrapping, before Properties are available.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/87/head:pull/87` \
`$ git checkout pull/87`

`$ git checkout pull/87` \
`$ git pull https://git.openjdk.org/crac.git pull/87/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 87`

`$ git pr show -t 87`

</details>

<a href="https://git.openjdk.org/crac/pull/87.diff">https://git.openjdk.org/crac/pull/87.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/87#issuecomment-1607812672)

### DIFF
--- a/src/java.base/share/classes/jdk/crac/Core.java
+++ b/src/java.base/share/classes/jdk/crac/Core.java
@@ -88,8 +88,15 @@ public class Core {
 
         private static ReferenceHandlerResource resource = new ReferenceHandlerResource();
 
-        public static void register() {
+        static {
             jdk.internal.crac.Core.Priority.REFERENCE_HANDLER.getContext().register(resource);
+        }
+
+        /**
+         * Performs one-time registration of the Reference handling resource
+         */
+        public static void register() {
+            // nothing to do: the resource registered in the static initializer
         }
     }
 
--- a/src/java.base/share/classes/jdk/crac/Core.java
+++ b/src/java.base/share/classes/jdk/crac/Core.java
@@ -27,9 +27,12 @@
 package jdk.crac;
 
 import jdk.crac.impl.*;
+import jdk.internal.access.SharedSecrets;
 import jdk.internal.crac.ClaimedFDs;
+import jdk.internal.crac.JDKResource;
 import jdk.internal.crac.LoggerContainer;
 import sun.security.action.GetBooleanAction;
+import sun.security.action.GetPropertyAction;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
@@ -69,7 +72,26 @@ public class Core {
             GetBooleanAction.privilegedGetProperty("jdk.crac.trace-startup-time");
     }
 
-    private static final Context<Resource> globalContext = new BlockingOrderedContext<>();
+    private static final Context<Resource> globalContext = GlobalContext.createGlobalContextImpl();
+
+    private static class ReferenceHandlerResource implements JDKResource {
+        @Override
+        public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+            System.gc();
+            // TODO ensure GC done processing all References
+            while (SharedSecrets.getJavaLangRefAccess().waitForReferenceProcessing());
+        }
+
+        @Override
+        public void afterRestore(Context<? extends Resource> context) throws Exception {
+        }
+
+        private static ReferenceHandlerResource resource = new ReferenceHandlerResource();
+
+        public static void register() {
+            jdk.internal.crac.Core.Priority.REFERENCE_HANDLER.getContext().register(resource);
+        }
+    }
 
     /** This class is not instantiable. */
     private Core() {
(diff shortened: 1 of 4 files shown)


src/java.base/share/classes/jdk/crac/Core.java
```diff
@@ -69,7 +72,33 @@ private FlagsHolder() {}
             GetBooleanAction.privilegedGetProperty("jdk.crac.trace-startup-time");
     }
 
-    private static final Context<Resource> globalContext = new BlockingOrderedContext<>();
+    private static final Context<Resource> globalContext = GlobalContext.createGlobalContextImpl();
+
+    private static class ReferenceHandlerResource implements JDKResource {
```
rvansa, line 77: I wouldn't pollute the `Core` class with domain-specific code; if the 'reliability' of registration is of concern I think it's fine to expose the static `ensureRegistered`/`register` method.
Or, at least refactor this into its own file.

src/java.base/share/classes/jdk/crac/impl/GlobalContext.java
```diff
@@ -0,0 +1,19 @@
+package jdk.crac.impl;
+
+import jdk.crac.Context;
+import jdk.crac.Resource;
+import sun.security.action.GetPropertyAction;
+
+public class GlobalContext {
+    private static final String GLOBAL_CONTEXT_IMPL_PROP = "jdk.crac.globalContext.impl";
```
rvansa, line 8: As mentioned in off-GH conversation, I don't think this option would be ever used, and if you are going to change the default I'd rather drop it to keep the code simple.

AntonKozlov: Since there is no way to employ BlockingContext other than as the global context implementation, I prefer to make it selectable.

src/java.base/share/classes/jdk/crac/Core.java
AntonKozlov: I considered register() interface. The waitForReferenceProcessing is an internal JDK interface. The reference processing itself does not require the special managment on checkpoint, but it's checkpoint requirement as a procedure that no reference processing is in progress at some certain point.

The class is a boilerplate code to call two methods, it is not shared and unlikely will, so a separate file is not justified.

rvansa: If it's a requirement (and not an optimization), how come it's not prevented from running? (the blocking was there for a time before you reverted that, wasn't it?)

src/java.base/share/classes/jdk/crac/impl/GlobalContext.java
rvansa: OK, so in the future you intend to just change the default and provide non-blocking implementation as the workaround for cases where the registration is not synchronized properly? In that case it makes sense...

src/java.base/share/classes/jdk/crac/Core.java
AntonKozlov: This piece of code is reletated to this part of the description

> To make properties available, it was required to delay Reference resource registration. Otherwise, the GlobalContext implementation decision had to be done too early, in the Reference initialization during JDK bootstrapping, before Properties are available.

Without this code, the ReferenceResource was registered during Reference initialization. At that point, Properties were not initalized yet, so any attempt to call System.getProperty() threw NPE (regardless parameters).

I.e. if we have the default context implementation configuration via properties, we need to delay ReferenceResource registration.

src/java.base/share/classes/jdk/crac/impl/GlobalContext.java
AntonKozlov: Exactly. At some point we'll set on a single, the best implementation. But after the experience, I think we'll need the configuration for a while even after we change the default.

src/java.base/share/classes/jdk/crac/Core.java
rvansa: I understand moving the registration from Reference clinit, and if you'd rather keep the implementation here, let's keep it that way. I was referring to

> it's checkpoint requirement as a procedure that no reference processing is in progress at some certain point

- I am not sure what creates this requirement, and it is not 100% guaranteed the way it's implemented now.

rvansa (approved):

src/java.base/share/classes/jdk/crac/Core.java
AntonKozlov: The requirement is stated in the quoted part :) "no reference processing is in progress at some certain point" =(at REFERENCE_HANDLER priority).

It's required by the following CLEANERS (hard requirement, otherwise cleaner may miss something). And in general, reference processing should be quiescent -- to avoid checkpointing reference processing in the middle without a reason (soft requirement). I don't like how it's done at the moment. I've tried to make CLEANERS to ensure reference processing by its own, to be more straightforward. But in the absence of any CLEANER (e.g. the whole Cleaner package is not referenced and is not initialized), we won't satisfy the soft requirement, which still makes sense.

PR #89, opened 2023-06-30
Workaround JDK-8311164: CPU_HT is set randomly on hybrid CPUs like Alder Lake

jankratochvil:
SSIA

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/89/head:pull/89` \
`$ git checkout pull/89`

`$ git checkout pull/89` \
`$ git pull https://git.openjdk.org/crac.git pull/89/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 89`

`$ git pr show -t 89`

</details>

<a href="https://git.openjdk.org/crac/pull/89.diff">https://git.openjdk.org/crac/pull/89.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/89#issuecomment-1614457379)

### DIFF
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -2533,6 +2533,10 @@ void VM_Version::crac_restore() {
 
   uint64_t       features_missing =       features_saved & ~      _features;
   uint64_t glibc_features_missing = glibc_features_saved & ~_glibc_features;
+
+  // Workaround JDK-8311164: CPU_HT is set randomly on hybrid CPUs like Alder Lake.
+  features_missing &= ~CPU_HT;
+
   if (features_missing || glibc_features_missing) {
     static const char part1[] = "You have to specify -XX:CPUFeatures=";
     tty->print_raw(part1, sizeof(part1) - 1);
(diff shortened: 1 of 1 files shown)


src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -2588,6 +2592,10 @@ void VM_Version::initialize() {
   uint64_t   CPUFeatures_x64 = CPUFeatures_parse(GLIBCFeatures_x64);
   uint64_t       features_missing =   CPUFeatures_x64 & ~      _features;
   uint64_t glibc_features_missing = GLIBCFeatures_x64 & ~_glibc_features;
+
+  // Workaround JDK-8311164: CPU_HT is set randomly on hybrid CPUs like Alder Lake.
+  features_missing &= ~CPU_HT;
```
AntonKozlov, line 2597: Note: I just realized these missing features are different from the others, it's more "extra features" -- they are specified via the option, but not available on the current CPU.

AntonKozlov (approved):

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: I am not sure what you mean here. This part of the code checks that the `-XX:CPUFeatures=0x...,0x...` specifies only a subset of the current options, never superset. Or do you have a suspection there exist some other options besides `CPU_HT` which are not fatal to use even if the CPU does not support them?

rvansa (approved):
The testsuite passes on my Alder Lake (i7-12700H)

src/hotspot/cpu/x86/vm_version_x86.cpp
AntonKozlov: > This part of the code checks that the -XX:CPUFeatures=0x...,0x... specifies only a subset of the current options, never superset.

Good, that was my understanding as well. I mean just the variables names are the same, but the meaning is different. But I don't suppose it should be fixed. The PR is good.

Thanks!

PR #93, opened 2023-07-25
Ensure G1 GC is completed before creating a snapshot

jankratochvil:
@rvansa did report his snapshots are about 2x-3x bigger than they should be. He then also found it only happens if the snapshot is done too quickly after GC should have been run.

One can reproduce the race case by:
```
--- a/src/hotspot/share/gc/g1/g1UncommitRegionTask.hpp
+++ b/src/hotspot/share/gc/g1/g1UncommitRegionTask.hpp
@@ -35,7 +35,7 @@ class G1UncommitRegionTask : public G1ServiceTask {
   // is short, while still making reasonable progress.
   static const uint UncommitSizeLimit = 128 * M;
   // Initial delay in milliseconds after GC before the regions are uncommitted.
-  static const uint UncommitInitialDelayMs = 100;
+  static const uint UncommitInitialDelayMs = 10*1000;
   // The delay between two uncommit task executions.
   static const uint UncommitTaskDelayMs = 10;
 
```

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/93/head:pull/93` \
`$ git checkout pull/93`

`$ git checkout pull/93` \
`$ git pull https://git.openjdk.org/crac.git pull/93/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 93`

`$ git pr show -t 93`

</details>

<a href="https://git.openjdk.org/crac/pull/93.diff">https://git.openjdk.org/crac/pull/93.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/93#issuecomment-1649655798)


src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
```diff
@@ -129,7 +129,21 @@ void G1UncommitRegionTask::execute() {
...
   }
 }
+
+void G1UncommitRegionTask::_wait_if_active() {
+  while (_active) {
+    _active_sem.wait();
+  }
+}
+
+void G1UncommitRegionTask::wait_if_active() {
```
rvansa, line 144: I guess this gets thread-safe because `_instance` is set in VM thread, and we're calling it from another VM operation. However a comment, and preferably an assertion on which thread is executing this would be helpful in the future.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -405,6 +406,7 @@ Handle crac::checkpoint(jarray fd_arr, jobjectArray obj_arr, bool dry_run, jlong
   Universe::heap()->set_cleanup_unused(true);
   Universe::heap()->collect(GCCause::_full_gc_alot);
   Universe::heap()->set_cleanup_unused(false);
+  G1UncommitRegionTask::wait_if_active();
```
rvansa, line 409: Should we create a generic virtual method on heap (noop by default) to make this part of code GC-agnostic?

rvansa (commented):
Thanks, I'll test this later today.

src/hotspot/share/runtime/crac.cpp
jankratochvil: Yes, that makes sense, done, thanks.

src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
jankratochvil: I have added there some assertions. Not sure they should stay this way, do you have more insight about it?

rvansa: Oh, I looked wrong, we're *not* calling it from VM operation (`VM_Crac::doit`). Asserting on the name of the thread makes no sense, that can be changed arbitrarily, and the checkpoint operation can be triggered from any thread, including a non-application thread if we use the `jcmd <pid> JDK.checkpoint`.

I haven't checked all code paths but it seems that the `_instance` is set from `VM_G1CollectFull`. If we are to skip synchronization of access to `_instance` we could call the `wait_if_active` from our `VM_Crac` and use `assert(Thread::current()->is_VM_thread())` as the assertion. However when we're in another VM operation I guess the GC thread we would be waiting for isn't suspended in a safepoint (so we would deadlock).

jankratochvil:
I do not see what is the concern. I do not have much overview what all threads and VM threads are running in JVM. But there could be two concerns:
- Whether `_instance` could be set only soon after it was checked as NULL by `G1UncommitRegionTask::wait_if_active`. That would mean nobody called garbage collection before calling `G1UncommitRegionTask::wait_if_active`. That is not the case for CRaC:
```
  Universe::heap()->collect(GCCause::_full_gc_alot);
...
  Universe::heap()->wait_for_collection_finish();
```
If someone does it there is a race but such call would not make sense. And after all the worst what can happen is that one snapshots (or whatever the caller will do) more memory, that is inconvenient but not fatal.
- Whether `_active_sem.wait();` can wait forever. That could happen for example if `G1UncommitRegionTask::wait_if_active()` was called by G1CollectedHeap::heap()->service_thread(). T
Hat is asserted now. When else it could happen?

rvansa:
I had no concrete suspicion that some code path would cause trouble; I've simply raised a concern about non-synchronized access to a variable. Now I think that you went the extra mile to validate that this is safe, so I am OK with this change.

Please revert the `UncommitInitialDelayMs`.

src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
```diff
@@ -129,7 +130,24 @@ void G1UncommitRegionTask::execute() {
...
     report_summary();
     clear_summary();
   }
 }
+
+void G1UncommitRegionTask::_wait_if_active() {
+  assert(Thread::current_or_null() != NULL, "no current thread");
+  assert(!Thread::current()->is_ConcurrentGC_thread(), "deadlock prevention");
+
+  while (_active) {
```
snazarkin, line 143: Is this cycle safe against infinite loop? _active is not atomic or volatile.

jankratochvil: It is true I would normally make it `volatile` (it is atomic as it is small enough), it was forgotten as other code around is using it without `volatile`. But you are right this is a different use case of `_active` where cachine may have some effect.
OTOH for my defense the `volatile` should not be required due to a memory barrier there: https://stackoverflow.com/questions/4476446/can-a-c-c-compiler-legally-cache-a-variable-in-a-register-across-a-pthread-lib

src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
```diff
@@ -129,7 +130,24 @@ void G1UncommitRegionTask::execute() {
   } else {
     // Nothing more to do, change state and report a summary.
     set_active(false);
+    _active_sem.signal();
```
AntonKozlov, line 133: AFAICS the signal() increments the semaphore count. So the count is the number of executions of the uncommit task. That's, another wait() may succeed even when the opeartion is in progress (by decrementing the value accumulated during previous exections).

And I share concerns about access to the _active. For safe access there should be a critical section with acquire and release. 

Do we really need to synchronize with the G1UncommitRegionTask? Maybe, can we call `G1CollectedHeap::uncommit_regions` directly (that is being called by the UncommitTask eventually)?

jankratochvil: If there are multiple `signal()` calls stored then it will just run multiple times with no effect in the `while` cycle: https://github.com/openjdk/crac/pull/93/files#diff-b7c9d7a6a2eb233cfe201f2b07c4623fda47f1ac06174b66a95b1e46d9ec6450R143
I do not think `_active` is needed to be atomic there, that would be also already fixing existing OpenJDK code. But I will put it there, I agree it cannot harm. atomic is definitely cheaper than a mutex acquire+release protecting a single variable with the same safety so no mutex is needed.
You suggest during CRaC snapshot the whole VM is already frozen so we do not need any synchronization? OK, if it is so I can just call `uncommit_regions`.

AntonKozlov: > You suggest during CRaC snapshot the whole VM is already frozen so we do not need any synchronization? OK, if it is so I can just call `uncommit_regions`.

Please check if that is safe. I propose to call g1h->uncommit_regions() from the VM operation. But I'm not a G1/GC expert. AFAICS G1ServiceThread (G1ServiceThread::run_service) calls Task's execute() without sync, so we need at most synchornization implied by G1UncommitRegionTask::execute [1]. I missed this in the first reading: what does SuspendibleThreadSetJoiner do? That can be unnecessary if the uncommit is performed from the VM op. I don't see any synchronization beyond that (excluding some inside of uncommit_regions and its callees). 

It may be worth to discuss this on hotspot-gc-dev mail list [2].

[1] https://github.com/openjdk/crac/blob/crac/src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp#L105
[2] https://mail.openjdk.org/pipermail/hotspot-gc-dev/

jankratochvil: I do not know much how all the threads and VM threads and GC pauses work but I disagree with this patch. The patch before was waiting until the asynchronous `G1UncommitRegionTask` has done its job. Currently we do the job ourselves but that means `G1UncommitRegionTask` can be executed asynchronously while we do the same job in our thread. And I do not see any `synchronized` keywords in the `uncommit_regions` code this patch calls now.

jankratochvil: I read now I should contact https://mail.openjdk.org/pipermail/hotspot-gc-dev/ , thanks for the advice.

AntonKozlov: AFAICS HeapRegionManager::uncommit_inactive_regions is synchronized by Uncommit_lock.

https://github.com/openjdk/crac/blob/fbeaadb611f355b0616ebd25d86b9a7ced83276a/src/hotspot/share/gc/g1/heapRegionManager.cpp#L311

src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
jankratochvil: It has been changed/simplified upon @AntonKozlov's comments so I find this discussion no longer relevant.

AntonKozlov (commented):
Sorry, I've missed this.

LGTM! To be sure, have you tested the change in a debug build with asserts enabled?

jankratochvil:
> To be sure, have you tested the change in a debug build with asserts enabled?

I do run a simple Linux x86_64 snapshot+restore which does work. As the full regression test takes me 10 hours I just rely on GHA. But it is true I check GHA only very briefly for a massive failure. Few unexplainable failures are always there so I ignore them. In this case I see some crac x86 (32-bit) failures.

This situation is not good longterm. In the past GDB always had some failures in the testsuite and it was too much work to really fix them all so I was just making a diff between results. Should I/we spend the time to fix the failures or make some scripts for diffs of the results?

AntonKozlov:
> > To be sure, have you tested the change in a debug build with asserts enabled?
> 
> I do run a simple Linux x86_64 [fastdebug] snapshot+restore which does work. 

Great, thanks!

> As the full regression test takes me 10 hours I just rely on GHA. But it is true I check GHA only very briefly for a massive failure. Few unexplainable failures are always there so I ignore them. In this case I see some crac x86 (32-bit) failures.

I can confirm this PR does not regress GHA checks. So they don't block the integration, so I propose to proceed.

Among the failures, I'm concerned about jdk tier1 failures. We'd better to look at them.

AntonKozlov:
@jankratochvil Could you also actualize the PR title? Thanks!

jankratochvil:
> @jankratochvil Could you also actualize the PR title? Thanks!

True, done.

AntonKozlov (approved):
LGTM, thank you!

PR #97, opened 2023-07-31
Print better diagnostics for spawning CRIU

jankratochvil:
It was always a mystery why it does not work this time as there are no error messages printed.
Still may system call errors are currently not reported but those errors at least do not happen commonly.
In fact `os::exec_child_process_and_wait` should be more verbose but that is already a part of OpenJDK so I did not modify it.
It would be much easier if `criuengine.c` was in C++. I was considering to switch it but then I wrote it already in C.
I understand this code will become obsolete after planned integration of CRIU into JVM but it may not yet be soon enough.

### DIFF
--- a/src/java.base/unix/native/criuengine/criuengine.c
+++ b/src/java.base/unix/native/criuengine/criuengine.c
@@ -61,6 +61,75 @@ static int kickjvm(pid_t jvm, int code) {
     return 0;
 }
 
+static char *join_args(const char **args) {
+    size_t size = 0;
+    for (const char **argp = args; *argp; ++argp) {
+      size += strlen(*argp) * 4 + 2 + 1;
+    }
+    char *retval = malloc(size);
+    if (!retval) {
+        perror("malloc");
+        exit(1);
+    }
+    char *d = retval;
+    for (const char **argp = args; *argp; ++argp) {
+        if (d != retval) {
+            *d++ = ' ';
+        }
+        // https://unix.stackexchange.com/a/357932/296319
+        if (!strpbrk(*argp, " \t\n!\"#$&'()*,;<=>?[\\]^`{|}~")) {
+            d = stpcpy(d, *argp);
+            continue;
+        }
+        *d++ = '\'';
+        for (const char *s = *argp; *s; ++s) {
+          if (*s != '\'') {
+            *d++ = *s;
+          } else {
+            d = stpcpy(d, "'\\''");
+          }
+        }
+        *d++ = '\'';
+    }
+    *d = 0;
+    return retval;
+}
+
+static char *path_join(const char *path1, const char *path2) {
+    char *retval;
+    if (path2[0] == '/') {
+        retval = strdup(path2);
+        if (!retval) {
+            perror("strdup");
+            exit(1);
+        }
+        return retval;
+    }
+    if (asprintf(&retval, "%s/%s", path1, path2) == -1) {
+        perror("asprintf");
+        exit(1);
+    }
+    return retval;
+}
+
+static char *path_abs(const char *rel) {
+    char *cwd = get_current_dir_name();
+    if (!cwd) {
+        perror("get_current_dir_name");
+        exit(1);
+    }
+    char *retval = path_join(cwd, rel);
+    free(cwd);
+    return retval;
+}
+
+static char *path_abs2(const char *rel1, const char *rel2) {
+    char *rel1_abs = path_abs(rel1);
+    char *retval = path_join(rel1_abs, rel2);
+    free(rel1_abs);
+    return retval;
+}
+
 static int checkpoint(pid_t jvm,
         const char *basedir,
         const char *self,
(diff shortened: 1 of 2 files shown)


src/java.base/unix/native/criuengine/criuengine.c
```diff
@@ -61,6 +61,75 @@ static int kickjvm(pid_t jvm, int code) {
...
+    free(cwd);
+    return retval;
+}
+
+static char *path_abs2(const char *rel1, const char *rel2) {
+    char *rel1_abs = path_abs(rel1);
+    char *retval = path_join(rel1_abs, rel2);
+    free(rel1_abs);
+    return retval;
+}
```
AntonKozlov, line 131: This function is closer to path_from, rather than to abs. But by using just path_join/path_from, we'll provide a shorter yet still correct path to the log (the only use of the function), so I propose to drop it.

src/java.base/unix/native/criuengine/criuengine.c
```diff
@@ -99,50 +168,69 @@ static int checkpoint(pid_t jvm,
...
+            criu_cmdline = join_args(args);
+            criu_log = path_abs2(imagedir, log_local);
+            fprintf(stderr, "Spawned CRIU \"%s\" has not properly exited: exit code %d - check %s\n", criu_cmdline, WEXITSTATUS(status), criu_log);
+        }
         kickjvm(jvm, -1);
     } else if (leave_running) {
         kickjvm(jvm, 0);
     }
+    free(criu_cmdline);
+    free(criu_log);
```
AntonKozlov, line 233: Extensive use of free in the short living application is questionable. By dropping this we'll make it simpler and faster.

AntonKozlov (commented):
A few comments, but the approach looks good.

src/java.base/unix/native/criuengine/criuengine.c
```diff
@@ -61,6 +61,75 @@ static int kickjvm(pid_t jvm, int code) {
...
+        *d++ = '\'';
+    }
+    *d = 0;
+    return retval;
+}
+
+static char *path_join(const char *path1, const char *path2) {
+    char *retval;
+    if (path2[0] == '/') {
+        retval = strdup(path2);
```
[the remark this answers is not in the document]
jankratochvil: The function name corresponds to: https://docs.python.org/3/library/os.path.html#os.path.join

> If a segment is an absolute path, then all previous segments are ignored and joining continues from the absolute path segment.

The function first parameter is not always used for `cwd`.

src/java.base/unix/native/criuengine/criuengine.c
```diff
@@ -61,6 +61,75 @@ static int kickjvm(pid_t jvm, int code) {
...
+        return retval;
+    }
+    if (asprintf(&retval, "%s/%s", path1, path2) == -1) {
+        perror("asprintf");
+        exit(1);
+    }
+    return retval;
+}
+
+static char *path_abs(const char *rel) {
```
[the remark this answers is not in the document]
jankratochvil: Not really as `realpath()` is resolving symlinks while this function does not. I agree the code is ugly, I originally expected it will be more nice in C++ but then I realized [std::filesystem::path](https://en.cppreference.com/w/cpp/filesystem/path) is `C++17` only.

src/java.base/unix/native/criuengine/criuengine.c
jankratochvil: It became now simple enough in C++, is it OK?

src/java.base/unix/native/criuengine/criuengine.c
```diff
@@ -99,50 +168,69 @@ static int checkpoint(pid_t jvm,
...
+        exit(77);
     }
 
+    char *criu_cmdline = NULL;
+    char *criu_log = NULL;
     int status;
-    if (child != wait(&status) || !WIFEXITED(status) || WEXITSTATUS(status)) {
+    if (child != wait(&status)) {
+        criu_cmdline = join_args(args);
+        fprintf(stderr, "Error waiting for spawned CRIU \"%s\": %m\n", criu_cmdline);
```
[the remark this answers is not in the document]
jankratochvil: With C++ `std::string` it has been solved, hasn't it?

src/java.base/unix/native/criuengine/criuengine.c
```diff
@@ -99,50 +168,69 @@ static int checkpoint(pid_t jvm,
...
+        criu_cmdline = join_args(args);
+        criu_log = path_abs2(imagedir, log_local);
+        fprintf(stderr, "Spawned CRIU \"%s\" has not properly exited: waitpid status %d - check %s\n", criu_cmdline, status, criu_log);
+        kickjvm(jvm, -1);
+    } else if (WEXITSTATUS(status)) {
+        if (WEXITSTATUS(status) != 77) {
+            criu_cmdline = join_args(args);
+            criu_log = path_abs2(imagedir, log_local);
+            fprintf(stderr, "Spawned CRIU \"%s\" has not properly exited: exit code %d - check %s\n", criu_cmdline, WEXITSTATUS(status), criu_log);
+        }
```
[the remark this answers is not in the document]
jankratochvil: Either it got fixed by C++ or describe more the difference, please.

src/java.base/unix/native/criuengine/criuengine.cpp
```diff
@@ -63,6 +66,52 @@ static int kickjvm(pid_t jvm, int code) {
...
+static std::string path_abs(std::string rel) {
+    char *cwd_s = get_current_dir_name();
+    if (!cwd_s) {
+        perror("get_current_dir_name");
+        exit(1);
+    }
+    std::string cwd = cwd_s;
+    free(cwd_s);
+    return path_from(cwd, rel);
+}
```
AntonKozlov, line 109: The problem with this function as abstraction is that it first calculates CWD, then checks was the argument absolute or relative.

src/java.base/unix/native/criuengine/criuengine.cpp
```diff
@@ -63,6 +66,52 @@ static int kickjvm(pid_t jvm, int code) {
...
+        }
+        retval += '\'';
+        for (; *s; ++s) {
+            if (*s != '\'') {
+                retval += *s;
+            } else {
+                retval += "'\\''";
+            }
+        }
+        retval += '\'';
```
AntonKozlov, line 91: It will be surprised to meet these symbols in the args, and IMHO this is overly complicated way to report them

src/java.base/unix/native/criuengine/criuengine.cpp
```diff
@@ -131,12 +181,23 @@ static int checkpoint(pid_t jvm,
...
+        std::cerr << "Error waiting for spawned CRIU \"" << join_args(args) << "\": " << strerror(errno) << std::endl;
+        kickjvm(jvm, -1);
+    } else if (!WIFEXITED(status)) {
+        std::cerr << "Spawned CRIU \"" << join_args(args) << "\" has not properly exited:"
+          " waitpid status %d - check " << path_abs(imagedir, log_local) << std::endl;
+        kickjvm(jvm, -1);
+    } else if (WEXITSTATUS(status)) {
+        if (WEXITSTATUS(status) != SUPPRESS_ERROR_IN_PARENT) {
+            std::cerr << "Spawned CRIU \"" << join_args(args) << "\" has not properly exited:"
+              " exit code " << WEXITSTATUS(status) << " - check " << path_abs(imagedir, log_local) << std::endl;
```
AntonKozlov, line 199: we are not required to abs(imagedir), as unprocessed that is in the same way as it was provided in the args

src/java.base/unix/native/criuengine/criuengine.c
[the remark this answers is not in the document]
AntonKozlov: I was thinking of something like 
```
if (child != wait(&status)) {
 msg = "unexpected child pid"
} else if (!WIFEXITED(status)) {
 msg = "child killed"
} else if ((ecode = WEXITSTATUS(status))) {
 msg = asprintf("exit code: %d", ecode);
}
fprintf("CRIU failure: %s", msg)
for (int i = 0; i < n_args; ++i) {
  fprintf(" %s", args[i]);
```

That would not also require join_args abstrastion, but OK.

AntonKozlov (approved):
OK, looks fine although a bit overcomplicated as for me.

src/java.base/unix/native/criuengine/criuengine.cpp
jankratochvil: The code was more simple that way and it was straightforward to replace it with C++17 `std::filesystem::path` after [OpenJDK makes the switch to C++17](https://bugs.openjdk.org/browse/JDK-8310260).

src/java.base/unix/native/criuengine/criuengine.cpp
jankratochvil: Should it therefore abort when it sees such symbol? Otherwise the problem will not be reproducible from the error message.

src/java.base/unix/native/criuengine/criuengine.cpp
jankratochvil: The problem is users do not execute `criuengine` themselves. From my experience:
```
$ $JTREG_HOME/bin/jtreg -jdk:$ZULU_HOME -v:all -timeout:10 $TEST_HOME/test/jdk/jdk/crac/BigCheckpointTest.java
Spawned CRIU "/home/azul/azul/crac-git/build/linux-x86_64-server-fastdebug/images/jdk/lib/criu dump -t 648896 -D cr --shell-job -v4 -o dump4.log" has not properly exited: exit code 1 - check dump4.log
$ ls -l dump4.log
ls: cannot access 'dump4.log': No such file or directory
```
How to find out `dump4.log` is hiding in the `JTwork/scratch_1/cr` subdirectory?
```
$ ls -l  ./JTwork/scratch_1/cr/dump4.log
-rw------- 1 azul azul 3547 Aug 10 06:58 ./JTwork/scratch_1/cr/dump4.log
```

src/java.base/unix/native/criuengine/criuengine.cpp
AntonKozlov: The interface is fine, the implementation just could be more straightforward and efficient.

The link you're refering has Resolution: Rejected.

src/java.base/unix/native/criuengine/criuengine.cpp
AntonKozlov: You mean the line is not copy-pastable. But it is not such anyway, as it makes sense only for specific pid, and in the specific point in time (in VM_Crac safepoint).

I'm fine with the current code.

src/java.base/unix/native/criuengine/criuengine.cpp
AntonKozlov: >> we are not required to abs(imagedir), as unprocessed that is in the same way as it was provided in the args
> The problem is users do not execute criuengine themselves. From my experience:
> $JTREG_HOME/bin/jtreg -jdk:$ZULU_HOME -v:all -timeout:10 $TEST_HOME/test/jdk/jdk/crac/BigCheckpointTest.java
Spawned CRIU "/home/azul/azul/crac-git/build/linux-x86_64-server-fastdebug/images/jdk/lib/criu dump -t 648896 -D cr --shell-job -v4 -o dump4.log" has not properly exited: exit code 1 - check dump4.log

My point was about absolute path for image dir. I.e. this one looks fine

> $ java -XX:CRaCheckpointTo=./cr
> Spawned CRIU "/home/azul/azul/crac-git/build/linux-x86_64-server-fastdebug/images/jdk/lib/criu dump -t 648896 -D ./cr --shell-job -v4 -o dump4.log" has not properly exited: exit code 1 - check *./cr/dump4.log* 

But Jtreg case is valid, OK.

AntonKozlov (approved):

PR #100, opened 2023-08-09
Support -XX:+IgnoreCPUFeatures during -XX:CRaCRestoreFrom

jankratochvil:
---------
### Progress

### Reviewers
 * [Anton Kozlov](https://openjdk.org/census#akozlov) (@AntonKozlov - **Reviewer**) ⚠️ Review applies to [fe9af53e](https://git.openjdk.org/crac/pull/100/files/fe9af53e33265ddb8a2010065342e41a373c8251)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/100/head:pull/100` \
`$ git checkout pull/100`

`$ git checkout pull/100` \
`$ git pull https://git.openjdk.org/crac.git pull/100/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 100`

`$ git pr show -t 100`

</details>

<a href="https://git.openjdk.org/crac/pull/100.diff">https://git.openjdk.org/crac/pull/100.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/100#issuecomment-1671582838)


rvansa (commented):
The 'ignore' option is not documented in the option description. Also it's not clear what happens when you set the option with a different (mask-like) value on restore; IIUC this is ignored without any (warning) message.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -2544,7 +2555,22 @@ void VM_Version::crac_restore() {
...
+      ignore_glibc_not_using = true;
+    } else {
+      static const char part1[] = "Unsupported -XX:CPUFeatures=";
+      tty->print_raw(part1, sizeof(part1) - 1);
+      for (const char *s = CPUFeatures; *s; ++s)
+        tty->put(*s);
+      static const char part2[] = ", only -XX:CPUFeatures=ignore is supported during -XX:CRaCRestoreFrom";
+      tty->print_raw(part2, sizeof(part2) - 1);
+      tty->cr();
+      vm_exit_during_initialization();
```
rvansa, line 2569: Will this work when you set `CPUFeatures=generic` before checkpoint and don't set anything during restore? The option will keep its value and this might trigger the error.
If you don't want to add some extra tests you might piggyback in `VMOptionsTest`.

src/hotspot/cpu/x86/globals_x86.hpp
```diff
@@ -222,10 +222,11 @@ define_pd_global(intx, InitArrayShortSize, 8*BytesPerLong);
              "Turn off JVM mitigations related to Intel micro code "        \
              "mitigations for the Intel JCC erratum")                       \
                                                                             \
-  product(ccstr, CPUFeatures, NULL, "CPU feature set, "                     \
+  product(ccstr, CPUFeatures, NULL, RESTORE_SETTABLE, "CPU feature set, "   \
       "use -XX:CPUFeatures=0xnumber with -XX:CRaCCheckpointTo when you "    \
       "get an error during -XX:CRaCRestoreFrom on a different machine; "    \
       "-XX:CPUFeatures=native is the default; "                             \
+      "-XX:CPUFeatures=ignore will disable the CPU features check; "        \
```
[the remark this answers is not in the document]
AntonKozlov: That should make to use the current CPU features, the option was supported, apparently just not documented.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -1210,6 +1210,14 @@ void VM_Version::print_using_features_cr() {
   }
 }
 
+bool VM_Version::nonlibc_str_equals(const char *a, const char *b) {
+  while (*a && *b) {
+    if (*a++ != *b++)
+      return false;
+  }
+  return !*a && !*b;
+}
```
AntonKozlov, line 1219: Compiler can inline a call to strcmp if it recognizes the pattern, so this is not really safe. And since we are calling the argument check after VM_Crac::read_shm, it's a bit late to worry about libc... It may be just fine to leave the one who are give up the safety with possible crashes.

src/hotspot/cpu/x86/vm_version_x86.cpp
AntonKozlov: Agree. And we are unable to handle e.g. `generic` specified for restore, although it's documented. 

Having dedicated IgnoreCPUFeatures would simplify the implementation a lot, and we'd also could make it Experimental, as it should be.

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: That was a gross mistake in this patch, I have fixed it so that it no longer crashes. That was the whole effort of the original patch so that it does not crash without any error if it is being restored on under-featured CPU.

src/hotspot/cpu/x86/globals_x86.hpp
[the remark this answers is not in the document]
jankratochvil: There is now `-XX:+IgnoreCPUFeatures` instead.

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: There are currently no tests. Implemented the `-XX:+IgnoreCPUFeatures` option.

src/hotspot/cpu/x86/vm_version_x86.hpp
```diff
@@ -621,12 +621,14 @@ class VM_Version : public Abstract_VM_Version {
...
-  // C++17: Make ignore_glibc_not_using inline.
-  static bool ignore_glibc_not_using;
+  // C++17: Make _ignore_glibc_not_using inline.
+  static bool _ignore_glibc_not_using;
+  static bool _crac_restore_missing_features;
   static void nonlibc_tty_print_uint64(uint64_t num);
   static void nonlibc_tty_print_uint64_comma_uint64(uint64_t num1, uint64_t num2);
   static void print_using_features_cr();
-  /*[[noreturn]]*/ static void fatal_missing_features(uint64_t features_missing, uint64_t glibc_features_missing);
+  static bool nonlibc_str_equals(const char *a, const char *b);
```
AntonKozlov, line 630: Stray declaration

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -2741,14 +2744,15 @@ void VM_Version::crac_restore() {
...
     tty->print_raw(part1, sizeof(part1) - 1);
     nonlibc_tty_print_uint64_comma_uint64(_features & features_saved, _glibc_features & glibc_features_saved);
     static const char part2[] = " together with -XX:CRaCCheckpointTo when making a checkpoint file; specified -XX:CRaCRestoreFrom file contains CPU features ";
     tty->print_raw(part2, sizeof(part2) - 1);
     nonlibc_tty_print_uint64_comma_uint64(features_saved, glibc_features_saved);
-    fatal_missing_features(features_missing, glibc_features_missing);
+    missing_features(features_missing, glibc_features_missing);
   }
 
   auto supports_exit = [&](const char *supports, bool file, bool this_cpu) {
```
AntonKozlov, line 2758: Just a note, not a part of this PR: all `_supports_*`, except _supports_cx8, are hard-wired to `true`, and all of them mostly serve to communicate CPU capabilities to the shared code.

The `_supports_cx8` equals to supports_cmpxchg8(), and I assume this is supported pretty widely. And supports_cmpxchg8 implemented via` _features & CPU_CX8`, so that is tested along CPU features check.

It looks the whole handling of `_supports_*` can be removed, or at least be turned into assert()s.

AntonKozlov (commented):
LGTM except the declaration nit.

AntonKozlov (approved):

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: The `_supports_*` code It has been removed by #103 so it is no longer relevant. `CX8` is not supported by 32-bit CPUs.

AntonKozlov:
I'm not sure why it's stuck, may be need another `/integrate`?

PR #106, opened 2023-08-20
Prevent an assertion: Attempting to acquire lock PeriodicTask_lock...

jankratochvil:
Prevent an assertion: `Attempting to acquire lock PeriodicTask_lock/safepoint out of order with lock Threads_lock/safepoint-1 -- possible deadlock`

### DIFF
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -292,12 +292,37 @@ class WakeupClosure: public ThreadClosure {
   }
 };
 
+// Prevent an assertion: Attempting to acquire lock PeriodicTask_lock/safepoint out of order with lock Threads_lock/safepoint-1 -- possible deadlock
+// SafepointSynchronize::begin() did lock Threads_lock.
+// WatcherThread::unpark() requires PeriodicTask_lock to be held locked.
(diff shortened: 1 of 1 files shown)


src/hotspot/share/runtime/crac.cpp
```diff
@@ -292,12 +292,37 @@ class WakeupClosure: public ThreadClosure {
...
+    initialize_thread_current();
+    delete this;
+  }
+};
+
 static void wakeup_threads_in_timedwait() {
   WakeupClosure wc;
   Threads::java_threads_do(&wc);
 
-  MonitorLocker ml(PeriodicTask_lock, Mutex::_no_safepoint_check_flag);
```
AntonKozlov, line 299: I see the problematic locking is performed during the safepoint. The code was added in https://github.com/openjdk/crac/pull/85. AFAICS it's enough to wake up all threads to make them recompute the current time. As more simple approach, would it be enough to move wakeup_threads_in_timedwait out from VM Operation?

jankratochvil: Yes, you are right.

rvansa (approved):
Works for me.

AntonKozlov (approved):
LGTM, thank you!

PR #109, opened 2023-09-03
Fix ArchivedEnumTest.java testsuite regression

jankratochvil:
```
java.lang.RuntimeException: 'object points to a static field that may be reinitialized at runtime' found in stdout
        at jdk.test.lib.process.OutputAnalyzer.shouldNotContain(OutputAnalyzer.java:267)
        at ArchivedEnumTest.main(ArchivedEnumTest.java:49)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at com.sun.javatest.regtest.agent.MainActionHelper$AgentVMRunnable.run(MainActionHelper.java:333)
        at java.base/java.lang.Thread.run(Thread.java:1570)

[1.337s][warning][cds,heap] Archive heap points to a static field that may be reinitialized at runtime:
[1.337s][warning][cds,heap] Field: jdk/crac/impl/GlobalContext::GLOBAL_CONTEXT_IMPL_NAME
[1.337s][warning][cds,heap] Value: java.lang.String
[1.337s][warning][cds,heap] {0x00000000d7561560} - klass: 'java/lang/String'
[1.337s][warning][cds,heap]  - string: ""
[1.337s][warning][cds,heap]  - ---- fields (total size 3 words):
[1.337s][warning][cds,heap]  - private 'hash' 'I' @12  0 (0x00000000)
[1.337s][warning][cds,heap]  - private final 'coder' 'B' @16  0 (0x00)
[1.338s][warning][cds,heap]  - private 'hashIsZero' 'Z' @17  true (0x01)
[1.338s][warning][cds,heap]  - injected 'flags' 'B' @18  1 (0x01)
[1.338s][warning][cds,heap]  - private final 'value' '[B' @20  [B{0x00000000d7561578} (0xd7561578)
[1.338s][warning][cds,heap] --- trace begin ---
[1.338s][warning][cds,heap] [ 0] (shared string table)
[1.338s][warning][cds,heap] [ 1] {0x00000000d7561560} java.lang.String
[1.338s][warning][cds,heap] --- trace end ---
```

I haven't found a way how to fix [src/java.base/share/classes/jdk/crac/impl/GlobalContext.java](https://github.com/openjdk/crac/blob/crac/src/java.base/share/classes/jdk/crac/impl/GlobalContext.java) not regressing its current static field performance without adding this exception.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/109/head:pull/109` \
`$ git checkout pull/109`

`$ git checkout pull/109` \
`$ git pull https://git.openjdk.org/crac.git pull/109/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 109`

`$ git pr show -t 109`

</details>

<a href="https://git.openjdk.org/crac/pull/109.diff">https://git.openjdk.org/crac/pull/109.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/109#issuecomment-1704325414)

### DIFF
--- a/src/java.base/share/classes/jdk/crac/impl/GlobalContext.java
+++ b/src/java.base/share/classes/jdk/crac/impl/GlobalContext.java
@@ -5,12 +5,12 @@ import jdk.crac.Resource;
 import sun.security.action.GetPropertyAction;
 
 public class GlobalContext {
-    private static final String GLOBAL_CONTEXT_IMPL_PROP = "jdk.crac.globalContext.impl";
-    private static final String GLOBAL_CONTEXT_IMPL_NAME =
-        GetPropertyAction.privilegedGetProperty(GLOBAL_CONTEXT_IMPL_PROP, "");
+    private static final String IMPL_PROP = "jdk.crac.globalContext.impl";
+    private static final String IMPL_NAME =
+        GetPropertyAction.privilegedGetProperty(IMPL_PROP, "");
 
     public static Context<Resource> createGlobalContextImpl() {
-        return switch (GLOBAL_CONTEXT_IMPL_NAME) {
+        return switch (IMPL_NAME) {
             case "BlockingOrderedContext" -> new BlockingOrderedContext<>();
             case "OrderedContext" -> new OrderedContext<>();
             default -> new OrderedContext<>(); // cannot report as System.out/err are not initialized yet
--- a/src/java.base/share/classes/jdk/crac/impl/GlobalContext.java
+++ b/src/java.base/share/classes/jdk/crac/impl/GlobalContext.java
@@ -5,12 +5,10 @@ import jdk.crac.Resource;
 import sun.security.action.GetPropertyAction;
 
 public class GlobalContext {
-    private static final String IMPL_PROP = "jdk.crac.globalContext.impl";
-    private static final String IMPL_NAME =
-        GetPropertyAction.privilegedGetProperty(IMPL_PROP, "");
-
     public static Context<Resource> createGlobalContextImpl() {
-        return switch (IMPL_NAME) {
+        String impl_prop = "jdk.crac.globalContext.impl";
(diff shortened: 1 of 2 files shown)


rvansa:
Given that the global context is created only once, I think it would be OK to drop the field and obtain the config value in runtime.

However, I wonder if the option really works; looks like this PR is disabling only the error in the test, though the `[B]` option does not really fit (it does not end up as a compile-time-known deterministic value). Can you override this in runtime or would JDK always load the default value?

jankratochvil:
> Given that the global context is created only once, I think it would be OK to drop the field and obtain the config value in runtime.

OK. It is not great to base code only on its current callers but I do not know about a better solution.

> However, I wonder if the option really works; looks like this PR is disabling only the error in the test, though the `[B]` option does not really fit (it does not end up as a compile-time-known deterministic value). Can you override this in runtime or would JDK always load the default value?

I was feeling the `[B]` is probably wrong but I did expect there will be some better private final static Java solution for the patch. So I did not find this `[B]` part important as it will get dropped anyway (which it got).

src/java.base/share/classes/jdk/crac/impl/GlobalContext.java
```diff
@@ -5,12 +5,10 @@
 import sun.security.action.GetPropertyAction;
 
 public class GlobalContext {
-    private static final String GLOBAL_CONTEXT_IMPL_PROP = "jdk.crac.globalContext.impl";
```
AntonKozlov, line 8: GLOBAL_CONTEXT_IMPL_PROP can be here, no need to make it local

rvansa (approved):

AntonKozlov (approved):

PR #113, opened 2023-09-19
CRaC: Fix fds opened for logging

jankratochvil:
---------
### Progress

### Reviewers
 * [Radim Vansa](https://openjdk.org/census#rvansa) (@rvansa - Committer) ⚠️ Review applies to [b941490f](https://git.openjdk.org/crac/pull/113/files/b941490f0ff688a8fc8edd301e79c8e8a6e815a2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/113/head:pull/113` \
`$ git checkout pull/113`

`$ git checkout pull/113` \
`$ git pull https://git.openjdk.org/crac.git pull/113/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 113`

`$ git pr show -t 113`

</details>

<a href="https://git.openjdk.org/crac/pull/113.diff">https://git.openjdk.org/crac/pull/113.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/113#issuecomment-1727728628)

### DIFF
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -311,6 +311,11 @@ void VM_Crac::doit() {
   // dry-run fails checkpoint
   bool ok = true;
 
+  AsyncLogWriter* aio_writer = AsyncLogWriter::instance();
+  if (aio_writer) {
+    aio_writer->flush();
+    aio_writer->stop();
+  }
   LogConfiguration::close();
 
   if (!check_fds()) {
@@ -365,6 +370,9 @@ void VM_Crac::doit() {
   memory_restore();
 
   LogConfiguration::reopen();
+  if (aio_writer) {
+    aio_writer->resume();
+  }
 
   wakeup_threads_in_timedwait_vm();
 
@@ -427,8 +435,6 @@ Handle crac::checkpoint(jarray fd_arr, jobjectArray obj_arr, bool dry_run, jlong
   Universe::heap()->collect(GCCause::_full_gc_alot);
   Universe::heap()->set_cleanup_unused(false);
   Universe::heap()->finish_collection();
-
-  AsyncLogWriter::instance()->flush();
 
   VM_Crac cr(fd_arr, obj_arr, dry_run, (bufferedStream*)jcmd_stream);
   {
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -311,12 +311,6 @@ void VM_Crac::doit() {
   // dry-run fails checkpoint
   bool ok = true;
 
-  AsyncLogWriter* aio_writer = AsyncLogWriter::instance();
-  if (aio_writer) {
-    aio_writer->stop();
-  }
-  LogConfiguration::close();
-
   if (!check_fds()) {
     ok = false;
   }
@@ -367,11 +361,6 @@ void VM_Crac::doit() {
   VM_Version::crac_restore_finalize();
 
   memory_restore();
-
-  LogConfiguration::reopen();
-  if (aio_writer) {
-    aio_writer->resume();
-  }
 
   wakeup_threads_in_timedwait_vm();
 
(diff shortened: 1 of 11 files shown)


rvansa:
Sorry, I misunderstood the issue. I though that the gc log is causing trouble during checkpoint, but in fact the log is opened before `close_extra_descriptors` and this method mistakes it for accidentally inherited FD, and closes it. With your fix, the FD is marked as inherited and ignored during CRaC FD checks; CRIU seems to restore the open file correctly.

I think that the fix would be completely fine had we considered CRIU as the only C/R implementation, but we would need to deal with reopening this in minicriu. Therefore, unless you find the implementation difficult, I would opt for close/reopen implementation.

jankratochvil:
> I think that the fix would be completely fine had we considered CRIU as the only C/R implementation, but we would need to deal with reopening this in minicriu. Therefore, unless you find the implementation difficult, I would opt for close/reopen implementation.

I did not understand why to implement the reopening when it just works now. OK, I will implement it for `minicriu`.

jankratochvil:
I am not sure when to call those functions, whether in VM Thread or before...

src/hotspot/share/runtime/crac.cpp
```diff
@@ -422,6 +428,8 @@ Handle crac::checkpoint(jarray fd_arr, jobjectArray obj_arr, bool dry_run, jlong
   Universe::heap()->set_cleanup_unused(false);
   Universe::heap()->finish_collection();
 
+  AsyncLogWriter::instance()->flush();
```
rvansa, line 431: As a non-java thread, AsyncLogWriter does not participate in the safepoint protocol. What happens if another non-java thread enqueues a message to be written while the output is closed?

jankratochvil: OK, I have to improve it, thanks.

jankratochvil: I hope the stop() and resume() do fix it. But there is no testcase for that, do you want it?

rvansa: I guess that a testcase for the synchronization would be quite difficult to do, as you'd be trying to simulate a race.

However, I am not convinced the `stop()` and `resume()` work correctly: the log writer does not need to own the lock while writing messages to the (potentially nulled) outputs; you flush the buffer and then acquire the lock, effectively blocking all log producers from enqueuing further messages. However if someone enqueues a message right after flush, the log writer has a chance to dequeue it (actually swap queues) and even if you acquire lock then, continue writing these messages to the closed output.

rvansa:
About when to call them: IMO the more you do outside the VM thread the better. OpenJDK CRaC has gone the way where we perform most of the C/R handling in parallel with running the application, as opposed to the approach in OpenJ9 where even the Java callbacks (Resource.beforeCheckpoint/afterRestore) are performed in a single-threaded mode (which has its own advantage in a reduced need for complicated synchronization), so it is better to keep to philosophy consistent.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -422,12 +424,23 @@ Handle crac::checkpoint(jarray fd_arr, jobjectArray obj_arr, bool dry_run, jlong
...
+  }
+  LogConfiguration::close();
+
   VM_Crac cr(fd_arr, obj_arr, dry_run, (bufferedStream*)jcmd_stream);
   {
     MutexLocker ml(Heap_lock);
     VMThread::execute(&cr);
   }
   if (cr.ok()) {
+    LogConfiguration::reopen();
```
rvansa, line 439: Looks like the log stays closed & blocked if the checkpoint fails?

src/hotspot/share/runtime/crac.cpp
jankratochvil: Race reproducers are also a part of testsuites - [ptrace-testsuite](https://sourceware.org/systemtap/wiki/utrace/tests) limits their run by its `TESTTIME` parameter. I sure did not try how easily would be to reproduce this race and I am sure happy not to write such a testcase.

src/hotspot/share/runtime/crac.cpp
jankratochvil: Hopefully it is fixed now?

jankratochvil:
There is a GHA error...

rvansa (approved):
LGTM, thank you!

jankratochvil:
Please do not sponsor it yet, maybe there is really some OSX compilation regression.

rvansa:
The build looks good, please set to integrate again.

PR #125, opened 2023-10-06
Stabilize GHA testsuite

rvansa:
At this point Github Actions always show some failures; let's ignore platforms that we don't handle for CRaC and some problematic tests in mainline JDK.

### DIFF
--- a/.github/workflows/build-cross-compile.yml
+++ b/.github/workflows/build-cross-compile.yml
@@ -58,9 +58,9 @@ jobs:
         target-cpu:
           - aarch64
           - arm
-          - s390x
-          - ppc64le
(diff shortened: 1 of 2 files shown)


AntonKozlov (approved):

wkia:
Why should we exclude builds, and why don't merge from upstream to fix them?

rvansa:
@wkia We really don't do much for the other platforms in CRaC, don't test it etc. I don't see a reason to delay GHA completion because of these regularly. Even experimental support is behind the horizon.

Merging upstream changes is certainly a good thing - I think that Anton is planning on that, please sync up with him.

PR #129, opened 2023-10-17
Fixing crash on restore when user name is not set

wkia:
This change fixes a crash occuring during restore in a container when user name is not set.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/129/head:pull/129` \
`$ git checkout pull/129`

`$ git checkout pull/129` \
`$ git pull https://git.openjdk.org/crac.git pull/129/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 129`

`$ git pr show -t 129`

</details>

<a href="https://git.openjdk.org/crac/pull/129.diff">https://git.openjdk.org/crac/pull/129.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/129#issuecomment-1765979644)

### DIFF
--- a/src/hotspot/os/posix/perfMemory_posix.cpp
+++ b/src/hotspot/os/posix/perfMemory_posix.cpp
@@ -1401,6 +1401,9 @@ bool PerfMemoryLinux::restore() {
 
   int vmid = os::current_process_id();
   char* user_name = get_user_name(geteuid());
+  if (!user_name) {
+    return false;
+  }
   char* dirname = get_user_tmp_dir(user_name, vmid, -1);
   if (!make_user_tmp_dir(dirname)) {
     return false;


src/hotspot/os/posix/perfMemory_posix.cpp
```diff
@@ -1401,6 +1401,9 @@ bool PerfMemoryLinux::restore() {
 
   int vmid = os::current_process_id();
   char* user_name = get_user_name(geteuid());
+  if (!user_name) {
+    return false;
```
rvansa, line 1405: Could we accompany this with an error message?

wkia: I'm not sure about an error message as it doesn't fail restore procedure. 
The same scenario from [here ](https://github.com/openjdk/crac/blob/crac/src/hotspot/os/posix/perfMemory_posix.cpp#L1032) doesn't show any message also.

rvansa: OK, that scenario would end up with an error message only with [develop flags](https://github.com/openjdk/crac/blob/crac/src/hotspot/os/posix/perfMemory_posix.cpp#L1261).

rvansa (approved):

PR #131, opened 2023-10-19
Terminate restored process when criuengine restorewait exits

rvansa:
With criuengine the restored process gets restorewait process as its parent; scripts not expecting two processes might signal (e.g. terminate) the parent process but the actual restored process would get orphaned.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/131/head:pull/131` \
`$ git checkout pull/131`

`$ git checkout pull/131` \
`$ git pull https://git.openjdk.org/crac.git pull/131/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 131`

`$ git pr show -t 131`

</details>

<a href="https://git.openjdk.org/crac/pull/131.diff">https://git.openjdk.org/crac/pull/131.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/131#issuecomment-1770689505)

### DIFF
--- a/src/hotspot/os/linux/crac_linux.cpp
+++ b/src/hotspot/os/linux/crac_linux.cpp
@@ -32,6 +32,10 @@
 #include "logging/logConfiguration.hpp"
 #include "classfile/classLoader.hpp"
 
+#include <limits.h>
+#include <string.h>
+#include <sys/prctl.h>
+
 class FdsInfo {
 public:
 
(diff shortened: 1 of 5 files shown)


src/hotspot/os/linux/crac_linux.cpp
```diff
@@ -462,3 +466,46 @@ bool crac::read_bootid(char *dest) {
   }
   return true;
 }
+
+void crac::set_terminate_with_parent() {
+  if (prctl(PR_SET_PDEATHSIG, SIGTERM)) {
```
AntonKozlov, line 471: The signal that has killed the parent should be SIGKILL, so it should be specified here as well.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -237,6 +246,10 @@ static int checkpoint_restore(int *shmid) {
   // TODO add sync processing
 #endif //LINUX
 
+  if (ends_with(_crengine, "criuengine")) {
```
AntonKozlov, line 249: It's strange the VM assumes the process hierarchy as implemented by criuengine. The cleaner way would be implementing something like this in criuengine itself. Probably setting PR_SET_PDEATHSIG will be complicated, but that can be overcomed by adding the code in our criu fork, if required. Anyway, the VM should not be aware that much how the checkpoint engine behaves.

rvansa:
Since the implementation should land in CRIU I am closing this PR.

PR #134, opened 2023-10-25
Fix CPUFeatures crash on new->old CPU

jankratochvil:
- reproducible on i7-1165G7 "qemu-kvm -cpu host" for a checkpoint and "qemu-kvm -cpu SandyBridge" for its restore

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/134/head:pull/134` \
`$ git checkout pull/134`

`$ git checkout pull/134` \
`$ git pull https://git.openjdk.org/crac.git pull/134/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 134`

`$ git pr show -t 134`

</details>

<a href="https://git.openjdk.org/crac/pull/134.diff">https://git.openjdk.org/crac/pull/134.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/134#issuecomment-1779515285)

### DIFF
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -515,24 +515,25 @@ void crac::restore() {
 }
 
 bool CracRestoreParameters::read_from(int fd) {
+  // crac_restore_finalize() may terminate the process if we run on (older) CPU where glibc string functions may crash.
+  // The header is read first as all the code of this function below is difficult to implement without the string functions.
+  header hdr;
+  // FIXME: Fix incomplete reads.
+  if (read(fd, &hdr, sizeof(hdr)) != sizeof(hdr)) {
+    perror("read (ignoring restore parameters)");
+    return false;
+  }
+  if (!hdr._ignore_cpu_features) {
+    VM_Version::crac_restore_finalize();
+  }
+
   struct stat st;
-  if (fstat(fd, &st)) {
+  if (fstat(fd, &st) || st.st_size < (ssize_t)sizeof(hdr)) {
     perror("fstat (ignoring restore parameters)");
     return false;
   }
 
-  // crac_restore_finalize() may terminate the process if we run on (older) CPU where glibc string functions may crash.
-  // The flag is stored separately as all the code of this function below is difficult to implement without the string functions.
-  bool IgnoreCPUFeatures_local;
-  if (read(fd, &IgnoreCPUFeatures_local, sizeof(IgnoreCPUFeatures_local)) != sizeof(IgnoreCPUFeatures_local)) {
-    perror("read (ignoring restore parameters)");
-    return false;
-  }
-  if (!IgnoreCPUFeatures_local) {
-    VM_Version::crac_restore_finalize();
-  }
-
-  size_t contents_size = st.st_size - sizeof(IgnoreCPUFeatures_local);
+  size_t contents_size = st.st_size - sizeof(hdr);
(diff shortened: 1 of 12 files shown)


src/hotspot/share/runtime/crac.cpp
```diff
@@ -521,8 +521,20 @@ bool CracRestoreParameters::read_from(int fd) {
...
-  if (read(fd, contents, st.st_size) < 0) {
+  // crac_restore_finalize() may terminate the process if we run on (older) CPU where glibc string functions may crash.
+  // The flag is stored separately as all the code of this function below is difficult to implement without the string functions.
+  bool IgnoreCPUFeatures_local;
+  if (read(fd, &IgnoreCPUFeatures_local, sizeof(IgnoreCPUFeatures_local)) != sizeof(IgnoreCPUFeatures_local)) {
+    perror("read (ignoring restore parameters)");
+    return false;
+  }
+  if (!IgnoreCPUFeatures_local) {
+    VM_Version::crac_restore_finalize();
```
rvansa, line 532: Is this function idempotent? It's called from `VM_Crac::doit()` as well (and it should be since reading the shm is optional), but could calling it twice cause any trouble?

src/hotspot/share/runtime/crac_structs.hpp
```diff
@@ -208,12 +212,23 @@ class VM_Crac: public VM_Operation {
...
+    *d++ = '0' + id % 10;
+  }
 public:
   CracSHM(int id) {
-    int shmpathlen = snprintf(_path, sizeof(_path), "/crac_%d", id);
-    if (shmpathlen < 0 || sizeof(_path) <= (size_t)shmpathlen) {
-      fprintf(stderr, "shmpath is too long: %d\n", shmpathlen);
+    assert(id > 0, "id is expected to be a PID and therefore > 0");
+    char *d = _path;
+    const char prefix[] = "/tmp/cracshm.";
```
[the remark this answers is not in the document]
jankratochvil: `shm_open` calls some string functions. So with the current (mis)design of CPUFeatures it cannot be used. One could call directly the syscall but that is too ugly. And then I do not see why to use SHM with the current design. One process creates the object, writes its data, closes the object. Second process opens the object, deletes the object, reads the data, closes the object. That is a normal usage of a tmp file. Shared memory has its use for some simultaneous data exchange but this is not the case so shared memory is overengineered. There is also limited amount of shared memory available so it is wasteful. Linuxes nowadays even have /tmp as tmpfs by default so even no disk operations are involved.
I agree "/tmp" should not be in this platform-independent file, I will fix that, thanks.
There are more portable functions for temporary files (`mktemp()` et al.) but these all use string functions.

src/hotspot/share/runtime/crac.cpp
jankratochvil: It is but it was just a luck. At least a comment there is wrong. I will improve it, thanks.

rvansa (approved):
Sorry for keeping this waiting for a long time. In the past I wasn't sure about using `/tmp/` rather than `shm_open` target but since `os::get_temp_directory()` has this part hard-coded, too, I think it's acceptable to use that one.

I would prefer `crac_restore_finalize` to use an enum rather than a `char` but otherwise this looks good - and if it fixes some usecase, let's integrate it.

jankratochvil:
The move the CPUFeatures protection to the parent JDK before the JDK is restored is done but downstream. This patch is no longer relevant.

PR #112, opened 2023-09-16
Simplify CPUFeatures code

jankratochvil:
There is no functionality change. I guess the new code should be more simple and shorter. It was originally suggested by @AntonKozlov to use more functions than macros.

```
 1 file changed, 44 insertions(+), 82 deletions(-)
 ```
So I find it clearly an improvement.
The readable sub-commit is: https://github.com/openjdk/crac/pull/112/commits/6d9cb72b7a838dd4c9f107c5b71c4275005a0c23
Otherwise it gets messed up by the other commit just renaming things.
As an explanation: Original code had two lists of the same CPU features. An `EXCESSIVE` list and an `GLIBC_DISABLE` list. The had to be kept in sync (their sets being equal) which was sanity checked:
```
    if (PASTE_TOKENS(disable_handled_, kind) != PASTE_TOKENS(excessive_handled_, kind))                                                  \
```
So now there is the list just once (`EXCESSIVE`, the `GLIBC_DISABLE` one has been deleted). It looks stupid but when coding it I did not see it.
Coding both variants of #136 was needlessly more difficult without this pull request applied and now I have to rebase this pull request.

---------
### Progress
- [x] Change must not contain extraneous whitespace

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/112/head:pull/112` \
`$ git checkout pull/112`

Update a local copy of the PR: \
`$ git checkout pull/112` \
`$ git pull https://git.openjdk.org/crac.git pull/112/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 112`

View PR using the GUI difftool: \
`$ git pr show -t 112`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/crac/pull/112.diff">https://git.openjdk.org/crac/pull/112.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/112#issuecomment-1722257521)

PR #136, opened 2023-10-31
Remove INCLUDE_LD_SO_LIST_DIAGNOSTICS

jankratochvil:
It was introduced in glibc-2.34 together with preferred CPU_FEATURE_ACTIVE.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/136/head:pull/136` \
`$ git checkout pull/136`

`$ git checkout pull/136` \
`$ git pull https://git.openjdk.org/crac.git pull/136/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 136`

`$ git pr show -t 136`

</details>

<a href="https://git.openjdk.org/crac/pull/136.diff">https://git.openjdk.org/crac/pull/136.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/136#issuecomment-1786565042)


src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -855,8 +853,7 @@ uint64_t VM_Version::CPUFeatures_parse(uint64_t &glibc_features) {
   return -1;
 }
 
-#if INCLUDE_LD_SO_LIST_DIAGNOSTICS
-
+//#if INCLUDE_LD_SO_LIST_DIAGNOSTICS
```
rvansa, line 856: Please remove commented out code.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -942,12 +941,12 @@ bool VM_Version::glibc_env_set(char *disable_str) {
   char *env_val = disable_str;
   const char *env = getenv(TUNABLES_NAME);
   if (env && strcmp(env, env_val) == 0) {
-#if !INCLUDE_CPU_FEATURE_ACTIVE && !INCLUDE_LD_SO_LIST_DIAGNOSTICS
-    if (ShowCPUFeatures) {
-      tty->print_cr("Environment variable already set, both glibc CPU_FEATURE_ACTIVE and ld.so --list-diagnostics are unavailable - re-exec suppressed: " TUNABLES_NAME "=%s", env);
+    if (!INCLUDE_CPU_FEATURE_ACTIVE && !INCLUDE_LD_SO_LIST_DIAGNOSTICS) {
```
rvansa, line 944: Have you tested compilation with these disabled? Looks to me like the compiler could complain about unreachable code when the condition is always false...

rvansa (commented):
While I like the certainty that the code is always compilable, I wonder if handling optional features like this (using macros in conditions) is a pattern that is used elsewhere in the JDK?

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: I wanted to indicate this function is used only for `INCLUDE_LD_SO_LIST_DIAGNOSTICS`. I can also put there a full sentence.

src/hotspot/cpu/x86/vm_version_x86.cpp
jankratochvil: I have tried all the 3 compilation cases only now. I do not see such compiler warnings (although different compilations and compilers produce different warnings) but I could fix some other problems. Which will make me update #112.

jankratochvil:
> While I like the certainty that the code is always compilable, I wonder if handling optional features like this (using macros in conditions) is a pattern that is used elsewhere in the JDK?

I haven't found such case. I expect to leave such decisions up to OpenJDK reviewers during CRaC merge request.

OTOH to make things easier - the whole `INCLUDE_LD_SO_LIST_DIAGNOSTICS` code should be probably removed. The support for `INCLUDE_LD_SO_LIST_DIAGNOSTICS` will probably never be used as it was added to glibc only few months before the (better) `INCLUDE_CPU_FEATURE_ACTIVE` support. Before doing such removal for real one should check whether any major/supported Linux distribution has been released with such glibc version supporting only `INCLUDE_LD_SO_LIST_DIAGNOSTICS` without `INCLUDE_CPU_FEATURE_ACTIVE`. I unfortunately did not check what was when implemented in glibc while writing the `INCLUDE_LD_SO_LIST_DIAGNOSTICS` code.

rvansa:
It's in our interest to provoke as few discussions as possible, to get CRaC merged without extra delay. So, the code style is questionable and there's actually very little chance it will be useful, do you still consider this worth merging? Or would you prefer to remove it altogether instead? (Could be done in this very PR)

jankratochvil:
So if there is the decision to remove it I will investigate first there is really no interesting Linux OS release with the "problematic" glibc version.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -944,9 +862,9 @@ bool VM_Version::glibc_env_set(char *disable_str) {
   char *env_val = disable_str;
   const char *env = getenv(TUNABLES_NAME);
   if (env && strcmp(env, env_val) == 0) {
-    if (!INCLUDE_CPU_FEATURE_ACTIVE && !INCLUDE_LD_SO_LIST_DIAGNOSTICS) {
+    if (!INCLUDE_CPU_FEATURE_ACTIVE) {
       if (ShowCPUFeatures) {
-        tty->print_cr("Environment variable already set, both glibc CPU_FEATURE_ACTIVE and ld.so --list-diagnostics are unavailable - re-exec suppressed: " TUNABLES_NAME "=%s", env);
+        tty->print_cr("Environment variable already set, both glibc CPU_FEATURE_ACTIVE is unavailable - re-exec suppressed: " TUNABLES_NAME "=%s", env);
```
rvansa, line 867: The "both" word is extra in here I guess.

rvansa:
I admit I have no idea about the IBT and SHSTK bits; it might be beneficial to include the description of what you're changing about those in the PR description so that whoever stumbles upon the commit has some pointers.

jankratochvil:
> I admit I have no idea about the IBT and SHSTK bits; it might be beneficial to include the description of what you're changing about those in the PR description so that whoever stumbles upon the commit has some pointers.

That is an unrelated change. So I will split it to another pull request.

rvansa (approved):
LGTM, thanks!

PR #137, opened 2023-11-02
Ignore open files in /var/lib/sss/mc

rvansa:
I was considering different ways to fix this - there are actually two problems:
1) how to detect that?
* This PR whitelisst all files in the directory (we could explicitly name `passwd`, `group`, `sid` and `initgroups`...)
* We could manually check all places in JDK that call `getpwuid*`, `getpwname*`, `getgrgid*`and `getgrname*` and maybe some other functions, and diff FDs opened before/after the call. However this a) has performance impact b) is prone to races
* Intercept the call: either catch syscalls (ptrace or seccomp), or patching `sss_open_cloexec` in memory, or preloading it? Rather complicated.
2) what to do with the open FD?
* Current solution is to leave this up to CRIU (or another C/R engine). Again the simplest
* We could close this; the FD is `fstat`-validated later on, so this would cause errors from these functions. We would need to patch the library, de-initializing the implementation (also risky).

Normally I would try to include a testcase but given that this is system-dependent (and the issue does not appear even in CentOS Stream 9 container) I've only did manual testing.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/137/head:pull/137` \
`$ git checkout pull/137`

`$ git checkout pull/137` \
`$ git pull https://git.openjdk.org/crac.git pull/137/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 137`

`$ git pr show -t 137`

</details>

<a href="https://git.openjdk.org/crac/pull/137.diff">https://git.openjdk.org/crac/pull/137.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/137#issuecomment-1790910293)

### DIFF
--- a/src/hotspot/os/linux/crac_linux.cpp
+++ b/src/hotspot/os/linux/crac_linux.cpp
@@ -332,8 +332,8 @@ bool VM_Crac::check_fds() {
     // On some systems using SSSD files in this directory are left open
     // after calling getpwuid_r, getpwname_r, getgrgid_r, getgrname_r
     // or other functions in this family.
-#define SSS_NSS_MEMCACHE "/var/lib/sss/mc/"
-    if (!strncmp(details, SSS_NSS_MEMCACHE, sizeof(SSS_NSS_MEMCACHE) - 1)) {
+    constexpr const char *sss_nss_memcache = "/var/lib/sss/mc/";
+    if (!strncmp(details, sss_nss_memcache, strlen(sss_nss_memcache))) {
       print_resources("OK: SSSD cache\n");
       continue;
     }
--- a/src/hotspot/os/linux/crac_linux.cpp
+++ b/src/hotspot/os/linux/crac_linux.cpp
@@ -22,6 +22,8 @@
  */
 
 // no precompiled headers
+#include <string.h>
+
 #include "jvm.h"
 #include "perfMemory_linux.hpp"
 #include "runtime/crac_structs.hpp"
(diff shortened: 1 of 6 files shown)


src/hotspot/os/linux/crac_linux.cpp
```diff
@@ -327,6 +329,15 @@ bool VM_Crac::check_fds() {
       }
     }
 
+    // On some systems using SSSD files in this directory are left open
+    // after calling getpwuid_r, getpwname_r, getgrgid_r, getgrname_r
+    // or other functions in this family.
+#define SSS_NSS_MEMCACHE "/var/lib/sss/mc/"
+    if (!strncmp(details, SSS_NSS_MEMCACHE, sizeof(SSS_NSS_MEMCACHE) - 1)) {
```
wkia, line 336: I believe GCC can manage with `const char *` and `strlen` call in compile time here. `constexpr` could be addedaswell, but nor necessary I guess.

rvansa: Right; this is not the most perf-critical part but for the fun I've checked the assembly and since this is 16 bytes it is easily replaced with 2 64-bit constant comparisons.
Btw. for some reason the compiler emits warning for just `constexpr char *...`, I have to write `constexpr const char *...`.

rvansa:
@AntonKozlov requested that the ignored open files might be configurable, hence the last commit.

src/hotspot/os/linux/globals_linux.hpp
```diff
@@ -114,4 +114,9 @@ define_pd_global(bool, UseLargePages, false);
 define_pd_global(bool, UseLargePagesIndividualAllocation, false);
 define_pd_global(bool, UseThreadPriorities, true) ;
 
+// On some systems using SSSD files in this directory are left open
+// after calling getpwuid_r, getpwname_r, getgrgid_r, getgrname_r
+// or other functions in this family.
+define_pd_global(ccstrlist, CRAllowedOpenFilePrefixes, "/var/lib/sss/mc/");
```
AntonKozlov, line 120: This have to be replicated over all platforms, e.g. macos build is broken https://github.com/rvansa/crac/actions/runs/6784349063/job/18440498579

rvansa: Spot on. For some reason my VSCode left those files unsaved (and therefore not comitted).

AntonKozlov (approved):
LGTM, thank you!

PR #116, opened 2023-09-25
Close files opened by Decoder before checkpoint

rvansa:
Native memory tracking needs to resolve some addresses for stack unwinding and the decoders keep some shared library files open as a cache. Since the decoder instances can be re-allocated anytime this fix just destroys them before a checkpoint.

---------
### Progress
- [x] Change must not contain extraneous whitespace

### Reviewers
 * [Anton Kozlov](https://openjdk.org/census#akozlov) (@AntonKozlov - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/116/head:pull/116` \
`$ git checkout pull/116`

Update a local copy of the PR: \
`$ git checkout pull/116` \
`$ git pull https://git.openjdk.org/crac.git pull/116/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 116`

View PR using the GUI difftool: \
`$ git pr show -t 116`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/crac/pull/116.diff">https://git.openjdk.org/crac/pull/116.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/116#issuecomment-1733708271)

PR #138, opened 2023-11-09
Fix Windows build of Decoder

rvansa:
Fixes build failure on Windows introduced in https://github.com/openjdk/crac/pull/116

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/138/head:pull/138` \
`$ git checkout pull/138`

`$ git checkout pull/138` \
`$ git pull https://git.openjdk.org/crac.git pull/138/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 138`

`$ git pr show -t 138`

</details>

<a href="https://git.openjdk.org/crac/pull/138.diff">https://git.openjdk.org/crac/pull/138.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/138#issuecomment-1803717744)

### DIFF
--- a/src/hotspot/share/utilities/decoder.hpp
+++ b/src/hotspot/share/utilities/decoder.hpp
@@ -122,6 +122,9 @@ public:
 
   static void print_state_on(outputStream* st);
 
+#ifdef _WINDOWS
+  static void before_checkpoint() { /* noop */ }
+#else
   static void before_checkpoint();
 
 protected:
@@ -142,7 +145,8 @@ private:
 protected:
   static Mutex* shared_decoder_lock();
 
-  friend class DecoderLocker;
+#endif // !_WINDOWS
+
 };
 
 #endif // SHARE_UTILITIES_DECODER_HPP


AntonKozlov (approved):

src/hotspot/share/utilities/decoder.hpp
```diff
@@ -122,6 +122,9 @@ class Decoder : AllStatic {
 
   static void print_state_on(outputStream* st);
 
+#ifdef _WINDOWS
```
wkia, line 125: I'm not sure we have to use #ifdef WINDOWS here. 
There is src\hotspot\os\windows\decoder_windows.cpp file already, I beileve it should be there.

rvansa: The thing is that `decoder.cpp` has most of the code guarded by similar ifdefs. I don't think it's right to expose something in headers that would fail during linking, so I've put most of it to the else branch.
Yes, I could have used just a declaration in here and added the noop impl to `decoder_windows.cpp`; this way I am modifying just 1 file rather than 2.

PR #141, opened 2023-11-16
Merge jdk/master branch

AntonKozlov:
Merge jdk/master to crac.


rvansa:
> Unable to find an artifact with the name: bundles-linux-x64-debug

Is this a random infra failure? (can you trigger GHA again?)

Otherwise, I don't think that we're going to review anything. It's good to see that testsuite mostly passes.

AntonKozlov:
Looks like a intermittent failure. I have re-run the tests, but GHA does not support re-run of failed tests. I'm doing another try with re-running everything.

rvansa:
Could you push this? We can't do this through Skara, and merge commits through UI are disabled via GitHub settings.

PR #143, opened 2023-11-21
Propagating CRaC event to JVMTI agents

wkia:
This change implements propagation of CRaC events to JVMTI.

- New JVMTI extension events are introduced - `EXT_EVENT_CRAC_BEFORE_CHECKPOINT` and `EXT_EVENT_CRAC_AFTER_RESTORE`.
- JVMTI agents should subscribe/unsubscribe to the events using `SetExtensionEventCallback` JVMTI method. 
  - Setting/unsetting a callback will enable/disable sending an appropriate event to an agent, i.e. there is no need to additionally call `SetEventNotificationMode` method by an agent.
  - See `test/jdk/jdk/crac/jvmtiEvents/libCracJvmtiAgent.c` for an example of CRaC-ready JVMTI agent

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/143/head:pull/143` \
`$ git checkout pull/143`

`$ git checkout pull/143` \
`$ git pull https://git.openjdk.org/crac.git pull/143/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 143`

`$ git pr show -t 143`

</details>

<a href="https://git.openjdk.org/crac/pull/143.diff">https://git.openjdk.org/crac/pull/143.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/143#issuecomment-1820890841)

### DIFF
--- a/src/hotspot/share/prims/jvmtiEventController.cpp
+++ b/src/hotspot/share/prims/jvmtiEventController.cpp
@@ -824,6 +824,14 @@ JvmtiEventControllerPrivate::set_extension_event_callback(JvmtiEnvBase *env,
     case EXT_EVENT_VIRTUAL_THREAD_UNMOUNT :
       ext_callbacks->VirtualThreadUnmount = callback;
       break;
+    case EXT_EVENT_CRAC_BEFORE_CHECKPOINT :
+      ext_callbacks->CracBeforeCheckpoint = callback;
+      env->env_event_enable()->set_user_enabled(event_type, enabling);
+      break;
+    case EXT_EVENT_CRAC_AFTER_RESTORE :
+      ext_callbacks->CracAfterRestore = callback;
+      env->env_event_enable()->set_user_enabled(event_type, enabling);
+      break;
     default:
       ShouldNotReachHere();
   }
--- a/src/hotspot/share/prims/jvmtiUtil.hpp
+++ b/src/hotspot/share/prims/jvmtiUtil.hpp
@@ -57,7 +57,7 @@ public:
       return _event_threaded[num];
     }
     if (num >= EXT_MIN_EVENT_TYPE_VAL && num <= EXT_MAX_EVENT_TYPE_VAL) {
-      return (num != EXT_EVENT_CLASS_UNLOAD);
+      return (num != EXT_EVENT_CLASS_UNLOAD && num != EXT_EVENT_CRAC_BEFORE_CHECKPOINT && num != EXT_EVENT_CRAC_AFTER_RESTORE);
     }
     ShouldNotReachHere();
     return false;
--- a/src/hotspot/share/prims/jvmtiEventController.hpp
+++ b/src/hotspot/share/prims/jvmtiEventController.hpp
@@ -44,10 +44,12 @@ class JvmtiEnvBase;
 
 // Extension events start JVMTI_MIN_EVENT_TYPE_VAL-1 and work towards 0.
 typedef enum {
+  EXT_EVENT_CRAC_AFTER_RESTORE = JVMTI_MIN_EVENT_TYPE_VAL-5,
+  EXT_EVENT_CRAC_BEFORE_CHECKPOINT = JVMTI_MIN_EVENT_TYPE_VAL-4,
   EXT_EVENT_VIRTUAL_THREAD_UNMOUNT = JVMTI_MIN_EVENT_TYPE_VAL-3,
   EXT_EVENT_VIRTUAL_THREAD_MOUNT = JVMTI_MIN_EVENT_TYPE_VAL-2,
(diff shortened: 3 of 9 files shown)


rvansa (approved):
Looks good, and kudos for the test! How is it built (into shared library I suppose?) I don't see any modifications in Makefiles...

wkia:
@rvansa  
> How is it built (into shared library I suppose?) 

Yes, it produces a shared library, with different names for different OS.
It is processed [here](https://github.com/openjdk/crac/blob/crac/make/test/JtregNativeJdk.gmk#L152), all we need is to create a file with "lib" name suffix, or "exe" in case of executable. It will be built and placed somewhere in build folder (`.../test/jdk/jtreg/...`)

src/hotspot/share/prims/jvmtiUtil.hpp
```diff
@@ -57,7 +57,7 @@ class JvmtiUtil : AllStatic {
       return _event_threaded[num];
     }
     if (num >= EXT_MIN_EVENT_TYPE_VAL && num <= EXT_MAX_EVENT_TYPE_VAL) {
-      return (num != EXT_EVENT_CLASS_UNLOAD);
+      return (num != EXT_EVENT_CLASS_UNLOAD && num != EXT_EVENT_CRAC_BEFORE_CHECKPOINT && num != EXT_EVENT_CRAC_AFTER_RESTORE);
```
AntonKozlov, line 60: Any idea why `EXT_EVENT_VIRTUAL_THREAD_MOUNT`/`UNMOUNT` are not listed here?

src/hotspot/share/prims/jvmtiEventController.cpp
```diff
@@ -824,6 +824,14 @@ JvmtiEventControllerPrivate::set_extension_event_callback(JvmtiEnvBase *env,
     case EXT_EVENT_VIRTUAL_THREAD_UNMOUNT :
       ext_callbacks->VirtualThreadUnmount = callback;
       break;
+    case EXT_EVENT_CRAC_BEFORE_CHECKPOINT :
+      ext_callbacks->CracBeforeCheckpoint = callback;
+      env->env_event_enable()->set_user_enabled(event_type, enabling);
```
AntonKozlov, line 829: Why UNLOAD/MOUNT/UNMOUNT do not require this?

src/hotspot/share/prims/jvmtiEventController.cpp
AntonKozlov, line 1: I see some potential places to modify, e.g. line 699. It would make sense at least place a comment where new extenstion eventns are intentionally not added.

wkia: "`should_post_NNN`" API is created for JDK classes/entities, which are originators of NNN event, to make them deciding if they need to post events or not - if no one subscribes, or no subscriber wants to receive it, they shouldn't. I guess this was done for JDK performance purposes.

For checkpoint/restore, we could create similar API to check if `crac::checkpoint()` should post an event or not, but currently it seems unecessary in my opinion, because CRaC events aren't events are posted dozens times per second or more. It may make sense in case of hundreds of JVMTI agents not supposing to consume CRaC events - in this case, posting an event will iterate all JVMTI agents with no payload. For now it seems like one more place for conflicts in future merging.

src/hotspot/share/prims/jvmtiEventController.cpp
wkia: For EXT_EVENT_CLASS_UNLOAD, it was here from the very beginning (https://github.com/openjdk/crac/blob/8153779ad32d1e8ddd37ced826c76c7aafc61894/hotspot/src/share/vm/prims/jvmtiEventController.cpp#L693), until VitualThreads implementation changed it. 
Now a consumer (an agent) should set a callback for an event, and then can control recieving an event by enabling/disabling it with a call of `SetEventNotificationMode ` function. I guess, this was done to make the event control flow easier for a consumer. This makes sense in case an event is supposed to posted regularly, that is not a case for CRaC. Again, we could implement a similar way, but it still seems like one more place for conflicts in future merging.

src/hotspot/share/prims/jvmtiUtil.hpp
wkia: This function checks if an event is related to a particular thread. This is `true` only for `VIRTUAL_THREAD_MOUNT/UNMOUNT`. For both `EXT_EVENT_CLASS_UNLOAD` and for CRaC events it doesn't make sense - so `false` is returned.

wkia:
@AntonKozlov 
Hi 
Could you continue the review, please?

PR #145, opened 2023-11-24
Improve C/R exception printout

rvansa:
Some users might get confused by the inner exceptions reported during C/R as *suppressed* exceptions. This PR changes the printout to make it look as if the exception had multiple causes. For example the DryRunTest will report this:
```
jdk.crac.CheckpointException: Failed with 2 inner exceptions
Cause 1/2: java.lang.RuntimeException: should not pass
	at DryRunTest$CRResource.beforeCheckpoint(DryRunTest.java:47)
	at java.base/jdk.crac.impl.AbstractContext.invokeBeforeCheckpoint(AbstractContext.java:44)
	... (redacted)
Cause 2/2: jdk.crac.impl.CheckpointOpenFileException: /tmp/jtreg-DryRunTest6956725915963168340.tmp
	at java.base/jdk.internal.crac.JDKFileResource.lambda$beforeCheckpoint$1(JDKFileResource.java:89)
	at java.base/jdk.crac.Core.checkpointRestore1(Core.java:174)
        ... (redacted)
```

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/145/head:pull/145` \
`$ git checkout pull/145`

`$ git checkout pull/145` \
`$ git pull https://git.openjdk.org/crac.git pull/145/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 145`

`$ git pr show -t 145`

</details>

<a href="https://git.openjdk.org/crac/pull/145.diff">https://git.openjdk.org/crac/pull/145.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/145#issuecomment-1825358097)
</details>

### DIFF
--- a/src/java.base/share/classes/javax/crac/CheckpointException.java
+++ b/src/java.base/share/classes/javax/crac/CheckpointException.java
@@ -26,6 +26,11 @@
 
 package javax.crac;
 
+import jdk.crac.impl.ExceptionPrinter;
+
+import java.io.PrintStream;
+import java.io.PrintWriter;
+
 /**
  * This exception works as an aggregate for all errors found during checkpoint;
  * these are recorded as {@linkplain #getSuppressed() suppressed exceptions}.
@@ -39,5 +44,15 @@ public final class CheckpointException extends Exception {
      */
     public CheckpointException() {
         super(null, null, true, false);
+    }
+
+    @Override
+    public void printStackTrace(PrintStream s) {
+        ExceptionPrinter.print(this, s);
+    }
+
+    @Override
+    public void printStackTrace(PrintWriter w) {
+        ExceptionPrinter.print(this, w);
     }
 }
--- a/src/java.base/share/classes/javax/crac/CheckpointException.java
+++ b/src/java.base/share/classes/javax/crac/CheckpointException.java
@@ -26,33 +26,25 @@
 
 package javax.crac;
 
-import jdk.crac.impl.ExceptionPrinter;
-
-import java.io.PrintStream;
-import java.io.PrintWriter;
-
 /**
  * This exception works as an aggregate for all errors found during checkpoint;
- * these are recorded as {@linkplain #getSuppressed() suppressed exceptions}.
+ * these are recorded as {@linkplain #getNestedExceptions() nested exceptions}.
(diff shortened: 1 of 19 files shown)


rvansa:
@AntonKozlov Could you review please?

AntonKozlov:
Thanks. I agree that reporting exceptions as suppressed may be confusing. The new report looks good. But it introduces a new term "inner" which is not specified anywhere. Another side of the same problem is that programmatic access to the "inner" exception still remains through getSuppressed(), per javadoc and the fact there is no other mean to get an inner exception. I would suggest to go away completely from suppression (which was introduced for try-with-resources) to a completely distinct mechanism to record inner/nested/causing exceptions for Checkpoint... or RestoreException. How does it sound? My point is to have a mechanism that is consistent in 1) user-visible report; 2) in programmatic API; 3) the textual part of javadoc specification, e.g. package-level javadoc [1]. 

[1]https://github.com/openjdk/crac/blob/474b1638568a1a84b10d3b0a821e5d6e37f9eeae/src/java.base/share/classes/jdk/crac/package-info.java#L63

rvansa:
"Nested" is probably a better term than "inner". Having method `getNestedExceptions()` that actually calls `getSuppressed()` would be trivial, and would provide space for javadocs. A completely distinct field is probably cleaner, though; however it feels that we are getting more code duplicated for Checkpoint and RestoreException; do you think it would be useful to have them share a common parent? (ofc. we will still have some duplication with the `javax.crac` package...).

AntonKozlov:
I also would prefer "nested" exceptions not to overlap with suppressed in implementation, to make sure the correct interface is always called. Indeed, a common parent class makes sense, and sometime in handling of the exception I had to write
```
catch (CheckpointException | RestoreException e) { ... }
```

Which is probably correct way to indicate both checkpoint and restore failures are handled there. But if you want to just to print the "nested" exception, apprantly you'll need a common base exception. I.e. a common base looks good.

rvansa:
@AntonKozlov Updated, dropping all use of suppressed exceptions.

rvansa:
Still relevant.

rvansa:
Still relevant, waiting for @AntonKozlov review.

PR #147, opened 2023-12-08
Add Checkpoint timeout

tkiriyama:
Java process sometimes hangs when checkpoint for some reasons.
For example, this problems occurs if you specify certain options for CRAC_CRIU_OPTS.

```
# export CRAC_CRIU_OPTS=-V
# java -XX:CRaCCheckpointTo=/work/cp CRACTest
CR: Checkpoint ...
```
CRACTest process is not killed and is waiting for checkpoint.

```
# ls /work/cp
cppath  perfdata
```

To avoid this problem, I want to add the checkpoint timeout.
Can I submit a pull request to this repository? I would like you to review this change.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/147/head:pull/147` \
`$ git checkout pull/147`

`$ git checkout pull/147` \
`$ git pull https://git.openjdk.org/crac.git pull/147/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 147`

`$ git pr show -t 147`

</details>

<a href="https://git.openjdk.org/crac/pull/147.diff">https://git.openjdk.org/crac/pull/147.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/147#issuecomment-1846839302)

### DIFF
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -224,10 +224,23 @@ static int checkpoint_restore(int *shmid) {
   sigemptyset(&waitmask);
   sigaddset(&waitmask, RESTORE_SIGNAL);
 
+  struct timespec timeout;
+  timeout.tv_sec = CRaCCheckpointTimeout;
+  timeout.tv_nsec = 0;
+
   siginfo_t info;
   int sig;
   do {
-    sig = sigwaitinfo(&waitmask, &info);
+    if (timeout.tv_sec == 0) {
+      sig = sigwaitinfo(&waitmask, &info);
+    } else {
+      sig = sigtimedwait(&waitmask, &info, &timeout);
+      if (sig == -1 && errno == EAGAIN) {
+        tty->print_cr("The checkpoint was not generated within %d seconds.", CRaCCheckpointTimeout);
+        tty->print_cr("You can change the time-out period by -XX:CRaCCheckpointTimeout=.");
+        return JVM_CHECKPOINT_ERROR;
+      }
+    }
   } while (sig == -1 && errno == EINTR);
   assert(sig == RESTORE_SIGNAL, "got what requested");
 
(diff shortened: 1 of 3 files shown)


rvansa:
Hello @tkiriyama , could you clarify a bit more under what situation CRIU gets stuck? I guess that you've used `-V` just to demonstrate a situation where CRIU does not checkpoint the application as expected.

The checkpointed JVM waits indefinitely because CRIU exits with 0 - had it been an unsuccessful invocation the JVM would get the signal to stop waiting. What `criuengine.c` could do extra is to do one more check if the process still exists (when CRIU is done) and signal JVM if it suspects that CRIU didn't do the checkpoint. However there would be a race with a situation where it is immediately restored. Given that `CRAC_CRIU_OPTS` are meant to be expert options, I am not sure if we should protect against shooting yourselves into the foot this way.

I appreciate the effort of creating the PR, and kudos for the test, but I'd like to first hear about any real-world usecase where this timeout is useful.

tkiriyama:
Thank you for your reply. 
As you commented, there are cases in which the checkpointed JVM waits indefinitely. At that time, the JVM does not accept jstack and other tools, so it is difficult to know its status from the outside. For a typical checkpoint failure, criu returns non-zero and Java throws an exception. Java applications can handle checkpoint failures by handling exceptions. However, the reported case are difficult to handle in the application and I think it should be avoided.
Other than specifying CRAC_CRIU_OPTS, I have not found any cases where this problem occurs. I'll see if there are any risks with more practical use.

tkiriyama:
I haven't found any other options that cause a similar problem. The --action-scripts specification may have an impact, but I haven't been able to do it well.

CRAC_CRIU_OPTS was just an example. For example, if users specify their own program implementing with -XX: CREngine, they might see similar problem.
The purpose of this fix is to allow applications to take appropriate action when the JVM process goes into an infinite wait state for some unknown reason by throwing an exception instead.

rvansa:
OK, I am still not convinced it's too useful in practice, but the risk associated with the change is low, so I'll accept it. However, please set the default timeout to infinite rather than 10 minutes to not cause regression dumping big process onto slow storage.

tkiriyama:
OK. I changed the timeout to infinite so that it works the same as before by default.

rvansa:
Thanks. Could you enable Github Actions on your fork so that we see if this compiles and runs the test without errors?

tkiriyama:
I'm sorry. I forgot about it. I enabled Github Actions and run it now.

PR #148, opened 2023-12-10
8321660: [CRaC] Trim native heap before checkpoint to decrease image size

simonis:
Before checkpointing, we should trim the native C-heap in order to decrease the size of the generated CRaC image.

Without C-heap trimming, the size of a HelloWorld image is ~32m and the one for SpringPetclinic ~209m. If we trim the C-heap before checkpointin, the image size decreses to ~30m for HelloWorld and 170m for SpringPetclinic.

I've also added a new logging tag `crac` because I think it would be nice to log more CRaC details/internals in universal logging.  

Please let me know what you think?

---------
### Progress

### Issue
 * [JDK-8321660](https://bugs.openjdk.org/browse/JDK-8321660): [CRaC] Trim native heap before checkpoint to decrease image size (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/148/head:pull/148` \
`$ git checkout pull/148`

`$ git checkout pull/148` \
`$ git pull https://git.openjdk.org/crac.git pull/148/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 148`

`$ git pr show -t 148`

</details>

<a href="https://git.openjdk.org/crac/pull/148.diff">https://git.openjdk.org/crac/pull/148.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/148#issuecomment-1849049205)

### DIFF
--- a/src/hotspot/share/logging/logTag.hpp
+++ b/src/hotspot/share/logging/logTag.hpp
@@ -63,6 +63,7 @@ class outputStream;
   LOG_TAG(continuations) \
   LOG_TAG(coops) \
   LOG_TAG(cpu) \
+  LOG_TAG(crac) \
   LOG_TAG(cset) \
   LOG_TAG(data) \
   LOG_TAG(datacreation) \
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -430,6 +430,18 @@ Handle crac::checkpoint(jarray fd_arr, jobjectArray obj_arr, bool dry_run, jlong
   Universe::heap()->set_cleanup_unused(false);
   Universe::heap()->finish_collection();
 
+  if (os::can_trim_native_heap()) {
+    os::size_change_t sc;
+    if (os::trim_native_heap(&sc)) {
+      if (sc.after != SIZE_MAX) {
+        const size_t delta = sc.after < sc.before ? (sc.before - sc.after) : (sc.after - sc.before);
+        const char sign = sc.after < sc.before ? '-' : '+';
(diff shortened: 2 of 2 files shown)


simonis:
> Just a question: have you considered pros and cons trimming the heap inside the safepoint (VM_Crac)?

What would be the benefit of doing the trim inside the safepoint? Currently, native heap trimming is deliberately not being done at a safpoint (see [PR for JDK-8293114: JVM should trim the native heap](https://github.com/openjdk/jdk/pull/14781)). Now I agree that `VM_Crac` is a special safoint in the sense that it is not returning anyway and thus the time it takes is irrelevant.  But still, I don't think we should move tasks into a safepoint unless they functionally require to be executed at a safepoint. What do you think?

Apart from that, I don't think most test failures (e.g. tier1 runtime/serviceablitly) are related to this change because I couldn't reproduce them locally.

rvansa:
The point is that if you want to squeeze the most of trimming heap, you don't want either the VM operation itself or any racing threads to inflate the memory again before the checkpoint happens.
On the other hand, if the trimming is ignored at safepoint (I haven't studied the PR in detail, but some of the comments suggest that), it's a valid choice to do it before.

The test failures look OK, the tier1 failures should be hopefully fixed after merging more recent mainline.

simonis:
@rvansa, thanks for the review. I just want to clarify that trimming is possible at a safepoint - it is just automatic trimming with `-XX:+GCTrimNativeHeap` which is omitted there due to latency concerns.

I'll integrate this PR as is. We can still move it to the safepoint section later if we observe insufficient trimming, but honestly speaking, I don't think it'll make a difference.

PR #151, opened 2023-12-24
Add debug flag to -XX:CREngine.

yukikimmura:
Hello everyone,

This is my first PR for the CRAC project. I propose three modifications to the criuengine.

Firstly, I suggest adding a debug flag to the criuengine, as some JDK C/C++ native codes have the debug flag. 
My proposal is to pass the debug flag using "-XX:CREngine=criuengine,debug,true".

Secondly, the criuengine has a useful function, print_command_args_to_stderr(), which prints the criu command line. 
However, it is only called when an error occurs. I believe it would be beneficial if the criu command line is printed when the debug flag is specified.

The criu command line is built from the parameter of -XX:CREngine and environment variables such as CRAC_CRIU_PATH and CRAC_CRIU_OPT. 
It would be helpful to see how it is actually assembled.

For example:
 $ export CRAC_CRIU_PATH=/work/criu-crac-release-1.4/sbin/criu
 $ ./jdk/bin/java -XX:CREngine=/work/jdk/lib/criuengine,-v,3,-o,output3.log,-d,true -XX:CRaCCheckpointTo=cr Test

 Command: /work/criu-crac-release-1.4/sbin/criu dump -t 3232214 -D cr --shell-job '--verbosity=3' -o output3.log

Thirdly, I propose that the criuengine command line, which is executed by the JavaVM, be printed when "-XX:CREngine=criuengine,debug,true" is specified. 
For example, 
 CRaC info executing: /work/jdk/lib/criuengine checkpoint -v 3 -o output3.log -d true cr

Testing:
I have verified the jdk/crac/VMOptionsTest.java, which is a test for -XX:CREngine. I believe it's unnecessary to add a test for the debug flag to it.

Could you please review these modifications?

Thank you,
Kimura Yukihiro

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/151/head:pull/151` \
`$ git checkout pull/151`

`$ git checkout pull/151` \
`$ git pull https://git.openjdk.org/crac.git pull/151/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 151`

`$ git pr show -t 151`

</details>

<a href="https://git.openjdk.org/crac/pull/151.diff">https://git.openjdk.org/crac/pull/151.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/151#issuecomment-1868614490)

### DIFF
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -205,6 +205,14 @@ static int call_crengine() {
   }
   _crengine_args[1] = "checkpoint";
   add_crengine_arg(CRaCCheckpointTo);
+  if (CREngine &&
+      (strstr(CREngine, "-d,true") || strstr(CREngine, "--debug,true"))) {
+      tty->print("CRaC info executing: ");
+      for (unsigned int i=0; i < _crengine_argc; i++) {
+          tty->print("%s ", _crengine_args[i]);
+      }
+      tty->print_cr("");
+  }
   return os::exec_child_process_and_wait(_crengine, _crengine_args);
 }
 
--- a/src/java.base/unix/native/criuengine/criuengine.c
+++ b/src/java.base/unix/native/criuengine/criuengine.c
@@ -51,6 +51,7 @@ static int g_pid;
 
 static char *verbosity = NULL; // default differs for checkpoint and restore
 static char *log_file = NULL;
+static bool debug = false;
 
 static int kickjvm(pid_t jvm, int code) {
     union sigval sv = { .sival_int = code };
@@ -187,6 +188,9 @@ static int checkpoint(pid_t jvm,
         }
     }
     *arg++ = NULL;
+    if (debug) {
+        print_command_args_to_stderr(args);
+    }
 
     pid_t child = fork();
     if (!child) {
@@ -356,10 +360,15 @@ static char *parse_options(int argc, char *argv[]) {
         .has_arg = 1,
         .flag = NULL,
         .val = 'o',
+    }, {
+        .name = "debug",
+        .has_arg = 1,
+        .flag = NULL,
+        .val = 'd',
     }, { NULL, 0, NULL, 0} };
     bool processing = true;
     do {
-        switch (getopt_long(argc, argv, "v:o:", opts, NULL)) {
+        switch (getopt_long(argc, argv, "v:o:d:", opts, NULL)) {
             case -1:
             case '?':
                 processing = false;
@@ -372,6 +381,11 @@ static char *parse_options(int argc, char *argv[]) {
                 break;
             case 'o':
                 log_file = optarg;
+                break;
+            case 'd':
+                if (!strcmp(optarg, "true")) {
+                    debug = true;
+                }
                 break;
         }
     } while (processing);


simonis:
@yukikimmura, why do you need a new debug flag if you can pass any CRIU flags through `CRAC_CRIU_OPT`?

yukikimmura:
Hello simonis,

Thank you for your comment.
The debug flag I propose is purely for tracing and debugging the internal operation of the ciruengine.
I understand that CRAC_CRIU_OPTS is an environment variable for specifying options for criu.
Printing the ciru command sequence when the debug flag is set to true is one example.
However, the purpose of using the debug flag is not limited to this.

Thank you,
Kimura Yukihiro

yukikimmura:
To clarify, this is not a feature to pass debug flags to the criu command. 
It is for tracing or debugging criuengine.c. 
There may be an argument that criuengine is not so complex as to require debug flags. 
The log output by ciru --verbosity does not display the criu command sequence, 
so I would like a feature to print the criu command sequence when the debug flag is specified.

Sorry, I'm on vacation so I won't be able to reply for a while.

Thank you,
Kimura Yukihiro

rvansa:
In retrospective, I think that adding those extra parameters to CREngine was a mistake. I overgeneralized the need to pass some options to enable verbose logging (and TBH missed that underdocumented `CRAC_CRIU_OPTS` hack). Rather than passing something criuengine-specific, logging should have own option independent of CRIU, or even better, fetch settings from JVM unified logging.

Do you have many CRIU versions present in your system, or why are you interested in logging the exact paths? The parameters passed to CRIU don't change much, and CRIU in itself is an implementation detail. Why do you think it's useful to print it out regularly?

yukikimmura:
I assume that some vendors provide the CRAC JDK, and other end users use them. 
When the JDK vendors support CRaC, I thought they might want to know from the logs how the end users specified CREngine, CRAC_CRIU_OPTS, CRTOOLS_SCRIPT_ACTION, etc.

rvansa:
Maybe the `CREngine` should be guarded by the `UnlockDiagnosticVMOptions` flag; all these ways are very low-level 'expert' options.

AntonKozlov:
Printing out implicitly set command line parameters (e.g. via CRAC_CRIU_OPTS) may be indeed useful. Passing down unified logging settings is an interesting idea, although it won't be able to supply arbitrary options. But agree, two different ways to provide options to the engine is too much. CRAC_CRIU_OPTS from one hand is criuengine specific, but it allows easy concatenation with previously set options
```
CRAC_CRIU_OPTS="$CRAC_CRIU_OPTIONS"
```

Implementing the same with VM options is possible, but will take some effort. Have we considered introducing `CREngineArg` of `ccstrlist` type (accumulated by existing VM parsing implementation)? So for example `-XX:CREngineArg=debug` would a) accumulate with previously set arguments; b) will cover not only criuengine.

BTW, CREngine is not strictly `diagnostic` ("options are not meant for VM tuning or for product modes."). The wast majority of VM options are "expert" ones.

yukikimmura:
Thank you for your comment. 
I will consider it further based on your feedback.

Thank you,
Kimura Yukihiro

rvansa:
I meant diagnostic in the current state where all you can do is change output verbosity/destination. Had you added e.g. `--no-mmap-page-image` it would become more (and logging aside, this is really the only CRIU option that makes sense to change).

@yukikimmura I hope you're not offended by questioning the usefulness of your contribution; I simply feel that I made a mistake adding the CREngine params and therefore I am reluctant to build on top of that when I'd prefer to implement verbose logging in a different way.

yukikimmura:
I'm sorry for not being able to reply to your comment. I will continue to consider it.

Thank you,
Kimura Yukihiro

PR #153, opened 2024-03-18
8328344: [CRaC] Avoid error when running with -XX:+PerfDisableSharedMem

simonis:
CRaC tries to unconditionally restore the hsperf file, even if the checkpointed JVM process was started with `-XX:+PerfDisableSharedMem`.

The fix is trivial: check for the existence of the hsperf file in `PerfMemoryLinux::restore()` just as we already do it in `PerfMemoryLinux::checkpoint()`.

---------
### Progress

### Issue
 * [JDK-8328344](https://bugs.openjdk.org/browse/JDK-8328344): [CRaC] Avoid error when running with -XX:+PerfDisableSharedMem (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/153/head:pull/153` \
`$ git checkout pull/153`

`$ git checkout pull/153` \
`$ git pull https://git.openjdk.org/crac.git pull/153/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 153`

`$ git pr show -t 153`

</details>

<a href="https://git.openjdk.org/crac/pull/153.diff">https://git.openjdk.org/crac/pull/153.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/153#issuecomment-2003615152)

### DIFF
--- a/test/jdk/jdk/crac/PerfMemoryRestoreTest.java
+++ b/test/jdk/jdk/crac/PerfMemoryRestoreTest.java
@@ -31,7 +31,6 @@ import jdk.test.lib.process.OutputAnalyzer;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 
 import static jdk.test.lib.Asserts.*;
@@ -46,6 +45,7 @@ import static jdk.test.lib.Asserts.*;
  */
 
 public class PerfMemoryRestoreTest implements CracTest {
+    private static final int PERFDATA_CREATE_DELAY_MS = 100;
 
     @CracTestArg(0)
     boolean perfDisableSharedMem;
@@ -65,7 +65,7 @@ public class PerfMemoryRestoreTest implements CracTest {
         // in os::get_temp_directory() to /tmp rather than using System.getProperty("java.io.tmpdir")
         Path perfdata = Path.of("/tmp", "hsperfdata_" + System.getProperty("user.name"), pid);
         if (perfDisableSharedMem) {
-            Thread.sleep(100);
+            Thread.sleep(PERFDATA_CREATE_DELAY_MS);
             assertFalse(perfdata.toFile().exists(), "Perf data file exists although we run with -XX:+PerfDisableSharedMem");
         } else {
             waitForFile(perfdata);
@@ -83,7 +83,7 @@ public class PerfMemoryRestoreTest implements CracTest {
         // as restored.pid() would be the criuengine restorewait process
         String pidString = String.valueOf(checkpoint.pid());
         if (perfDisableSharedMem) {
-            Thread.sleep(100);
+            Thread.sleep(PERFDATA_CREATE_DELAY_MS);
             assertFalse(perfdata.toFile().exists(), "Perf data file exists although we run with -XX:+PerfDisableSharedMem");
         } else {
             waitForFile(perfdata);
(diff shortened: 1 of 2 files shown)


rvansa:
Thanks, do you think we should have a test for the flag?

simonis:
I've now added a test for `-XX:+/-PerfDisableSharedMem`, i.e. I've extended `PerfMemoryRestoreTest` to run with both, `-XX:-PerfDisableSharedMem` and `-XX:+PerfDisableSharedMem`.

test/jdk/jdk/crac/PerfMemoryRestoreTest.java
```diff
@@ -53,37 +66,61 @@ public void test() throws Exception {
         long start = System.nanoTime();
         while (!perfdata.toFile().exists()) {
             if (System.nanoTime() - start > TimeUnit.SECONDS.toNanos(10)) {
-                throw new IllegalStateException("Perf data file did not appear within time limit in the checkpointed process: " + perfdata);
+                if (perfDisableSharedMem) {
```
rvansa, line 69: The file won't ever appear, so this would busy loop for the whole 10 seconds, extending the testsuite. Could you avoid the loop completely, or in case of `perfDisableSharedMem == true` reduce the delay to, say, 100 ms?

simonis: If we expect that the file will appear we give it up to ten seconds to appear. If we want to be sure that it doesn't appear, we should then wait at least as long, right? Now for the positive case we expect that the file will appear quickly, so sleeping for a small amount of time like 10ms is fine. For the negative case, we already know that we will have to wait for the full ten seconds anyway, so we can just do that and not loop at all by sleeping 10 seconds.

I'll change the sleep time accordingly.

rvansa: As you say, in normal case the file should appear very quickly. The 10 second duration is there only to avoid false positives. Yes, if you want to assert that the file does not appear you would have to wait much longer than what is the 'expected' time.

I hope you don't take this as too much nitpicking; I am trying to balance the duration of the testsuite (esp. for manual runs) - not only in this case but also setting a precedent in style - vs. the risk of false negatives. I hate false positives as these reduce the credibility of the testsuite, but in my view asserting the non-existence of the file is not important to justify stalling of tests - I expect that if the VM option is completely ignored the tests should blow somewhere else.

simonis: Please feel free to take this PR and modify it however you think. Unfortunately I won't have bandwidth to work on it in the next time.

rvansa:
@AntonKozlov I've replaced the 2x 10 second wait with 2x 100 ms - could you review please to give it a third pair of eyes?

test/jdk/jdk/crac/PerfMemoryRestoreTest.java
```diff
@@ -38,52 +41,74 @@
...
         Path perfdata = Path.of("/tmp", "hsperfdata_" + System.getProperty("user.name"), pid);
-        long start = System.nanoTime();
-        while (!perfdata.toFile().exists()) {
-            if (System.nanoTime() - start > TimeUnit.SECONDS.toNanos(10)) {
-                throw new IllegalStateException("Perf data file did not appear within time limit in the checkpointed process: " + perfdata);
-            }
-            //noinspection BusyWait
-            Thread.sleep(10);
+        if (perfDisableSharedMem) {
+            Thread.sleep(100);
```
AntonKozlov, line 68: I like the refactoring and that we can control this delay independently. 

Nit-picking: should not be this a named constant? And does 100ms has any reason behind, or can it be 1sec (a little bit safer)?

rvansa: Extracted to constant.

My expert opinion says that the almost arbitrary choice of 100 ms is super good enough.

AntonKozlov (approved):
Thanks, LGTM

simonis:
Thanks @rvansa, @AntonKozlov for finishing this issue! I'll integrate it.

PR #155, opened 2024-06-04
Portable C/R

TimPushkin:
Implements a proof-of-concept "portable mode" for CRaC: a checkpoint-restore mechanism that does not rely on platform-dependent tools like CRIU instead saving VM state in terms of the Java specification (with some HotSpot specifics) — this allows to restore the saved state on machines with different CPU architecture and OS. A demo is available [here](https://github.com/TimPushkin/portable-crac-demo).

Expected downsides compared to the traditional CRaC are restrictions on platform-dependent code usage (e.g. at the moment of checkpoint no native methods can be executing, off-heap memory obtained via `sun.misc.Unsafe` should be released) and somewhat slower restoration speeds (because platform-dependent state, including JIT-compiled code, should be re-created). In the future, Project Leyden may help with the latter.

The mechanism is implemented as an internal part of HotSpot, it gets activated when an empty `CREngine` VM option is passed (i.e. `-XX:CREngine=""`, this is a temporary solution). Main implementation details are described in [this doc](https://github.com/TimPushkin/crac/blob/portable-cr/trimmed/doc/portable-cr.md).

Since this is a proof-of-concept implementation, it currently lacks some important features. E.g. at the moment some early-initialized classes are not restored, most of JDK classes have not yet been properly adapted, checkpointing via `jcmd` is not fully supported, additional tests and optimizations are needed.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/155/head:pull/155` \
`$ git checkout pull/155`

`$ git checkout pull/155` \
`$ git pull https://git.openjdk.org/crac.git pull/155/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 155`

`$ git pr show -t 155`

</details>

<a href="https://git.openjdk.org/crac/pull/155.diff">https://git.openjdk.org/crac/pull/155.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/crac/pull/155#issuecomment-2148011485)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,28 +1,5 @@
 # CRaC JDK
 
-## Build
-
-CRaC JDK have extended build procedure.
-
-1. Build JDK as usual
-```
-bash configure
-make images
-mv build/linux-x86_64-server-release/images/jdk/ .
-```
-2. Download a build of [modified CRIU](https://github.com/CRaC/criu/releases/tag/release-1.4)
-3. Extract and copy `criu` binary over a same named file in the JDK
-```
-cp criu-dist/sbin/criu jdk/lib/criu
-```
-Grant permissions to allow regular user to run it
-```
-sudo chown root:root jdk/lib/criu
-sudo chmod u+s jdk/lib/criu
-```
-
-# JDK
-
 For build instructions please see the
 [online documentation](https://openjdk.org/groups/build/doc/building.html),
 or either of these files:
@@ -33,3 +10,19 @@ or either of these files:
 See <https://openjdk.org/> for more information about the OpenJDK
 Community and the JDK and see <https://bugs.openjdk.org> for JDK issue
 tracking.
+
+## Building with CRIU
+
+To be able to use the default CRIU-based CRaC implementation follow these
+additional steps after building the JDK:
+
+1. Download a build of [modified CRIU](https://github.com/CRaC/criu/releases/tag/release-1.4)
+2. Extract and copy `criu` binary over a file with the same name in the JDK
+   ```
+   cp $CRIU_DIR/sbin/criu $JDK_DIR/lib/criu
+   ```
+3. Grant permissions to allow a regular user to run it
+   ```
+   sudo chown root:root $JDK_DIR/lib/criu
+   sudo chmod u+s $JDK_DIR/lib/criu
+   ```
--- a/doc/portable-cr.md
+++ b/doc/portable-cr.md
@@ -1,0 +1,121 @@
+# Portable checkpoint-restore algorithm
+
+This document describes the algorithm used by CRaC in the portable mode.
+
+On checkpoint, three dumps are created:
+
+1. Dump of Java threads' states — includes only user-created threads, while GC, compiler and other VM-created threads
+   are excluded since their state is not portable.
+    - If an included thread executes native code the checkpoint gets postponed in hopes that the thread will leave the
+      native code soon
+    - Some special cases (e.g. CRaC's own native code) are handled specifically and are allowed to be checkpointed
+2. Class dump — basically a list of serialized `InstanceKlass`es.
+3. Heap dump — currently an HPROF dump.
+
+The following sections describe the process of restoration from the above files. Text given in _italics_ relates only to
+the current proof-of-concept implementation and is not fundamental to the algorithm.
+
+## Metaspace and heap
+
+Restoration of class metadata and objects begins after the second phase of VM initialization because "only java.base
+classes can be loaded until phase 2 completes" while the classes being restored can come from arbitrary modules. At that
+point a lot of JDK's own Java code has been executed (e.g. foundational classes such as `java.lang.Class` have been
+loaded and initialized, initialization methods of `java.lang.System` have been executed and even CRaC's own
+`jdk.crac.Core` with the global context has been used), so many system classes already exist in a new state. And such
+re-initialization cannot be omitted since it is platform-dependent, e.g. system properties which are set up during the
+first initialization phase must be retrieved from the current system and not restored from the old one.
+
+_The current implementation even begins the restoration after the initialization fully completes as to not be bothered
+by stuff like proper system class loader and security manager initialization, so even some user code can get executed
+(from a user-provided system class loader, for instance), but this is a known temporary imperfection to be fixed later._
+
+### Defining the dumped classes
+
+First, class dump is parsed and the dumped classes are defined. The dump is structured in such a way that the classes
+can be defined as the dump is being parsed: super classes and interfaces of a class as well as its class loader's class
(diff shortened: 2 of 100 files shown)


rvansa:
Hello Tim,

this looks like a tremendous amount of work, despite some features being missed. Can you describe what is the ultimate motivation for the portable C/R from the practical point of view? CRaC if often perceived as a way to speed up application boot. As you have noted that the restore takes longer, portable C/R does not improve on that front. Also, valuable information from JIT has to be discarded as you claim portability across OSes and CPUs.

I see that you tested example-jetty as demo; can you share some ballpark performance figures that would compare CRIU-based startup times vs. Portable?

TimPushkin:
Hello Radim,

The ultimate motivation is to speed up application boot while keeping the level of portability expected from Java platform. As a specific practical example where this is useful, users will not have to update the image once shared libraries on which JVM depends get updated in the target environment — deployment of the image gets simpler.

Yes, this method will be slower than CRaC on CRIU (or other technologies operating on the level of OS processes) but it should be faster than restarting the program, at least in some scenarios. The greatest increase in speed is expected when on boot the app performs a lot of work in terms of Java: loading classes, making heavy use of reflection, performing computations and saving them in Java fields — the results of that work is in most cases platform-independent and thus will not have to be repeated on restore.

Regarding JIT, currently it is indeed discarded but, in theory, it can be added to the portable image in the future and used when the source and the target environments are similar enough. There are also hopes that Project Leyden, which also aims to reduce start up time, can be somehow integrated with portable CRaC, but since Leyden is currently mostly theoretical I cannot really elaborate on that (I imagine applying Leyden's condensers to the portable image after it has been created to speed it up for a specific target platform which can differ from the current platform).

As of now, performance of the proposed method is not impressive because my current goal is to develop the main parts of the method and make them work correctly. Many possible optimizations are currently out of focus (e.g. restoration is performed on a single thread since it is easier to implement this way, but some of its parts can be performed concurrently). On Jetty example (which I chose because it appeared to be the least complex to restore, other CRaC examples require some more work) I currently don't see improvements in time compared to starting the app anew, but, firstly, I expect to somewhat improve on that later making the code more optimal, and secondly, Jetty is probably not the best scenario to demonstrate performance of portable restoration (e.g. I expect to get better results with Spring since it heavily relies on reflection, but Spring examples are more complex and this proof-of-concept cannot handle them yet).

This PR is to show the work being done in this direction, I'm not sure about how and when this can be integrated in CRaC since currently it is more of a research which is not yet anywhere near being production ready.

rvansa:
Thanks you for the answers so far. If the PR is not yet ready for integration, would you mind marking it as 'Draft'? I think that such significant addition (resulting in some maintenance costs) would need a broader acceptance; probably worth introducing your ideas on the [crac-dev mailing list](https://mail.openjdk.org/mailman/listinfo/crac-dev).

I am curious about the problems you experience with Springboot - what aspects get harder to handle with your approach?

TimPushkin:
Thanks for the recommendations.

Regarding the problems encountered with Spring, these are not caused by the approach itself but by the fact that some features are yet unfinished. For example, states of some system classes used very early in the VM initiation process, like `java.lang.System` are not yet being restored — this does not cause issues for many programs including Jetty, but in general this is of course incorrect and breaks Spring in particular. This is the thing I'm planning to fix first, but wanted to share the work in the current state since a lot has been done already.

AntonKozlov:
Hi all,

I've participated as an advisor in this work, and I also think the R&D that @TimPushkin did is very impressive.

The portable, or Java-level, CRaC be used in many use-cases, like providing an implementation on non-linux platforms,  help even more with debugging CRaC applciation, or as a fallback if restore is not possible on the current platform (e.g. if CPU architectures between checkpoint and restore mismatch). But among those, I belive there are two notable cases.

First, it provides zero-effort java-level optimization framework. Many frameworks which target start-up optimizations opt to heavy build time optimizations to serialize their anticipated state to simple java programs recreating that state. Java-level CRaC can serialize that on the platform level. As with CRaC in general, the state is defined explicitly by Java program execution rather than more cumbersome generation of the java bytecode.

Another case, Java-level CRaC makes it easier to use Java tools for operations and transformations on the image. In the current state, the image format for the heap is .hprof, which makes it trivial to peek into the image and its state. And, such Java checkpoint can then be further optimized with AOT compilations, e.g. generation of the JIT code on the time of checkpoint (assuming multiple architectures), or more likely, feed java checkpoint to Leyden for the optimization(s).

The work looks great, but indeed it's too premature to be integrated to the main crac branch. I'm going to create `crac-portable` branch and integrate the PR there.

TimPushkin:
Still relevant

AntonKozlov:
Please retarget to `portable-cr` branch, and I'll integrate.

TimPushkin:
@AntonKozlov Done

AntonKozlov (approved):

headius:
I am excited to see this land!

As a test case, may I propose JRuby?

The JRuby startup story is very complex. We are an implementation of Ruby on the JVM, and since Ruby applications start from source we must boot up our pure-Java parser, compiler, and interpreter every time. Portions of JRuby itself are written in Ruby and must be parsed and loaded, and even the base "hello world" startup loads a large amount of Ruby code to bootstrap the standard library. On top of it all, we load thousands of classes at boot and generate hundreds more for invokedynamic method handle chains and early JIT-to-bytecode of early-loaded Ruby code.

Bypassing any of this would be a help for us.

I have had good results from using CRaC that bring us within a few times slower than the C implementation of Ruby (which is designed to start quickly), but the limitation of running only on Linux does not help the majority of Ruby developers that use MacOS.

Our other startup-time strategies are to disable all but tier 1 (-XX:TieredStopAtLevel=1, yay C1), or the new winner Leyden EA builds (which give us a 50% reduction in baseline startup).

I am interested in giving this a try, but someone else might get to it first! Running JRuby's baseline startup is easy:

* Download JRuby tarball or zip from [jruby.org](jruby.org)
* Run `bin/jruby -e 1`
* Bonus: run `bin/jruby -S gem list` which will list all preinstalled standard library "gems" and takes a couple of seconds on most systems.

Please feel free to reach out to me here or as @headius just about everywhere else!

TimPushkin:
@headius Glad to see your interest in the project! Portable CRaC should indeed help by reducing the time it takes JVM to load classes and process invokedynamics.

Currently the project is still in its early stages, it has been merged in a branch separate from CRaC's main one where it will be developed further. I am still working on making it work correctly in all scenarios, so I am not sure if JRuby, being a relatively complex project, will be handled correctly on the current development stage. But thanks for the suggestion, I will be trying JRuby on portable CRaC when I get back to the project.

PR #164, opened 2024-12-17
8346386: [CRaC] Move CRaC’s Java code out of restricted packages

TimPushkin:
To circumvent CRaC's limiting not-yet-standardized status:
1. Moved CRaC interface from `javax.crac` in `java.base` to `jdk.crac` in a new `jdk.crac` module.
2. Moved CRaC implementation from `jdk.crac` to `jdk.internal.crac.mirror`, both inside `java.base`. The whole CRaC implementation now resides in `jdk.internal.crac` which is exported (partially) only to the new `jdk.crac` module.

---------
### Progress

### Issue
 * [JDK-8346386](https://bugs.openjdk.org/browse/JDK-8346386): [CRaC] Move CRaC’s Java code out of restricted packages (**Task** - P4)

### Reviewers
 * [Radim Vansa](https://openjdk.org/census#rvansa) (@rvansa - Committer) ⚠️ Review applies to [29e0e41b](https://git.openjdk.org/crac/pull/164/files/29e0e41be5f7e445b0a12230c6b022d7c6c7f55b)

### Contributors
 * Timofei Pushkin `<pushkin.td@gmail.com>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/164/head:pull/164` \
`$ git checkout pull/164`

`$ git checkout pull/164` \
`$ git pull https://git.openjdk.org/crac.git pull/164/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 164`

`$ git pr show -t 164`

</details>

<a href="https://git.openjdk.org/crac/pull/164.diff">https://git.openjdk.org/crac/pull/164.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/164#issuecomment-2548361797)
</details>


src/java.base/share/classes/java/lang/System.java
```diff
@@ -833,9 +833,9 @@ public static native void arraycopy(Object src,  int  srcPos,
      * {@link #clearProperty(String)} may not have the desired effect.
      *
      * @crac System properties can be updated on restore from a checkpoint.
-     * The application can {@link javax.crac.Context#register(javax.crac.Resource) register}
+     * The application can {@link jdk.internal.crac.mirror.Context#register(jdk.internal.crac.mirror.Resource) register}
```
rvansa, line 836: The docs should use the `jdk.crac` package; `jdk.internal.crac.mirror` should not be used from application. Could you check that the generated javadocs will then link correctly to the different module?

TimPushkin: No, just swapping `jdk.internal.crac.mirror` with `jdk.crac` here does not work: javadoc complains that it cannot find the reference. I am not sure how to create inter-module links, there is `{@link java.logging/java.util.logging.Logger java.util.logging.Logger}` syntax in this file which successfully links to another module but using `jdk.crac/jdk.crac` does not work.

The current `jdk.internal.crac.mirror` also creates no link, probably because no docs are generated for this package. There is no error, just no hyperlink in the resulting HTML.

rvansa: > The current jdk.internal.crac.mirror also creates no link, probably because no docs are generated for this package. 

It's expected that there's no docs for `jdk.internal.crac.mirror`; it should not be exposed, neither from code nor from docs. If you can't get links to `jdk.crac` work correctly you could fallback to just `{@code ...}`. The correctness of the information for users has priority over the form.

TimPushkin: Added a fix, turned out no docs were being built at all for the new module. After fixing that just changing the package to `jdk.crac` starting working but I used the more verbose `jdk.crac/jdk.crac.Something` syntax anyway to be inline with the rest of inter-module links in the file.

src/hotspot/share/classfile/vmSymbols.hpp
```diff
@@ -744,7 +744,7 @@ class SerializeClosure;
   template(toFileURL_signature,                             "(Ljava/lang/String;)Ljava/net/URL;")                 \
   template(url_array_classloader_void_signature,            "([Ljava/net/URL;Ljava/lang/ClassLoader;)V")          \
                                                                                                                   \
-  template(jdk_crac_Core,                          "jdk/crac/Core")                                               \
+  template(jdk_crac_Core,                          "jdk/internal/crac/mirror/Core")                               \
```
rvansa, line 747: I think I've missed this one in the first review - shouldn't the first parameter be fixed as well?

src/java.base/share/classes/java/lang/System.java
TimPushkin: Sadly, this fix is not enough. The docs themselves build fine but when compiling there are warnings that the referenced module cannot be found. Investigating.

rvansa (approved):

TimPushkin:
41adb74 is a workaround to get `jdk/internal/jimage/JImageOpenTest.java` passing. It is not a full fix, a "more complete" fix (actually also a workaround) should come later as a separate change (it is not related to the package structure tweaking).

rvansa:
`jdk/crac/java/lang/System/TimedWaitingTest` is failing but it's unrelated to this PR; we've seen failures before, though I haven't figured out why could it be failing.

PR #167, opened 2024-12-23
8346766: [CRaC] tools/launcher/HelpFlagsTest.java and tools/launcher/VersionCheck.java tests fail on Windows

TimPushkin:
As a simple fix, CRaC's CREngine binaries are added as exceptions to the aforementioned tests.

In future the binaries can be made compliant with the tests (i.e. `-help`, `-version` and other required flags can be added).


rvansa (approved):
LGTM, thanks

PR #193, opened 2025-01-28
8348869: [CRaC] Restore args are split by whitespaces incorrectly

TimPushkin:
Fixes the issue by creating a new property `sun.java.crac_command` with spaces escaped. On restore, JVM parses this string de-escaping symbols.

Replaces abandoned #101.

---------
### Progress

### Issue
 * [JDK-8348869](https://bugs.openjdk.org/browse/JDK-8348869): [CRaC] Restore args are split by whitespaces incorrectly (**Bug** - P3)

### Reviewers

### Contributors
 * Roman Marchenko `<rmarchenko@openjdk.org>`
 * Timofei Pushkin `<tpushkin@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/193/head:pull/193` \
`$ git checkout pull/193`

`$ git checkout pull/193` \
`$ git pull https://git.openjdk.org/crac.git pull/193/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 193`

`$ git pr show -t 193`

</details>

<a href="https://git.openjdk.org/crac/pull/193.diff">https://git.openjdk.org/crac/pull/193.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/193#issuecomment-2619287610)
</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
+++ b/src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
@@ -240,22 +240,20 @@ public class Core {
             // Parse arguments into array, unescape spaces
             final char escChar = '\\';
             final char sepChar = ' ';
-            ArrayList<String> argList = new ArrayList<String>();
-            argList.add("");
+            final var argList = new ArrayList<String>();
+            final var curArgBuilder = new StringBuilder();
             for (int i = 0; i < newArguments.length(); ++i) {
                 final char curChar = newArguments.charAt(i);
-                final int lastIdx = argList.size() - 1;
                 switch (curChar) {
-                    case sepChar:
-                        argList.add("");
-                        break;
-                    case escChar:
-                        argList.set(lastIdx, argList.get(lastIdx) + newArguments.charAt(++i));
-                        break;
-                    default:
-                        argList.set(lastIdx, argList.get(lastIdx) + curChar);
+                    case sepChar -> {
+                        argList.add(curArgBuilder.toString());
+                        curArgBuilder.setLength(0);
+                    }
+                    case escChar -> curArgBuilder.append(newArguments.charAt(++i));
+                    default -> curArgBuilder.append(curChar);
                 }
             }
+            argList.add(curArgBuilder.toString());
 
             final String[] args = argList.toArray(new String[argList.size()]);
             if (args.length > 0) {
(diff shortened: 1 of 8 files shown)


src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
```diff
@@ -236,7 +237,27 @@ public String run() {
         }
 
         if (newArguments != null && newArguments.length() > 0) {
-            String[] args = newArguments.split(" ");
+            // Parse arguments into array, unescape spaces
+            final char escChar = '\\';
+            final char sepChar = ' ';
+            ArrayList<String> argList = new ArrayList<String>();
```
rvansa, line 243: Could you please rewrite this using `ArrayList<StringBuilder>`?

TimPushkin: Done

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
```diff
@@ -236,7 +237,25 @@ public String run() {
...
+            final var argList = new ArrayList<String>();
+            final var curArgBuilder = new StringBuilder();
+            for (int i = 0; i < newArguments.length(); ++i) {
+                final char curChar = newArguments.charAt(i);
+                switch (curChar) {
+                    case sepChar -> {
+                        argList.add(curArgBuilder.toString());
+                        curArgBuilder.setLength(0);
+                    }
+                    case escChar -> curArgBuilder.append(newArguments.charAt(++i));
```
rvansa, line 252: This version looks much better, but I wonder what happens here if the escape is the last char - this code looks like it would throw OOBE (as there's no check after incrementation). It seems that situation is covered in the test case as well - I don't understand why it won't fail.

TimPushkin: It does not fail because this code receives not the original arguments string but the escaped one constructed [here](https://github.com/openjdk/crac/pull/193/files/cadaa8bb86404d883885324b6a95be2105e0c1fb#diff-c3caaf3e347b1a477e2b7278cb6a35da04a597de5632c13351a6e4e5a2924d21R1869). For example, if the original string was "\" it will be "\\" here, the code will parse the first slash and skip the second one. And since the code constructing the escaped string is ours, unless there is will be a bug there, the escaped string should never end with a single slash.

I thought about this too and considered placing an assert here (or even an explicit check) but since the single-slash situation should happen only if there is a bug in our code I decided that the default OOBE will be enough for us to identify the issue.

rvansa: Thanks for the explanation, I think it is fine then.

rvansa (approved):

PR #202, opened 2025-02-06
JDK-8349560: [CRaC] Close EPoll FDs when FD policies close registered sockets

rvansa:
FD policies are applied as the last set of JDK resources. When these are meant to close server sockets, it’s rather too late: open EPoll FDs won’t be closed automatically as these have the socket still registered when the automatic close is performed.

We can apply FD policies on sockets and close these if requested before EPoll FD handling. The downside is that we would not handle sockets opened during checkpoint, but generally the application should not do that. 

Besides changing the order of closing this PR also improves the way EPoll FDs are reported: the Selector throws BusySelectorException listing channels registered on that selector, and EPoll FD and Event Epoll FD are claimed through Java code with message referencing the selector (this suppresses native open FD detection that wouldn't provide much info).

### DIFF
--- a/src/java.base/linux/classes/sun/nio/ch/EPollSelectorImpl.java
+++ b/src/java.base/linux/classes/sun/nio/ch/EPollSelectorImpl.java
@@ -32,19 +32,20 @@ import jdk.internal.access.JavaIOFileDescriptorAccess;
 import jdk.internal.access.SharedSecrets;
 import jdk.internal.crac.Core;
 import jdk.internal.crac.JDKResource;
+import jdk.internal.crac.mirror.impl.CheckpointOpenResourceException;
+import jdk.internal.crac.mirror.impl.CheckpointOpenSocketException;
 
 import java.io.FileDescriptor;
 import java.io.IOException;
-import java.nio.channels.IllegalSelectorException;
+import java.io.Serial;
+import java.nio.channels.SelectableChannel;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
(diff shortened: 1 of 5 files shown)


src/java.base/linux/classes/sun/nio/ch/EPollSelectorImpl.java
```diff
@@ -425,11 +428,24 @@ public void beforeCheckpoint(Context<? extends Resource> context) throws Excepti
...
+                ex.epollFds.add(claimFd(access, this.epfd, "EPoll FD "));
+                ex.epollFds.add(claimFd(access, this.eventfd.efd(), "EPoll Event FD "));
+                currentChannels = null;
+                throw ex;
             }
         }
     }
 
+    private FileDescriptor claimFd(JavaIOFileDescriptorAccess access, int fdval, String type) {
+        FileDescriptor fd = new FileDescriptor();
```
rvansa, line 442: Maybe I could use `IOUtil.newFD(fdval)` here...

TimPushkin (approved):

TimPushkin:
Just noticed that `fileDescriptors/PollerTest` failed on Windows. That's a time out but can be related?

rvansa:
I've tried some rebuilds, sometimes failing on jtreg fetch but now we have a second failure in `PollerTest` on Windows. I don't see why it could be related but I'll need to check it out in depth.

rvansa:
@TimPushkin Could you please check and re-approve the last changes?

TimPushkin (approved):
LGTM

PR #207, opened 2025-02-27
8350845: [CRaC] Support C/R engines in form of a library

TimPushkin:
Adds support for C/R engines implemented in form of dynamic libraries.

In the proposed API the engine controls the set of the configurable options — they are mainly passed by the user via the new `CRaCEngineOptions` VM option, but some may also be passed by the VM. Other VM and/or engine implementations may extend the API via "extensions".

The JVM is supposed to know what configuration options and API extensions are supported by calling `can_configure` and `get_extension`.

The user is supposed to know which options they can pass by asking the engine with `-XX:CRaCEngineOptions=help` (support for this is engine-dependent). Example of how this looks for bundled engines:
```
$ java -XX:CRaCEngine=simengine -XX:CRaCEngineOptions=help -XX:CRaCCheckpointTo=cr -version

crexec — pseudo-CRaC-engine used to relay data from JVM to a "real" engine implemented as an executable (instead of a library).
The engine executable is expected to have CRaC-CRIU-like CLI. Support of the options below also depends on the engine executable.

Configuration options:
* keep_running=<true/false> (default: false) — keep the process running after the checkpoint or kill it.
* direct_map=<true/false> (default: false) — on restore, map process data directly from saved files. This may speedup the restore but the resulting process will not be the same as before the checkpoint.
* args=<string> (default: "") — free space-separated arguments passed directly to the engine executable, e.g. "--arg1 --arg2 --arg3".
* help — print this message.

openjdk version "24-internal" 2025-03-18
OpenJDK Runtime Environment (slowdebug build 24-internal-adhoc.timpushkin.crac)
OpenJDK 64-Bit Server VM (slowdebug build 24-internal-adhoc.timpushkin.crac, mixed mode, sharing)
```

Notable related changes included in the patch:
- The ability to pass arguments to C/R engine was removed from `CRaCEngine` because it can now be implemented through `CRaCEngineOptions`. E.g. `-XX:CRaCEngine=criu,--verbose` is now `-XX:CRaCEngine=criu -XX:CRaCEngineOptions=args=--verbose`.
- `CRaCEngine` (as well as the new `CRaCEngineOptions`) VM Option is now not updated in the restored VM. This seems more correct because engine path and options will very likely differ on restore and on checkpoint, so it is not obvious how to combine them — just replacing won't always work. Until now the updates were ignored by the CRaC implementation anyway, so this should only be visible by inspecting the options through JMX.
- `CRaCRestoreFrom` is now considered set (and thus an attempt to restore is made) only if it is set to a non-empty value, i.e. `-XX:CRaCRestoreFrom=""` won't trigger a (failing) restore attempt anymore. This is to be consistent with how `-XX:CRaCCheckpointTo=""` does not trigger a checkpoint setup.


src/hotspot/share/include/crlib/crlib.h
```diff
@@ -0,0 +1,104 @@
...
+  bool (*can_configure)(crlib_conf_t *, const char *key);
+  // Sets a configuration option. Returns true on success.
+  // Key and value are valid C-strings.
+  bool (*configure)(crlib_conf_t *, const char *key, const char *value);
+
+  // Returns an API extension with the given name (C-string) and size, or null if an extension with
+  // such name is not present or its size is lower than requested.
+  // The extension should have static storage duration. The application is supposed to cast it to
+  // the actual extension type.
+  const crlib_extension_t *(*get_extension)(const char *name, size_t size);
```
TimPushkin, line 74: There is no way to get the full list of extensions and configuration options because I don't see why it should be needed. VM only needs to know if a particular extension/option is supported — for that it can call `get_extension`/`can_configure`.

rvansa:
I like the idea of `-XX:CRaCEngineOptions=help` on the UX side, though it would be even better to have it working without `-XX:CRaCCheckpointTo=...`. However I don't think that `configure(...)` method on the API should cause any side-effects, like printing output. When this is already an *encouraged* - in other words optional but standard - way to provide output, it should be a standalone method; I don't have strong opinion about whether it would be `const char *help()` or `void print_help(int fd)` or anything else, but `configure` should only modify `conf`.

As for the extensions listing, I don't have a clear use case in mind but some way of identification in a management interface; in general it is nice if a pluggable component can self-identify (name, version, ...). `help` provides this in a non-structured way, which might be sufficient, though it is not complex at all to implement few functions returning some self-identifying strings.

rvansa:
> How about adding an optional extension for this, not sure how exactly to call it (`description`, `identification`, `docs`, `reflection`...), but it would have methods `const char *id_string()` and `const char *configure_description_string()`. Maybe also `const char *extensions_description_string()` but I'm still not sure why this would be needed.

Yes, extension is a sensible way to do this if you don't want to make this obligatory part of the interface.

> Then, if JVM detects that the user asked for help, it could construct the help message by combining the output of these two methods to get roughly the same as in the PR description. To make it so that it does not require `CRaCCheckpointTo`/`CRaCRestoreFrom` to work we can either make `CRaCEngineOptions=help` a special case for which the VM will check regardless of the CRaC-enabling options being set, or we can add another VM option `CRaCPrintEngineHelpAndExit` (or `PrintCRaCEngineHelpAndExit`, `PrintCRaCEngineInfoAndExit`?) that would load the engine, print the help message and exit VM analogous to how `PrintSharedArchiveAndExit` from CDS behaves.

I would stick to `CRaCEngineOptions` (and a special case). Bonus points if you don't require any other option.

src/hotspot/share/include/crlib/crlib.h
```diff
@@ -0,0 +1,104 @@
...
+  int (*checkpoint)(crlib_conf_t *);
+  // Triggers a restore. Does not normally return, but if it does returns a error code.
+  int (*restore)(crlib_conf_t *);
+
+  // Returns true if the given configuration key is supported by the engine, false otherwise.
+  // Key is a valid C-string.
+  // Use of this before configuring is not a requirement.
+  bool (*can_configure)(crlib_conf_t *, const char *key);
+  // Sets a configuration option. Returns true on success.
+  // Key and value are valid C-strings.
```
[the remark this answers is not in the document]
TimPushkin: This is documented in the description of the whole API structure [here](https://github.com/openjdk/crac/pull/207/files/4076f28e380ec96fabd9ab218f21c7bb5d58ffb3#diff-2dae7c218d19b9a6accbd269da4b1944a7bcc94c6a9f54e6b84b82d247325d6fR49-R50)

src/hotspot/share/include/crlib/crlib_restore_data.h
```diff
@@ -0,0 +1,57 @@
...
+#define CRLIB_EXTENSION_RESTORE_DATA(api) \
+  CRLIB_EXTENSION(api, crlib_restore_data_t, CRLIB_EXTENSION_RESTORE_DATA_NAME)
+
+// API for passing data from a restoring application to a restored application.
+struct crlib_restore_data {
+  crlib_extension_t header;
+
+  // Called by the restoring application to pass data to the restored application.
+  // 'data' must not be null, size must be greater than 0.
+  // The engine may impose limits on the data size and return false if it is not accepted.
```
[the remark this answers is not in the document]
TimPushkin: Yes, again, I decided to write this in a [single place](https://github.com/openjdk/crac/pull/207/files/4076f28e380ec96fabd9ab218f21c7bb5d58ffb3#diff-2dae7c218d19b9a6accbd269da4b1944a7bcc94c6a9f54e6b84b82d247325d6fR49-R50) for all of the API methods

test/jdk/jdk/crac/CracEngineOptionsTest.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    private void test(String engine, String opts, int expectedExitValue, String... expectedTexts) throws Exception {
+        ProcessBuilder pb = ProcessTools.createTestJavaProcessBuilder(
+                "-XX:CRaCCheckpointTo=cr",
+                "-XX:CRaCEngine=" + engine,
+                "-Xlog:crac=debug",
+                "-version");
+        if (opts != null) {
+            pb.command().add(pb.command().size() - 2, "-XX:CRaCEngineOptions=" + opts);
```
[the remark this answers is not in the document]
TimPushkin: I actually missed this and `\n` is not treated as a separator at all. Will fix and add a test case, thanks.

src/java.base/share/native/libcrexec/crexec.c
```diff
@@ -0,0 +1,527 @@
...
+void exec_in_this_process(const char *path, const char *argv[], const char *env[]);
+
+typedef bool (*configure_func)(crlib_conf_t *, const char *value);
+
+static bool configure_help(crlib_conf_t *conf, const char *ignored) {
+  // Internal options which are expected to be set by the program crexec is linked against are
+  // omitted from the print since users are not supposed to pass them directly:
+  // * image_location=<path> — path to a directory with checkpoint/restore files.
+  // * exec_location=<path> — path to the engine executable.
+  const int ret = printf(
```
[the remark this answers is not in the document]
rvansa: Would it be a problem to turn this into C++?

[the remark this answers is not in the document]
TimPushkin: I've converted crexec in C++ but left these as usual strings because with raw string literals the long lines are harder to read in code editors (in the terminal they are wrapped automatically)

src/hotspot/share/include/crlib/crlib.h
TimPushkin: Now there is a way to do this via an API extension

TimPushkin:
Updated `-XX:CRaCEngineOptions=help` usage example in PR's description

src/hotspot/share/runtime/crac.cpp
```diff
@@ -521,45 +476,62 @@ void crac::prepare_restore(crac_restore_data& restore_data) {
...
-    fprintf(stderr, "-XX:CRaCRestoreFrom parameter is not a directory: %s\n", CRaCRestoreFrom);
+    log_error(crac)("CRaCRestoreFrom=%s is not a directory", CRaCRestoreFrom);
     return;
   }
 
-  compute_crengine();
-
-  const int id = os::current_process_id();
+  // Note that this is a local, i.e. the handle will be destroyed if we fail to restore
+  auto engine = CracEngine(CRaCRestoreFrom);
```
rvansa, line 492: A nitpick/question: why do you use this syntax that does an extra copy rather than
```
CracEngine engine(CRaCRestoreFrom);
```
?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1960,11 +1960,18 @@ const int ObjectAlignmentInBytes = 8;
       "Reset JVM's start time and uptime on restore")                       \
                                                                             \
   product_pd(ccstr, CRaCEngine, RESTORE_SETTABLE,                           \
-      "Path or name of a program implementing checkpoint/restore and "      \
-      "optional extra parameters as a comma-separated list: "               \
-      "-XX:CRaCEngine=program,--key,value,--anotherkey results in calling " \
-      "'program --key value --anotherkey'. Commas used as part of args "    \
-      "should be escaped with a backslash character ('\\').")               \
+      "Path or name of a program or a shared library implementing "         \
+      "checkpoint and restore. On restore this option applies only to "     \
```
rvansa, line 1964: I would even say that the CREngine must match on checkpoint and restore.

test/jdk/jdk/crac/CracEngineOptionsTest.java
```diff
@@ -0,0 +1,180 @@
...
+    @Test
+    public void test_options() throws Exception {
+        test("simengine", "");
+        test("simengine", "image_location=cr", 0,
+                "Internal CRaC engine option provided, skipping: image_location");
+        if (Platform.isLinux()) {
+            test("criuengine", "keep_running=true,args=-v -v -v -v,keep_running=false", 0,
+                    "CRaC engine option: 'keep_running' = 'true'",
+                    "CRaC engine option: 'args' = '-v -v -v -v'",
+                    "CRaC engine option: 'keep_running' = 'false'");
```
rvansa, line 90: I think that providing the same option twice should result in warning, at least.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -375,31 +285,73 @@ void VM_Crac::doit() {
...
+      return;
+    }
+    vm_exit(0);
+    ShouldNotReachHere();
+  }
+  postcond(status == CracEngine::ApiStatus::OK);
+
+  const char *description = engine.description();
+  if (description == nullptr) {
+    log_error(crac)("CRaC engine failed to provide its textual description");
```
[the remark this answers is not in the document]
TimPushkin: What I meant is: the API doc says "Returns ... null on error" => if `description == nullptr` then there has been an error in the engine (e.g. it decided to malloc the string and failed) => log the error and exit.

src/hotspot/share/runtime/crac.cpp
TimPushkin: This is a mistake, thanks for catching

src/hotspot/share/runtime/crac.cpp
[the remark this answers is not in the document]
rvansa: OK, it can easily return `""` when it wants to stay enigmatic.

rvansa:
LGTM, thank you!

rvansa (approved):

test/jdk/jdk/crac/CracEngineOptionsTest.java
rvansa: The engine is not interpreting JVM options; engine receives its programmatic configuration and has to deal with that. JVM defines the user API and we can decide that repeated use of the same argument is prohibited (as that would be more likely an user error).

TimPushkin: I mean that an engine can interpret reconfiguration of the same key not as overwriting of the previous value but somehow differently, e.g. as concatenation in my example. Since the VM is just passing the keys through I would leave the interpretation up to the engine. But it's not a strong opinion, I can move the warning to JVM if you still think it would be better.

TimPushkin: After thinking a bit more, yes, let's move the warning to JVM to simplify the engine implementation. We can always remove the warning if we find a sensible use case for repeating keys.

TimPushkin:
`gc/shenandoah/mxbeans/TestPauseNotifications#generational` failure is a known issue (JDK-8345895) — can be ignored. Windows build failures are strange, seem related to JDK-8275405 but not quite — I need some time to investigate how to fix or workaround this.

TimPushkin:
Regarding [the problem with Windows debug builds](https://github.com/TimPushkin/crac/actions/runs/13674162120/job/38230839206#step:10:2700), it looks like Windows debug build [tries to export all symbols](https://github.com/openjdk/crac/blob/884d0746b168550f13bdc687b1d96d468aec4411/make/hotspot/lib/CompileJvm.gmk#L235) which don't have "type_info" or "lambda" in their name regardless of whether they have external or internal linkage. Not sure for what purpose (for testing/debugging?).

In my particular case, if the hashing/equality functions I've introduced have internal linkage, the hashtable template instantiation has internal linkage too and linker fails to export it. If the functions have external linkage, the template instantiation has external linkage too and everything is fine. This is the workaround I've chosen (also put them in a class just to have a "name space" for them).

Alternatively, if I add "type_info" or "lambda" to the name of any of the functions the resulting hashtable template instantiation symbol will have this substring in its name and thus it will be filtered from the export list by the build system, the error will not occur (and the symbols in question will have internal linkage). This seems fragile so I've decided against it.

In release builds this doesn't happen probably because [the symbol table used to construct the export list](https://learn.microsoft.com/en-us/cpp/build/reference/symbols) is only filled for debug builds.

rvansa:
I think that the solution you've chosen won't hurt; just a few extra methods exported.

PR #210, opened 2025-03-11
8351695: [CRaC] Cannot restore immediately without default restore data

TimPushkin:
The issue was not caught by the test suite because `crexec` always provides restore data even if none was set.

---------
### Progress

### Issue
 * [JDK-8351695](https://bugs.openjdk.org/browse/JDK-8351695): [CRaC] Cannot restore immediately without default restore data (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/210/head:pull/210` \
`$ git checkout pull/210`

`$ git checkout pull/210` \
`$ git pull https://git.openjdk.org/crac.git pull/210/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 210`

`$ git pr show -t 210`

</details>

<a href="https://git.openjdk.org/crac/pull/210.diff">https://git.openjdk.org/crac/pull/210.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/210#issuecomment-2715183994)
</details>

### DIFF
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -125,6 +125,10 @@ int crac::checkpoint_restore(int *shmid) {
     case CracEngine::ApiStatus::OK: {
       constexpr size_t required_size = sizeof(*shmid);
       const size_t available_size = _engine->get_restore_data(shmid, sizeof(*shmid));
+      if (available_size == 0) { // Possible if we were not killed by the engine and thus there is no restoring JVM
+        *shmid = 0; // Not an error, just no restore data
+        break;
+      }
       if (available_size == required_size) {
         break;
       }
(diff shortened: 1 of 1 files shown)


TimPushkin:
We can also make `crexec` not provide restore data unless it was previously set to be able to regression-test this. But I am not sure if this is needed.

rvansa:
Do I understand that this happens with `CRAC_CRIU_LEAVE_RUNNING`? If we are missing a test for that (I suppose rather using `-XX:CRaCEngineOpts=keep_running=true`, could you please add one?

TimPushkin:
@rvansa There is a test for keep running/leave running (`jdk/crack/LeaveRunning`). But since crexec always returns some restore data, even if it hasn't been set (in that case it returns zero), the issue doesn't appear. So, to get a test for this not a test should be added but the behavior of crexec should be changed.

rvansa:
OK, if this is handling a hypothetic case that doesn't happen with `crexec` I am fine with the change.

rvansa (approved):

PR #211, opened 2025-03-11
8351701: [CRaC] CRaCEngineOptions=help cannot be used with CRaCRestoreFrom

TimPushkin:
Fixes the issue and adds a regression test for it.

---------
### Progress

### Issue
 * [JDK-8351701](https://bugs.openjdk.org/browse/JDK-8351701): [CRaC] CRaCEngineOptions=help cannot be used with CRaCRestoreFrom (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/211/head:pull/211` \
`$ git checkout pull/211`

`$ git checkout pull/211` \
`$ git pull https://git.openjdk.org/crac.git pull/211/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 211`

`$ git pr show -t 211`

</details>

<a href="https://git.openjdk.org/crac/pull/211.diff">https://git.openjdk.org/crac/pull/211.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/211#issuecomment-2715373622)
</details>

### DIFF
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -2261,6 +2261,11 @@ bool Arguments::parse_options_for_restore(const JavaVMInitArgs* args) {
 
   postcond(CRaCRestoreFrom != nullptr);
 
+  if (CRaCEngineOptions && strcmp(CRaCEngineOptions, "help") == 0) {
+    crac::print_engine_info_and_exit(); // Does not return on success
+    return false;
+  }
+
   return true;
 }
 
--- a/test/jdk/jdk/crac/CracEngineOptionsTest.java
+++ b/test/jdk/jdk/crac/CracEngineOptionsTest.java
@@ -157,13 +157,9 @@ public class CracEngineOptionsTest {
 
     @Test
     public void test_options_help() throws Exception {
-        ProcessBuilder pb = ProcessTools.createTestJavaProcessBuilder(
-                "-XX:CRaCEngineOptions=help");
-        OutputAnalyzer out = new OutputAnalyzer(pb.start());
-        out.shouldHaveExitValue(0);
-        out.stdoutShouldContain("Configuration options:");
-        out.stderrShouldBeEmpty();
-        out.shouldNotContain("CRaC engine option:");
+        testHelp();
+        testHelp("-XX:CRaCCheckpointTo=cr");
+        testHelp("-XX:CRaCRestoreFrom=cr");
     }
 
     private void test(String engine) throws Exception {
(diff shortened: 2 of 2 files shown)


src/hotspot/share/runtime/arguments.cpp
```diff
@@ -2261,6 +2261,11 @@ bool Arguments::parse_options_for_restore(const JavaVMInitArgs* args) {
 
   postcond(CRaCRestoreFrom != nullptr);
 
+  if (CRaCEngineOptions && strcmp(CRaCEngineOptions, "help") == 0) {
```
rvansa, line 2264: Since you compare the full value, rather than one option... What will
```
java -XX:CRaCEngineOptions=help -XXCRaCEngineOptions=foo.bar ...
```
do, and to we have a test for that?

TimPushkin: `help` will be treated as just another engine options and passed through to the engine, which in case of crexec will lead to `crexec: unknown configure option: help` error. Using `-XX:CRaCEngineOptions=help,args=--arg` or `-XX:CRaCEngineOptions=help=true` will do the same. Both on checkpoint and on restore. There were tests for this but I removed most of them as excessive, brought them back now.

This is intentional, I made it this way for simplicity: would need to parse options and compare all their names to `help` otherwise — and I don't think it would bring much UX benefit. By saying "use '-XX:CRaCEngineOptions=help' to ..." in the VM option docs I meant exactly that: that "help" should be specified as the only option — this can be stated more explicitly but that would be more wordy and I think it is understandable enough as is.

TimPushkin: We can replace the above with a separate `-XX:CRaCPrintEngineHelpAndExit` VM option if the behavior above seems too unintuitive

TimPushkin:
`[jdk/crac/fileDescriptors/ReopenListeningAsynchronousSocketChannelTest](https://github.com/TimPushkin/crac/actions/runs/13795590141/attempts/1#user-content-jdk_crac_filedescriptors_reopenlisteningasynchronoussocketchanneltest)` failed once (judging by the exit code, there was an access violation error) but it was fine after re-run, I couldn't reproduce that locally (run the test 10 times) and the change seems unrelated to the test.

rvansa (approved):

src/hotspot/share/runtime/arguments.cpp
rvansa: I am fine with the existing state, was just wondering whether it will both print help and error, or just error out somewhere else... Looks like there won't be a surprise so all's good.

PR #212, opened 2025-03-13
8351923: [CRaC] Engine help message may be truncated

TimPushkin:
Fixes the help message truncation by using `tty->print_raw*` instead of `tty->print*`. The latter uses a fixed-size buffer and truncates strings which don't fit there, the former does not.


rvansa:
LGTM, thanks!

PR #213, opened 2025-03-19
JDK-8352394: [CRaC] AOT class linking cannot archive jdk.internal.crac.Core$Priority

TimPushkin:
Fixes (or rather works around) the incompatibility with AOT class linking by converting `jdk.internal.crac.Core$Priority` into a usual class. We can convert it back to enum when CDS either starts supporting archiving `Reference` objects or stops aborting the archiving when it stumbles upon an unsupported object.

An alternative approach would be to make CDS treat this particular enum class specially and don't archive it in initialized state but messing with CDS internals can lead to many conflicts later.

This is a draft because after implementing I noticed [JDK-8344140](https://github.com/openjdk/jdk/pull/22291) which restricted the set of AOT-initialized enums — maybe `jdk.internal.crac.Core$Priority` is not in the restricted set. We should get JDK-8344140 integrated as part of jdk 25+6 and see if the issue resolves itself.

### DIFF
--- a/src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
+++ b/src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
@@ -176,18 +176,9 @@ public class BasicImageReader implements AutoCloseable {
             return;
         }
         try {
-            Object[] priorities = priorityClass.getEnumConstants();
-            if (priorities == null) {
-                return;
-            }
-            Object normalPriority = null;
-            for (int i = 0; i < priorities.length; ++i) {
-                if ("NORMAL".equals(priorities[i].toString())) {
-                    normalPriority = priorities[i];
-                }
-            }
+            Object normalPriority = priorityClass.getDeclaredField("NORMAL").get(null);
             if (normalPriority == null) {
-                throw new IllegalStateException();
+                throw new IllegalStateException("CRaC resource priority NORMAL is null");
             }
             Class<?> resourceClass = Class.forName("jdk.internal.crac.mirror.Resource");
             Method getContextMethod = priorityClass.getMethod("getContext");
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/jdk/internal/jimage/BasicImageReader.java
```diff
@@ -176,18 +176,9 @@ private void registerIfCracPresent() {
...
-            Object[] priorities = priorityClass.getEnumConstants();
-            if (priorities == null) {
-                return;
-            }
-            Object normalPriority = null;
-            for (int i = 0; i < priorities.length; ++i) {
-                if ("NORMAL".equals(priorities[i].toString())) {
-                    normalPriority = priorities[i];
-                }
-            }
```
TimPushkin, line 188: I'm not sure, maybe this piece of code should be retained as an alternative for compatibility with the current versions? I don't fully understand the scope of compatibility this class has to support (JDK 8, non-CRaC JDKs...)

rvansa: I wouldn't strive for proper CRaC support on any version of JDK, including older CRaCs  - that could explode into code bloat - but if it runs on JDK 8 it should run without errors on older CRaC, too. In other words, if the field is not present, I would silently suppress the `NoSuchFieldException` and return rather than throw `IllegalStateException`.

rvansa (commented):
While we're refactoring it I would still prefer to have priority as a top-level class rather than under `Core`, or nest it under `JDKResource`. But let's see if we really need it after the mentioned AOT enum patch.

TimPushkin:
Still relevant, at least until 25+6 is merged

TimPushkin:
After 25+6 has been merged all of the problematic tests now pass so this change is no longer needed

PR #219, opened 2025-03-31
8353241: [CRaC] Support ZGC

rvansa:
During my tests with https://github.com/CRaC/example-spring-boot I could not get the image size as low as with G1, but the presented changes improve the image footprint. As as anecdotal data, the image is 177 MB with G1 while 215 MB with ZGC (fastdebug build, `-Xmx1G`).

---------
### Progress

### Issue
 * [JDK-8353241](https://bugs.openjdk.org/browse/JDK-8353241): [CRaC] Support ZGC (**Enhancement** - P4)

### Reviewers
 * [Timofei Pushkin](https://openjdk.org/census#tpushkin) (@TimPushkin - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/219/head:pull/219` \
`$ git checkout pull/219`

`$ git checkout pull/219` \
`$ git pull https://git.openjdk.org/crac.git pull/219/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 219`

`$ git pr show -t 219`

</details>

<a href="https://git.openjdk.org/crac/pull/219.diff">https://git.openjdk.org/crac/pull/219.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/219#issuecomment-2765487486)
</details>


rvansa:
Conflicts resolved.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -422,6 +423,10 @@ Handle crac::checkpoint(jarray fd_arr, jobjectArray obj_arr, bool dry_run, jlong
 
   JFR_ONLY(Jfr::before_checkpoint();)
 
+  if (PrintNMTStatistics) {
+    MemTracker::final_report(tty);
+  }
+
```
TimPushkin, line 429: Is this left intentionally? If yes, I think nothing will be printed on the real VM exit (after restore) since there is `Atomic::cmpxchg(&g_final_report_did_run, false, true) == false` in `MemTracker::final_report()`.

src/hotspot/share/gc/z/zPageAllocator.cpp
```diff
@@ -1050,3 +1050,18 @@ void ZPageAllocator::threads_do(ThreadClosure* tc) const {
...
+    flushed = uncommit(&timeout, 0);
+    uncommitted += flushed;
+  } while (flushed > 0);
+  if (uncommitted > 0) {
+    EventZUncommit event;
+    log_info(gc, heap)("Uncommitted (cleanup): " SIZE_FORMAT "M(%.0f%%)",
+      uncommitted / M, percent_of(uncommitted, ZHeap::heap()->max_capacity()));
+    event.commit(uncommitted);
+  }
+}
```
TimPushkin, line 1067: No newline at the EOF

src/hotspot/share/gc/z/zPageCache.cpp
```diff
@@ -278,18 +280,20 @@ class ZPageCacheFlushForUncommitClosure : public ZPageCacheFlushClosure {
...
 public:
-  ZPageCacheFlushForUncommitClosure(size_t requested, uint64_t now, uint64_t* timeout)
+  ZPageCacheFlushForUncommitClosure(size_t requested, uint64_t now, uint64_t* timeout, uintx delay)
     : ZPageCacheFlushClosure(requested),
       _now(now),
-      _timeout(timeout) {
+      _timeout(timeout),
+      _delay(delay) {
     // Set initial timeout
     *_timeout = ZUncommitDelay;
```
TimPushkin, line 292: Shouldn't this also use `delay`? This shouldn't have any real influence now (the code that has `delay != ZUncommitDelay` doesn't use the timeout) but anyway.

src/hotspot/share/gc/z/zPageCache.cpp
```diff
@@ -22,13 +22,15 @@
...
 #include "precompiled.hpp"
+#include "gc/shared/gcLogPrecious.hpp"
 #include "gc/z/zGlobals.hpp"
 #include "gc/z/zList.inline.hpp"
 #include "gc/z/zNUMA.hpp"
 #include "gc/z/zPage.inline.hpp"
 #include "gc/z/zPageCache.hpp"
 #include "gc/z/zStat.hpp"
 #include "gc/z/zValue.inline.hpp"
+#include "logging/log.hpp"
```
TimPushkin, line 33: There is no logging code added so these new imports shouldn't be needed.

test/jdk/jdk/crac/fileDescriptors/ZGCTest.java
```diff
@@ -0,0 +1,50 @@
...
+import jdk.test.lib.crac.CracTest;
+
+/**
+ * @test
+ * @library /test/lib
+ * @build ZGCTest
+ * @run driver jdk.test.lib.crac.CracTest
+ * @requires (os.family == "linux")
+ */
+public class ZGCTest implements CracTest {
```
TimPushkin, line 37: A bit unclear why this is in `jdk/crac/fileDescriptors` directory. I guess because of the required `memfd` support but I would still put it in the general `jdk/crac` directory.

src/hotspot/share/gc/z/zPageAllocator.hpp
```diff
@@ -169,6 +169,8 @@ class ZPageAllocator {
   void handle_alloc_stalling_for_old(bool cleared_soft_refs);
 
   void threads_do(ThreadClosure* tc) const;
+
+  void cleanup_unused();
```
TimPushkin, line 173: This should not be addressed in this PR, but I find "cleanup_unused" name pretty undescriptive, both here and in `CollectedHeap` in general. It wasn't obvious to me that this is something used only by CRaC. I would propose `cleanup_before_checkpoint`, for example.

src/hotspot/share/runtime/crac.cpp
rvansa: Yes, I've left this intentionally. Well spotted, I'll make sure that the final report will run as well.

TimPushkin: I actually think that this should be done in as a separate change then. I would propose to either document that `PrintNMTStatistics` prints on checkpoint, or make it have several modes (0 — off, 1 — print on exit, 2 — print on checkpoint, 4 — print on both exit and checkpoint), or add a separate `PrintNMTStatisticsOnCheckpoint`. The reason why I am proposing this is because I believe we had a related request from community for this somewhere, so it would be nice to have this documented.

rvansa: OK, while I've used this for diagnostics on ZGC it would really make sense for its own PR.

rvansa:
Updated.

TimPushkin (approved):

PR #220, opened 2025-03-31
8353243: [CRaC] Show all options in engine help

TimPushkin:
C/R engines are now advised to list all options in `configuration_doc`. If JVM does not let users to control some options it states that in the engine help message.

crexec now documents internal options, such as `image_location` and `exec_location`, in its doc message.

This is how crexec's help looks with this change:
```
$ java -XX:CRaCEngineOptions=help
crexec - pseudo-CRaC-engine used to relay data from JVM to a "real" engine implemented as an executable (instead of a library). The engine executable is expected to have CRaC-CRIU-like CLI. Support of the configuration options also depends on the engine executable.

Configuration options:
* image_location=<path> (no default) - path to a directory with checkpoint/restore files.
* exec_location=<path> (no default) - path to the engine executable.
* keep_running=<true/false> (default: false) - keep the process running after the checkpoint or kill it.
* direct_map=<true/false> (default: true) - on restore, map process data directly from saved files. This may speedup the restore but the resulting process will not be the same as before the checkpoint.
* args=<string> (default: "") - free space-separated arguments passed directly to the engine executable, e.g. "--arg1 --arg2 --arg3".

Configuration options controlled by the JVM: image_location, exec_location
```

---------
### Progress

### Issue
 * [JDK-8353243](https://bugs.openjdk.org/browse/JDK-8353243): [CRaC] Show all options in engine help (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/220/head:pull/220` \
`$ git checkout pull/220`

`$ git checkout pull/220` \
`$ git pull https://git.openjdk.org/crac.git pull/220/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 220`

`$ git pr show -t 220`

</details>

<a href="https://git.openjdk.org/crac/pull/220.diff">https://git.openjdk.org/crac/pull/220.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/220#issuecomment-2765606173)
</details>

### DIFF
--- a/src/hotspot/share/runtime/crac_engine.cpp
+++ b/src/hotspot/share/runtime/crac_engine.cpp
@@ -1,20 +1,48 @@
+/*
+ * Copyright (c) 2025, Azul Systems, Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
 #include "precompiled.hpp"
 
 #include "crlib/crlib.h"
 #include "crlib/crlib_restore_data.h"
 #include "logging/log.hpp"
 #include "memory/allStatic.hpp"
+#include "nmt/memTag.hpp"
 #include "runtime/crac_engine.hpp"
 #include "runtime/globals.hpp"
 #include "runtime/os.hpp"
 #include "utilities/debug.hpp"
+#include "utilities/growableArray.hpp"
 #include "utilities/macros.hpp"
 #include "utilities/resourceHash.hpp"
 
 #include <cstddef>
 #include <cstring>
 
-// CRaC engine configuration options JVM sets directly instead of relaying from the user
+// CRaC engine configuration options that JVM sets directly.
+// If an option is controlled exctusively by JVM (i.e. the user is not allowed
+// to pass the option through CRaCEngineOptions) it should be listed in
+// CracEngine::vm_controlled_options().
 #define ENGINE_OPT_IMAGE_LOCATION "image_location"
 #define ENGINE_OPT_EXEC_LOCATION "exec_location"
 
@@ -157,7 +185,7 @@ static crlib_conf_t *create_conf(const crlib_api_t &api, const char *image_locat
 
   if (exec_location != nullptr) { // Only passed when using crexec
     guarantee(api.can_configure(conf, ENGINE_OPT_EXEC_LOCATION),
-              "crexec does not support an internal option: " ENGINE_OPT_EXEC_LOCATION);
+              "crexec does not support expected option: " ENGINE_OPT_EXEC_LOCATION);
     if (!api.configure(conf, ENGINE_OPT_EXEC_LOCATION, exec_location)) {
       log_error(crac)("crexec failed to configure: '" ENGINE_OPT_EXEC_LOCATION "' = '%s'", exec_location);
       api.destroy_conf(conf);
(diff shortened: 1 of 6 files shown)


src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -366,3 +394,31 @@ const char *CracEngine::configuration_doc() const {
...
+    log_debug(crac)("CRaC engine does not support expected option: " ENGINE_OPT_IMAGE_LOCATION);
+  }
+
+  const char *id = _description_api->identity(_conf);
+  if (id == nullptr) {
+    log_debug(crac)("CRaC engine failed to identify itself");
+    return opts;
+  }
+
+  if (strcmp(id, "crexec") == 0) {
```
rvansa, line 415: I don't think you need the `crexec` check here. I would also drop the logging parts, unlikely to be useful in the practice.

rvansa: Despite `crexec` is part of JVM codebase and it allows JVM to use executable-based engine implementations, I don't consider it a part of JVM; the separation line should be drawn at CRE API level. So it should not be a "JVM's implementation detail".

In the first version of the API it was called 'library_path`: informing the engine about a place where it should load other executables/libraries from. In fact, it is non-trivial to programmatically figure out from within a shared library was loaded from (if e.g. it needs to load some extra resource that should be in the same directory) so it might be useful.

TimPushkin: I agree with what you say about "implementation detail", looks like we just interpret these words a bit differently.

> In the first version of the API it was called 'library_path`: informing the engine about a place where it should load other executables/libraries from.

Yes, but just `library_path` is not enough, engine executable does not always come from there sines the user can provide an arbitrary absolute path. In the first version of the API `crexec` ended up using `args` instead of `library_path` for this.

We could pass `exec_location` to all engines that accept it (would be better to rename it to `engine_location` then) but I am not sure how useful this would be to other engines besides `crexec`: I've never tried to get a file address of a shared library from within itself but after some googling it doesn't seem too complicated (for a not so commonly needed thing).

rvansa: > Yes, but just library_path is not enough, engine executable does not always come from there sines the user can provide an arbitrary absolute path. In the first version of the API crexec ended up using args instead of library_path for this.
...
> would be better to rename it to `engine_location`

In the current impl it is telling the `lib` path within JDK installation location, rather than location of the engine (shared library). Anyway this doesn't have to be exhaustive: there might be other option (JVM wouldn't know about) to hint about some other path.

> I've never tried to get a file address of a shared library from within itself but after some googling it doesn't seem too complicated (for a not so commonly needed thing).

Looking into this again I see that it is actually quite simple using `dladdr`. I am not sure why I have resorted to reading `/proc/self/maps` in some of my code... However, if the engine is set using absolute path outside JVM it might be problematic to get the JVM path (`JAVA_HOME` not being set...).

> We could pass exec_location to all engines that accept it ... but I am not sure how useful this would be to other engines

I might have focused too much on this being "handy", but my main objective was to not have any code in JVM that would treat different engines differently.

Btw. there's third option: we could use the `dladdr/GetModuleFileName` in `crexec` and drop `exec_location` completely.

TimPushkin: > I might have focused too much on this being "handy", but my main objective was to not have any code in JVM that would treat different engines differently.

Well, JVM will treat `crexec` somewhat specially anyway because this is the only engine that is not requested by the user directly but rather by the JVM (when it finds that the user requested an executable engine).

> Btw. there's third option: we could use the dladdr/GetModuleFileName in crexec and drop exec_location completely.

Through `dladdr`/`GetModuleFileName` crexec can find where it is located, i.e. this is indeed a substitution for `library_path`. But we'll still need to pass the location of the engine executable because it can be any user-provided absolute path, not only `lib` within JDK installation. The most convenient way would be to use an option which leads us to `exec_location` again.

TimPushkin: > Anyway this doesn't have to be exhaustive: there might be other option (JVM wouldn't know about) to hint about some other path.

If JVM doesn't know about it then it should be passed by the user directly: `-XX:CRaCEngine=crexec -XX:CRaCEngineOptions=engine=criu` — this would be somewhat inconvenient from UX perspective

rvansa: > But we'll still need to pass the location of the engine executable (otherwise crexec won't know what exactly to call), and it can be any user-provided absolute path,

I had to re-read the code again - my understanding of the option meaning was really wrong, it is passing the resolved location of the executable. So now I understand your argument, and the "third option" is really not viable. And I have to withdraw the argument that it's useful to external implementations; in fact it's set to `nullptr` if `CRaCEngine` does not refer to a shared library.

>> Anyway this doesn't have to be exhaustive: there might be other option (JVM wouldn't know about) to hint about some other path.

> If JVM doesn't know about it then it should be passed by the user directly: -XX:CRaCEngine=crexec -XX:CRaCEngineOptions=engine=criu — this would be somewhat inconvenient from UX perspective

Due to my confusion we were talking about two different things. Let's imagine a hypotetical engine `foobar`; this engine would live in `$JAVA_HOME/lib/libfoobar.so` and be invoked with `-XX:CRaCEngine=foobar`.
* if the engine would for some reason need load `libjsig.so` it would need the `library_path`
* if the engine would need to execute binary `foobar-tool` it could likely refer to the same directory through `dladdr`
* if the engine needs to load `libzip.so` from a non-standard location, you would pass `-XX:CRaCEngineOptions=foobar.lib_dir=/opt/myzip` <- by non-exhaustive I meant that you could do this rather than lookup in engine's directory or `$JAVA_HOME/lib`

In any case, I would suggest to have a fixed list of reserved CRE options, and as you noted, omit engine-specific handling as much as possible. Let's keep things as simple as these can be.

PR #225, opened 2025-04-16
8354679: [CRaC] jdk.crac.management makes JdkManagementCheckSince fail

TimPushkin:
Fixes the failing test, by using `@since CRaC` in CRaC docs and `--ignoreSince` in since-checker tests which check them.

A since-checker test for `jdk.crac` module is also added but it is disabled until [JDK-8354921](https://bugs.openjdk.org/browse/JDK-8354921) is fixed.

Also adds the since-checking tests to CI.

---------
### Progress

### Issue
 * [JDK-8354679](https://bugs.openjdk.org/browse/JDK-8354679): [CRaC] jdk.crac.management makes JdkManagementCheckSince fail (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/225/head:pull/225` \
`$ git checkout pull/225`

`$ git checkout pull/225` \
`$ git pull https://git.openjdk.org/crac.git pull/225/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 225`

`$ git pr show -t 225`

</details>

<a href="https://git.openjdk.org/crac/pull/225.diff">https://git.openjdk.org/crac/pull/225.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/225#issuecomment-2809570986)
</details>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,6 +59,7 @@ jobs:
       matrix:
         test-name:
           - 'jdk/crac'
+          - 'jdk/since checks'
           - 'jdk/tier1 part 1'
           - 'jdk/tier1 part 2'
           - 'jdk/tier1 part 3'
@@ -75,6 +76,9 @@ jobs:
         include:
           - test-name: 'jdk/crac'
             test-suite: 'test/jdk/jdk/crac/'
+
+          - test-name: 'jdk/since checks'
+            test-suite: 'test/jdk/:jdk_since_checks'
 
           - test-name: 'jdk/tier1 part 1'
             test-suite: 'test/jdk/:tier1_part1'
--- a/src/jdk.crac/share/classes/jdk/crac/package-info.java
+++ b/src/jdk.crac/share/classes/jdk/crac/package-info.java
@@ -91,7 +91,7 @@ restoring instances from the image.
  * </li>
  * </ul>
  *
- * @since TBD
+ * @since CRaC
  */
 
 package jdk.crac;
--- a/src/jdk.crac/share/classes/module-info.java
+++ b/src/jdk.crac/share/classes/module-info.java
@@ -23,6 +23,13 @@
  * questions.
  */
 
+/**
+ * CRaC (Coordinated Restore at Checkpoint) API.
+ *
+ * @moduleGraph
+ * @since CRaC
+ */
+
 module jdk.crac {
 
     exports jdk.crac;
--- a/src/jdk.management/share/classes/jdk/crac/management/package-info.java
+++ b/src/jdk.management/share/classes/jdk/crac/management/package-info.java
@@ -25,6 +25,8 @@
 
 /**
  * This package contains management interfaces for CRaC.
+ *
+ * @since CRaC
  */
 
 package jdk.crac.management;
--- a/test/jdk/tools/sincechecker/modules/jdk.crac/JdkCracCheckSince.java
+++ b/test/jdk/tools/sincechecker/modules/jdk.crac/JdkCracCheckSince.java
@@ -1,5 +1,6 @@
 /*
- * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2025, Azul Systems, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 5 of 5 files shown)


rvansa:
I think that it's OK to add the symbols to 24 only; CRaC versioning can be considered orthogonal to upstream JDK versioning. At this stage we keep backward compatibility for users' convenience, not as a rule. Testing API compatibility is useful as CRaC'ed JDK should be 100% usable as a drop-in replacement for upstream JDKs.

If `jdk.crac` can't be addressed, we can leave it out of scope. However could you file a bug for `SinceChecker` in upstream?

Last but not least; please outline the steps (can be just a comment on this PR) what will we have to do when we rebase on top of JDK 26.

rvansa (approved):

TimPushkin:
> However could you file a bug for SinceChecker in upstream?

Filed the `SinceChecker` bug as [JDK-8354921](https://bugs.openjdk.org/browse/JDK-8354921).

> Last but not least; please outline the steps (can be just a comment on this PR) what will we have to do when we rebase on top of JDK 26.

Whenever we merge an update of the symbols from the mainline (can happen multiple times between a rampdown and the respective GA) we'll need to first overwrite our symbols with the incoming changes and then run `make/scripts/generate-symbol-data.sh` to update the symbols with CRaC's ones.

Whenever we add a method to the public CRaC Java API we also need to run the script to update the symbols.

I also found [JDK-8345212](https://bugs.openjdk.org/browse/JDK-8345212), which we don't yet have in this fork. It should allow us to continue using `@since TBD` (or `@since CRaC`, or whatever) and the test will pretend like the CRaC API has been added in the current version — we won't need to do the symbols updates. @rvansa  WDYT?

rvansa:
> Whenever we merge an update of the symbols from the mainline (can happen multiple times between a rampdown and the respective GA) we'll need to first overwrite our symbols with the incoming changes and then run make/scripts/generate-symbol-data.sh to update the symbols with CRaC's ones.

OK, I thought that we could somehow preserve the oldest CRaC JDK version where the symbol appeared. But we don't need to change the taglet on too many places, so let's use this as is.

> I also found [JDK-8345212](https://bugs.openjdk.org/browse/JDK-8345212), which we don't yet have in this fork. It should allow us to continue using @since TBD (or @since CRaC, or whatever) and the test will pretend like the CRaC API has been added in the current version — we won't need to do the symbols updates.

That sounds useful, but we don't need to merge changes out-of-band, this PR is good as it is. Let's integrate this and use the `-ignoreSince` the next time this code needs updating anyway.

TimPushkin:
> OK, I thought that we could somehow preserve the oldest CRaC JDK version where the symbol appeared. But we don't need to change the taglet on too many places, so let's use this as is.

Symbols are fixed after GA so after GA they'll be preserved. But between rampdown and GA we'll need to be updating them (only the version that has not yet been GA-ed) when they are updated upstream.

Example for JDK 26:
1. 25 rampdown happens, 26 is created and symbols for 25 along with it — when merging this, we'll take the new 25 symbols and run the script to update them from the last 25-CRaC build.
2. 25's symbols get updated — when merging this, we'll overwrite our 25's symbols with the incoming ones, then run the script again to update them with the last 25-CRaC build (the same build as in step one), then fixup the symbols manually to remove the changes unrelated to CRaC (there will be such changes because the last 25-CRaC build will be based on an older 25 than the updated symbols). This step can happen a few times.
3. 25 GA happens — after this moment 25's symbols are fixed, we won't need to touch them ever again.

> That sounds useful, but we don't need to merge changes out-of-band, this PR is good as it is. Let's integrate this and use the -ignoreSince the next time this code needs updating anyway.

I was proposing to wait until that change gets merged-in. I believe we need to decide now: we either merge this PR and start generating CRaC symbols from now on or wait for the `SinceChecker` update that should allow us to continue without generating the symbols (I actually need to try this locally by cherry-picking to see how it works before we decide). It will be weird if now we start generating the symbols but then we suddenly stop.

TimPushkin:
Oh, actually, I believe we are still merging 25 updates that occurred before 24 GA, so we'll go through the steps 2 and 3 above even for 24's symbols.

TimPushkin:
Still active. I need to check if [JDK-8345212](https://bugs.openjdk.org/browse/JDK-8345212) will work as expected and then we'll decide whether we want to go that way or the way this PR was originally done.

TimPushkin:
[JDK-8345212](https://bugs.openjdk.org/browse/JDK-8345212) works as expected. We will be able to just put `@since CRaC` to our `package-info`s and `module-info`s and add `--ignoreSince CRaC` to `SinceChecker` tests that touch CRaC sources.

I think this approach is better because we won't have to update the symbols on every release (possibly multiple times) and it seems to be the accepted way for long-running non-mainline projects (this is the reasoning in JDK-8345212).

JDK-8345212 will come with jdk-25+10, we are now at jdk-25+4. I'll leave this PR as a draft until then. After that I'll change it to satisfy JDK-8345212.

rvansa:
@TimPushkin 25+10 was just merged.

TimPushkin:
We should now use `@since CRaC` in CRaC docs and add `--ignoreSince CRaC` to since-checker tests of modules containing documented CRaC code.

I also added a disabled since-checker test for `jdk.crac` module. It can be enabled after [JDK-8354921](https://bugs.openjdk.org/browse/JDK-8354921) is fixed.

Since-checker will also be run in our CI.

PR #229, opened 2025-05-23
8357641: [CRaC] Support PID change during checkpoint-restore

TimPushkin:
Adds updates of cached PID on restore.


rvansa (approved):
LGTM, thanks!

PR #237, opened 2025-06-09
8358956: [CRaC] Fix NULL usage in hotspot

dimitryc:
jtreg test TestNoNULL.java fails due to a number of usages of NULL in CRaC code.

The test becomes a part of hg/tier1_common after [JDK-8352645](https://bugs.openjdk.org/browse/JDK-8352645)  (integrated in b17) and GHA starts reporting the failures after merge with jdk-25+17.

The patch replaces NULL with nullptr (and with null in comments and messages in asserts). The test passes with the change.


rvansa (approved):
LGTM, thanks!

dimitryc:
Thanks for the review!

PR #239, opened 2025-06-30
8361090: [CRaC] Make dev engines less confusing

TimPushkin:
- Makes `pauseengine` Linux-only since it is not fully implemented on MacOS and Windows
   - Also moves `criuengine` sources from `unix` to `linux` directory since it only supports Linux anyway
- Makes `simengine` and `pausengine` report errors and status in more details


test/jdk/jdk/crac/recursiveCheckpoint/Test.java
```diff
@@ -26,6 +26,7 @@
 /*
  * @test
  * @summary check that the recursive checkpoint is not allowed
+ * @requires (os.family == "linux")
```
TimPushkin, line 29: I believe this test does not need `pauseengine` at all and there even may be errors because of it, but I am going to test this internally first and create a separate PR for that later

src/java.base/linux/native/pauseengine/pauseengine.c
```diff
@@ -57,33 +53,30 @@ int main(int argc, char *argv[]) {
...
-            perror("fopen pidfile");
+            perror(PAUSEENGINE "fopen pidfile");
             kickjvm(jvm, -1);
             return 1;
         }
 
         fprintf(pidfile, "%d\n", jvm);
         fclose(pidfile);
 
+        printf(PAUSEENGINE "pausing the process, restore in another process to unpause it\n");
```
rvansa: I would rather see all these logging messages going into `stderr`.

TimPushkin: Warp prints its "Checkpoint successful" message to stout IIRC, then this should.go there too since it is essentially the same thing. We can make it disable-able via `-XX:CRaCEngineOptions=args=-quiet` for example.

TimPushkin: NVMD, Warp actually prints to stderr, I was mistaken

rvansa (approved):

PR #241, opened 2025-07-10
8364737: [CRaC] Support checkpointing with --patch-module by treating patch JARs as persistent

mz1999:
#### Summary

This change enhances CRaC's checkpointing capabilities to correctly handle applications launched with the `--patch-module` option. Previously, using a JAR file with `--patch-module` would cause checkpointing to fail due to unmanaged file descriptors. This patch resolves the issue by treating these JARs as persistent resources, similar to how classpath JARs are handled.

#### Problem

When a JAR file is specified via `--patch-module`, it is opened by two distinct mechanisms within the JVM:

1.  **Native `ClassLoader`**: The C++ `ClassLoader` implementation opens the JAR file to build its internal class path structures. This results in a file descriptor that is not managed by CRaC's Java-level resource tracking, leading to it being flagged as a "BAD: opened by application" file during `check_fds`.
2.  **Java `ModulePatcher`**: The Java-level `jdk.internal.module.ModulePatcher` also opens the same JAR to scan for packages and resources. While this could be made CRaC-aware, the native-level file descriptor remains an issue.

This dual-opening mechanism leads to `CheckpointOpenFileException` being thrown for both file descriptors during a checkpoint, preventing a successful image creation. The log output clearly illustrates this problem:

```
$ jcmd 3403046 JDK.checkpoint
3403046:
...
JVM: FD fd=4 type=regular path="/home/mazhen/works/patch-demo/legacy-patch.jar" BAD: opened by application
JVM: FD fd=5 type=regular path="/home/mazhen/works/patch-demo/legacy-patch.jar" OK: claimed by java code
...
An exception during a checkpoint operation:
jdk.internal.crac.mirror.CheckpointException
    Suppressed: jdk.internal.crac.mirror.impl.CheckpointOpenFileException: legacy-patch.jar
        at java.base/jdk.internal.crac.JDKFileResource.lambda$beforeCheckpoint$0(JDKFileResource.java:90)
        ...
    Caused by: java.lang.Exception: This file descriptor was created by main at epoch:1752027938791 here
        at java.base/jdk.internal.crac.JDKFdResource.<init>(JDKFdResource.java:40)
        ...
        at java.base/jdk.internal.module.ModulePatcher$JarResourceFinder.<init>(ModulePatcher.java:433)
        ...
    Suppressed: jdk.internal.crac.mirror.impl.CheckpointOpenFileException: FD fd=4 type=regular path=/home/mazhen/works/patch-demo/legacy-patch.jar
        at java.base/jdk.internal.crac.mirror.Core.translateJVMExceptions(Core.java:115)
        ...
```

#### Solution

The fundamental assumption behind this solution is that JARs provided via `--patch-module` should be treated as persistent resources, just like those on the classpath. It is expected that these files will be available at the same location after a restore.

This PR implements a two-pronged approach to address both file handles:

1.  **Java-level Fix**: The `jdk.internal.module.ModulePatcher$JarResourceFinder` is modified to use `jdk.internal.util.jar.PersistentJarFile` instead of the standard `java.util.jar.JarFile`. This ensures that the file handle opened by the Java module system is registered as a CRaC `JDKResource`, making it a managed and permissible resource during checkpointing.

2.  **Native-level Fix**: To handle the file descriptor opened by the native `ClassLoader`, we introduce a new mechanism to whitelist it:
    *   A new function, `ClassLoader::get_patch_module_entry_fds()`, is added in `classLoader.cpp`. This function iterates through the `_patch_mod_entries` and collects the file descriptors of all associated JAR files.
    *   In `crac_linux.cpp`, the `VM_Crac::check_fds` function now calls this new method to retrieve the list of patch-module file descriptors. These FDs are then added to the allowlist, and `check_fds` will correctly identify them as `OK: claimed by patch-module`, thus avoiding the exception.

By implementing these changes, we ensure that both file descriptors associated with a patched module's JAR are correctly recognized as persistent and valid, allowing the checkpoint to proceed successfully.

---------
### Progress

### Issue
 * [JDK-8364737](https://bugs.openjdk.org/browse/JDK-8364737): [CRaC] Support checkpointing with --patch-module by treating patch JARs as persistent (**Enhancement** - P3)

### Reviewers
 * [Radim Vansa](https://openjdk.org/census#rvansa) (@rvansa - Committer) ⚠️ Review applies to [5ec6037a](https://git.openjdk.org/crac/pull/241/files/5ec6037aa97e2af3903b353c6938fcd71f4c0ea1)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/241/head:pull/241` \
`$ git checkout pull/241`

`$ git checkout pull/241` \
`$ git pull https://git.openjdk.org/crac.git pull/241/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 241`

`$ git pr show -t 241`

</details>

<a href="https://git.openjdk.org/crac/pull/241.diff">https://git.openjdk.org/crac/pull/241.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/241#issuecomment-3151946979)
</details>


rvansa (commented):
Thank you for the contribution! The change looks good in general, could you add a JTReg test that demonstrates that this is working as expected, and make sure that the CI runs it?

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -757,6 +757,24 @@ GrowableArray<int> ClassLoader::get_classpath_entry_fds() {
   return fds;
 }
 
+GrowableArray<int> ClassLoader::get_patch_module_entry_fds() {
+  GrowableArray<int> fds;
+  assert(Thread::current()->is_VM_thread(), "should be called from VM op");
+  if (_patch_mod_entries == NULL) {
```
TimPushkin:
```suggestion
  if (_patch_mod_entries == nullptr) {
```

src/hotspot/share/classfile/classLoader.cpp
```diff
@@ -757,6 +757,24 @@ GrowableArray<int> ClassLoader::get_classpath_entry_fds() {
...
+    return fds;
+  }
+  // we don't use mutexes here because it is called from VM op
```
TimPushkin: As I see it, all patch module entries are opened once at VM initialization (in `ClassLoader::setup_patch_mod_entries()`) and checkpoint cannot be called before then. If that is so, there should be no VM thread / mutex requirements to read it.

TimPushkin (commented):
Small comments from me, nothing functionally important

mz1999:
> Thank you for the contribution! The change looks good in general, could you add a JTReg test that demonstrates that this is working as expected, and make sure that the CI runs it?

Thank you for the feedback! I've added a new JTReg test,  `crac/fileDescriptors/PatchModuleTest.java`.

The test is designed to cover all aspects of the fix:

1.  **Simulates a Real-World Patch**: It dynamically compiles a new class and packages it into a JAR. This JAR is then used with `--patch-module` to add the new class to the `java.base` module at runtime. 

2.  **Verifies Patch Functionality**: The test confirms that the class loaded from the patched module works as expected both before the checkpoint and after the restore.

3. **Ensures Checkpoint/Restore Succeeds**: A key part of the test is to ensure that the entire checkpoint and restore process completes successfully without throwing a CheckpointException. This directly confirms that the original bug (unhandled file descriptors causing checkpoint failure) is resolved.

 4.  **Validates Both Fixes**: The test analyzes the output of the checkpointing process to explicitly verify both parts of the solution.

src/hotspot/share/classfile/classLoader.cpp
mz1999: Thanks! Good catch. Using nullptr is definitely better practice. I'll make the change.

src/hotspot/share/classfile/classLoader.cpp
mz1999: That's a great point. Your analysis of the `_patch_mod_entries` lifecycle makes perfect sense. Since it's effectively read-only after initialization, the VM thread assertion isn't necessary. I'll remove it and the related comment. Thanks for spotting this!

rvansa:
@TimPushkin I wonder if we should make `simengine` the default in `CracBuilder`; that way we can encourage everyone to test everywhere and use CRIU only when the test is really specific to any of its behaviour (which it shouldn't be for the most part - everything `Resource`-related is applicable to all engines).

TimPushkin:
@rvansa Yes, I agree

test/jdk/jdk/crac/fileDescriptors/PatchModuleTest.java
```diff
@@ -0,0 +1,140 @@
...
+        // Verify the patch is still active after restore
+        String infoAfter = invokePatchedMethod();
+        assertEquals(PATCH_MESSAGE, infoAfter, "Patch should remain active after restore.");
+    }
+
+    private String invokePatchedMethod() throws Exception {
+        Class<?> patcherClass = Class.forName(PATCHED_CLASS_FULL_NAME);
+        return (String) patcherClass.getMethod("getInfo").invoke(null);
+    }
+}
```
TimPushkin: I haven't looked through the test but if you will be applying Radim's suggestions please also add a newline at the end of file

mz1999: Good catch, thanks! I've pushed a commit with the added newline.

I've also left a comment on Radim's suggestion explaining why I'm sticking with the criu engine for this test to ensure a full physical restore is validated.

Appreciate the review!

mz1999:
> Great work, let's just sort out the OCA and enable running the testsuite in your repo.
> 
> Btw. the test could work using just `simengine` (there's nothing engine-specific and that would let us run on other platforms than only Linux) but it's OK to keep this on a real engine (CRIU).

Hi @rvansa , 

a quick follow-up on the OCA status. I've been waiting for approval for over a month and it seems to be stalled.  Would you have any advice or be able to help nudge the process internally? Any help would be a huge boost. Thanks!

rvansa:
@mz1999 OCA is handled by Oracle, so we don't really have such internal connection. When I was requesting the approval it also took a while and I had to ping `opensource_ww_grp@oracle.com` asking them to check on the request. Are you requesting individual OCA or with company? In the latter case they request official approval that you can contribute to OSS.

rvansa:
@mz1999 Hi, any luck getting a response from Oracle?

mz1999:
> @mz1999 Hi, any luck getting a response from Oracle?

Hi @rvansa ,

I am applying for an Individual OCA. I suspected there might have been an issue with the original submission, so I withdrew it and submitted a fresh Individual Agreement last week (on July 22nd).

It's currently 'under review', and I'm hopeful this one will be processed smoothly. I'll post an update here as soon as it's approved.

rvansa:
@mz1999 It seems that your OCA has landed, please type `/integrate` and I'll sponsor this.

PR #246, opened 2025-07-19
8362837: [CRaC] jdk/crac/MXBean.java can fail on macOS

jankratochvil:
The CI runs fine but locally (Linux Fedora 42 x86_64) it is failing for me - for any testcase trying to use OutputAnalyzer for the restore-only part. Such testcases failing for me are for example:
- test/jdk/jdk/crac/fileDescriptors/CheckpointWithOpenFdsTest.java
- test/jdk/jdk/crac/fileDescriptors/LoggingFileOpenTest.java
- test/jdk/jdk/crac/SecureRandom/ReseedTest.java

It does not implement the part

> close to 0

as that is the problem being fixed.

---------
### Progress

### Issue
 * [JDK-8362837](https://bugs.openjdk.org/browse/JDK-8362837): [CRaC] jdk/crac/MXBean.java can fail on macOS (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/246/head:pull/246` \
`$ git checkout pull/246`

`$ git checkout pull/246` \
`$ git pull https://git.openjdk.org/crac.git pull/246/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 246`

`$ git pr show -t 246`

</details>

<a href="https://git.openjdk.org/crac/pull/246.diff">https://git.openjdk.org/crac/pull/246.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/246#issuecomment-3092569085)
</details>

### DIFF
--- a/test/jdk/jdk/crac/MXBean.java
+++ b/test/jdk/jdk/crac/MXBean.java
@@ -76,21 +76,23 @@ public class MXBean implements CracTest {
             process.waitForPausePid();
             restoreStartTime = System.currentTimeMillis();
             builder.doRestore();
-            System.out.println("RestoreStartTime " + restoreStartTime + " " + formatTime(restoreStartTime));
         } else {
+            restoreStartTime = System.currentTimeMillis(); // A very rough approximation
             output = builder.engine(CracEngine.SIMULATE).startCheckpoint().waitForSuccess().outputAnalyzer();
-            restoreStartTime = -1; // Unknown
         }
+        System.out.println("RestoreStartTime " + restoreStartTime + " " + formatTime(restoreStartTime));
 
         final long uptimeSinceRestore = Long.parseLong(output.firstMatch("UptimeSinceRestore ([0-9-]+)", 1));
         assertGTE(uptimeSinceRestore, 0L, "Bad UptimeSinceRestore");
-        assertLT(uptimeSinceRestore, TIME_TOLERANCE, "UptimeSinceRestore should be close to 0");
+        assertLT(uptimeSinceRestore, TIME_TOLERANCE, "UptimeSinceRestore should be a bit greater than 0");
 
         final long restoreTime = Long.parseLong(output.firstMatch("RestoreTime ([0-9-]+)", 1));
         assertGTE(restoreTime, 0L, "Bad RestoreTime");
         if (Platform.isLinux()) {
-            assertLT(Math.abs(restoreTime - restoreStartTime), TIME_TOLERANCE,
-                    "RestoreTime " + restoreTime + " should be close to " + restoreStartTime);
+            assertLT(restoreTime - restoreStartTime, TIME_TOLERANCE,
+                    "RestoreTime " + restoreTime + " should be a bit greater than " + restoreStartTime);
+        } else {
+            assertGT(restoreTime, restoreStartTime, "Time has gone backwards?");
         }
     }
 }
--- a/test/jdk/jdk/crac/MXBean.java
+++ b/test/jdk/jdk/crac/MXBean.java
@@ -26,6 +26,7 @@
 import jdk.crac.*;
 import jdk.crac.management.*;
 
+import jdk.test.lib.Platform;
 import jdk.test.lib.crac.CracBuilder;
 import jdk.test.lib.crac.CracEngine;
 import jdk.test.lib.crac.CracTest;
@@ -64,19 +65,34 @@ public class MXBean implements CracTest {
     @Override
     public void test() throws Exception {
         long start = System.currentTimeMillis();
+        CracBuilder builder = new CracBuilder();
 
-        OutputAnalyzer output = new CracBuilder().engine(CracEngine.SIMULATE)
-                .captureOutput(true)
-                .startCheckpoint().waitForSuccess().outputAnalyzer();
+        OutputAnalyzer output;
+        if (Platform.isLinux()) {
+	    builder.doCheckpoint();
+
+	    long restoreStart = System.currentTimeMillis();
+	    output = builder.captureOutput(true).doRestore().outputAnalyzer();
+
+	    long restoreTimePassed = System.currentTimeMillis() - restoreStart;
+	    System.err.println("restoreTimePassed="+restoreTimePassed);
+	    if (restoreTimePassed < 0 || TIME_TOLERANCE < restoreTimePassed) {
+		throw new Error("bad time since restore started: " + restoreTimePassed);
+	    }
+        } else {
+	    output = builder.engine(CracEngine.SIMULATE)
+		    .captureOutput(true)
+		    .startCheckpoint().waitForSuccess().outputAnalyzer();
+        }
 
         long restoreUptime = Long.parseLong(output.firstMatch("UptimeSinceRestore ([0-9-]+)", 1));
-        if (restoreUptime < 0 || TIME_TOLERANCE < restoreUptime) {
+        System.err.println("restoreUptime=" + restoreUptime);
+        if (restoreUptime < 0) {
             throw new Error("bad UptimeSinceRestore: " + restoreUptime);
         }
 
         long restoreTime = Long.parseLong(output.firstMatch("RestoreTime ([0-9-]+)", 1));
-        restoreTime -= start;
-
-        assertLT(Math.abs(restoreTime), TIME_TOLERANCE, "bad RestoreTime: " + restoreTime);
+        System.err.println("restoreTime=" + restoreTime);
+        assertLT(start, restoreTime, "bad RestoreTime: " + restoreTime);
     }
 }
--- a/test/jdk/jdk/crac/MXBean.java
+++ b/test/jdk/jdk/crac/MXBean.java
@@ -53,8 +53,6 @@ public class MXBean implements CracTest {
 
         Core.checkpointRestore();
 
-        System.out.println("UptimeSinceRestore " + cracMXBean.getUptimeSinceRestore());
-
         long restoreTime = cracMXBean.getRestoreTime();
         System.out.println("RestoreTime " + restoreTime + " " +
             DateTimeFormatter.ofPattern("E dd LLL yyyy HH:mm:ss.n").format(
@@ -75,7 +73,7 @@ public class MXBean implements CracTest {
             output = builder.captureOutput(true).doRestore().outputAnalyzer();
 
             long restoreTimePassed = System.currentTimeMillis() - restoreStart;
-            System.err.println("restoreTimePassed="+restoreTimePassed);
+            System.err.println("restoreTimePassed=" + restoreTimePassed);
             if (restoreTimePassed < 0 || TIME_TOLERANCE < restoreTimePassed) {
                 throw new Error("bad time since restore started: " + restoreTimePassed);
             }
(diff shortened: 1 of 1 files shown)


jankratochvil:
> GitHub Actions should run in your repo just like on PRs here, so you should not need to create PRs to test CI. Except for whitespace checking...

True, thanks.

test/jdk/jdk/crac/MXBean.java
```diff
@@ -64,19 +65,34 @@ public void exec() throws CheckpointException, RestoreException {
...
+        OutputAnalyzer output;
+        if (Platform.isLinux()) {
+            builder.doCheckpoint();
+
+            long restoreStart = System.currentTimeMillis();
+            output = builder.captureOutput(true).doRestore().outputAnalyzer();
+
+            long restoreTimePassed = System.currentTimeMillis() - restoreStart;
+            System.err.println("restoreTimePassed="+restoreTimePassed);
+            if (restoreTimePassed < 0 || TIME_TOLERANCE < restoreTimePassed) {
```
TimPushkin: What is the purpose of this check? As I understand it, it checks that restore + restored code execution takes time close to zero which I don't think is a reasonable expectation (it passes because the tolerance is high).

Also the test is supposed to be about `CRaCMXBean` which is not involved here.

test/jdk/jdk/crac/MXBean.java
```diff
@@ -64,19 +65,34 @@ public void exec() throws CheckpointException, RestoreException {
...
+        } else {
+            output = builder.engine(CracEngine.SIMULATE)
+                    .captureOutput(true)
+                    .startCheckpoint().waitForSuccess().outputAnalyzer();
+        }
 
         long restoreUptime = Long.parseLong(output.firstMatch("UptimeSinceRestore ([0-9-]+)", 1));
-        if (restoreUptime < 0 || TIME_TOLERANCE < restoreUptime) {
+        System.err.println("restoreUptime=" + restoreUptime);
+        if (restoreUptime < 0) {
```
TimPushkin: AFAIK we have not had problems with this check, only with the below one. "UptimeSinceRestore" should be close to 0 on all platforms

TimPushkin (commented):
Maybe I do not understand the changes but to me it looks like they won't function right.

<details>

<summary>Here is how I envisioned it in my original suggestion</summary>

I reduced the tolerance since the measured time should be less now, but maybe it is too bold of a reduction. I just feel like 10 sec is a too big tolerance.

```diff
diff --git a/test/jdk/jdk/crac/MXBean.java b/test/jdk/jdk/crac/MXBean.java
index 336e3eecd12..0bb6605432e 100644
--- a/test/jdk/jdk/crac/MXBean.java
+++ b/test/jdk/jdk/crac/MXBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, Azul Systems, Inc. All rights reserved.
+ * Copyright (c) 2022, 2025, Azul Systems, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -26,6 +26,7 @@
 import jdk.crac.*;
 import jdk.crac.management.*;
 
+import jdk.test.lib.Platform;
 import jdk.test.lib.crac.CracBuilder;
 import jdk.test.lib.crac.CracEngine;
 import jdk.test.lib.crac.CracTest;
@@ -35,7 +36,7 @@
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 
-import static jdk.test.lib.Asserts.assertLT;
+import static jdk.test.lib.Asserts.*;
 
 /**
  * @test
@@ -44,7 +45,12 @@
  * @run driver jdk.test.lib.crac.CracTest
  */
 public class MXBean implements CracTest {
-    static final long TIME_TOLERANCE = 10_000; // ms
+    static final long TIME_TOLERANCE = 1000; // ms
+
+    private static String formatTime(long t) {
+        return DateTimeFormatter.ofPattern("E dd LLL yyyy HH:mm:ss.n").format(
+                Instant.ofEpochMilli(t).atZone(ZoneId.systemDefault()));
+    }
 
     @Override
     public void exec() throws CheckpointException, RestoreException {
@@ -55,28 +61,36 @@ public void exec() throws CheckpointException, RestoreException {
         System.out.println("UptimeSinceRestore " + cracMXBean.getUptimeSinceRestore());
 
         long restoreTime = cracMXBean.getRestoreTime();
-        System.out.println("RestoreTime " + restoreTime + " " +
-            DateTimeFormatter.ofPattern("E dd LLL yyyy HH:mm:ss.n").format(
-                Instant.ofEpochMilli(restoreTime)
-                    .atZone(ZoneId.systemDefault())));
+        System.out.println("RestoreTime " + restoreTime + " " + formatTime(restoreTime));
     }
 
     @Override
     public void test() throws Exception {
-        long start = System.currentTimeMillis();
+        final var builder = new CracBuilder().captureOutput(true);
 
-        OutputAnalyzer output = new CracBuilder().engine(CracEngine.SIMULATE)
-                .captureOutput(true)
-                .startCheckpoint().waitForSuccess().outputAnalyzer();
-
-        long restoreUptime = Long.parseLong(output.firstMatch("UptimeSinceRestore ([0-9-]+)", 1));
-        if (restoreUptime < 0 || TIME_TOLERANCE < restoreUptime) {
-            throw new Error("bad UptimeSinceRestore: " + restoreUptime);
+        final OutputAnalyzer output;
+        final long restoreStartTime;
+        if (Platform.isLinux()) { // Linux is currently the only platform supporting non-immediate restore
+            final var process = builder.engine(CracEngine.PAUSE).startCheckpoint();
+            output = process.outputAnalyzer();
+            process.waitForPausePid();
+            restoreStartTime = System.currentTimeMillis();
+            builder.doRestore();
+            System.out.println("RestoreStartTime " + restoreStartTime + " " + formatTime(restoreStartTime));
+        } else {
+            output = builder.engine(CracEngine.SIMULATE).startCheckpoint().waitForSuccess().outputAnalyzer();
+            restoreStartTime = -1; // Unknown
         }
 
-        long restoreTime = Long.parseLong(output.firstMatch("RestoreTime ([0-9-]+)", 1));
-        restoreTime -= start;
+        final long uptimeSinceRestore = Long.parseLong(output.firstMatch("UptimeSinceRestore ([0-9-]+)", 1));
+        assertGTE(uptimeSinceRestore, 0L, "Bad UptimeSinceRestore");
+        assertLT(uptimeSinceRestore, TIME_TOLERANCE, "UptimeSinceRestore should be close to 0");
 
-        assertLT(Math.abs(restoreTime), TIME_TOLERANCE, "bad RestoreTime: " + restoreTime);
+        final long restoreTime = Long.parseLong(output.firstMatch("RestoreTime ([0-9-]+)", 1));
+        assertGTE(restoreTime, 0L, "Bad RestoreTime");
+        if (Platform.isLinux()) {
+            assertLT(Math.abs(restoreTime - restoreStartTime), TIME_TOLERANCE,
+                    "RestoreTime " + restoreTime + " should be close to " + restoreStartTime);
+        }
     }
 }

```

If we suppose that the system time cannot jump a lot backwards during the tests, we can make the assertions stronger (this diff is on top of the diff above):
```diff
diff --git a/test/jdk/jdk/crac/MXBean.java b/test/jdk/jdk/crac/MXBean.java
index 0bb6605432e..1458d14fcf4 100644
--- a/test/jdk/jdk/crac/MXBean.java
+++ b/test/jdk/jdk/crac/MXBean.java
@@ -76,21 +76,23 @@ public void test() throws Exception {
             process.waitForPausePid();
             restoreStartTime = System.currentTimeMillis();
             builder.doRestore();
-            System.out.println("RestoreStartTime " + restoreStartTime + " " + formatTime(restoreStartTime));
         } else {
+            restoreStartTime = System.currentTimeMillis(); // A very rough approximation
             output = builder.engine(CracEngine.SIMULATE).startCheckpoint().waitForSuccess().outputAnalyzer();
-            restoreStartTime = -1; // Unknown
         }
+        System.out.println("RestoreStartTime " + restoreStartTime + " " + formatTime(restoreStartTime));
 
         final long uptimeSinceRestore = Long.parseLong(output.firstMatch("UptimeSinceRestore ([0-9-]+)", 1));
         assertGTE(uptimeSinceRestore, 0L, "Bad UptimeSinceRestore");
-        assertLT(uptimeSinceRestore, TIME_TOLERANCE, "UptimeSinceRestore should be close to 0");
+        assertLT(uptimeSinceRestore, TIME_TOLERANCE, "UptimeSinceRestore should be a bit greater than 0");
 
         final long restoreTime = Long.parseLong(output.firstMatch("RestoreTime ([0-9-]+)", 1));
         assertGTE(restoreTime, 0L, "Bad RestoreTime");
         if (Platform.isLinux()) {
-            assertLT(Math.abs(restoreTime - restoreStartTime), TIME_TOLERANCE,
-                    "RestoreTime " + restoreTime + " should be close to " + restoreStartTime);
+            assertLT(restoreTime - restoreStartTime, TIME_TOLERANCE,
+                    "RestoreTime " + restoreTime + " should be a bit greater than " + restoreStartTime);
+        } else {
+            assertGT(restoreTime, restoreStartTime, "Time has gone backwards?");
         }
     }
 }

```

</details>

test/jdk/jdk/crac/MXBean.java
jankratochvil: This check is implementing your words:

> For platforms that have engines with restore time passing (only Linux currently, I think pauseengine is enough here) — this will be almost the same test as it is now, but it will do an actual restore call and measure the restore start time from that call.

Regarding `CRaCMXBean` - I do not see any need for it to implement your words above. Although maybe I did not understand it as I do not see why/how to use `pauseengine` when you want `actual restore call`.

test/jdk/jdk/crac/MXBean.java
jankratochvil: What do you exactly mean by `close to 0`? I already mention here in the initial comment I did not implement `close to 0`. Do you mean less than `TIME_TOLERANCE`? But that is exactly the problem why it sometimes fails on Mac CI.

I am going to remove the `restoreUptime < 0` test when you do not like it.

test/jdk/jdk/crac/MXBean.java
TimPushkin: I meant recording the time of restore start and then comparing it with the restore start time `MXBean` provides. 

`pauseengine` should be enough because we need an engine that allows to start restoring at some particular moment, not immediately after checkpoint as `simengine` does.

Anyway, I posted the code that implements what I meant, it should be clearer.

test/jdk/jdk/crac/MXBean.java
TimPushkin: CI was failing for the check below AFAIK. In my understanding this one should not be failing the way it is.

TimPushkin (approved):
LGTM, although it is my own suggestion.

If the reduced time tolerance will be too low for our CI we can raise it downstream. But I hope it won't be necessary.

jankratochvil:
So either it is
- a performance test - then it needs to track the time per platform, the limit should be tight and it must not run in parallel with other parts of the testsuite, IIRC perf lab dept. should take care of it
- or it just checks whether it completes at all - and then it needs no limit at all as there is already a JTREG testcase limit

TimPushkin:
@rvansa In our downstream CI it looks like 10 seconds is not enough for MacOS, that is why I proposed to rework the test to use an engine that allows to propagate restore time on platforms where such engine exists (Linux) and not use `TIME_TOLERANCE` at all for `getRestoreTime()` on platforms where there is no such engine (MacOS, Windows). To me it looks like a better solution than just increasing the tolerance because that would make the `getUptimeSinceRestore()` check and `getRestoreTime()` check on other platforms less strict.

TimPushkin:
Maybe I am indeed overcomplicating it and just checking that the values are somewhat meaningful with a large tolerance is enough

jankratochvil:
@TimPushkin OK so going to submit only a oneliner downstream, sorry for your time.

PR #251, opened 2025-07-28
8364156: [CRaC] Recompile methods decompiled during or shortly after C/R

TimPushkin:
Implements recording and triggering recompilation of methods decompiled during and shortly after C/R. The length of the "shortly-after" period can be controlled by the user via `-Djdk.crac.recompilation-delay-ms=<long>`, the default is 10ms.

Measuring the after-restore regression before and after this change shows improvement on hello-world apps. The numbers are the regression of the first after-restore 1 sec iteration vs average over 5 last before-checkpoint 1 sec iterations, i.e. lower means better; mean ± std dev over 30 runs. Measured in a Linux VM on MacOS, with CRIU, with 5ms after-restore recompilation delay.

|                          | Before      | After       |
|--------------------------|-------------|-------------|
| Helidon hello-world      | 0.9 ± 1.2 % | 0.4 ± 1.0 % |
| Spring boot CRaC example | 9.5 ± 2.4 % | 6.4 ± 2.5 % |

Currently decompilations are recorded during C/R + during the after-restore delay and then they all get recompiled together. An enhancement planned in the future is to record only during C/R recompiling immediately after restore and during the after-restore delay recompile immediately without recording. This way the recompilations would finish faster, also setting `jdk.crac.recompilation-delay-ms` to a higher value would have less negative impact.

Implementation notes:
- Native code (`nmethod`) is considered "decompiled" when `make_not_entrant(...)` is called on it — a new mandatory parameter is added to the signature of this method that controls whether CRaC should consider this particular decompilation for recompilation or not.
  - The parameter is needed because we do not want to recompile, for example, when `nmethod` is thrown away once a replacement for it has been compiled (e.g. on another compilation level).
  - The parameter has no default so that we get compilation errors when new calls to `make_not_entrant(...)` are introduced in the mainline — we will be deciding whether these new calls should be considered for recompilation or not. If CRaC was in the mainline I would enable the recompilation by default.
- The new native `CRaCRecorder` class is the core of the implementation — it records the decompilations and handles their recompilation.
  - `jdk.crac.internal.mirror.Core` notifies it when to start and stop the recording, it also handles the after-restore recompilation delay.
  - `nmethod.make_not_entrant(...)` notifies it about decompiled code.

---------
### Progress

### Issue
 * [JDK-8364156](https://bugs.openjdk.org/browse/JDK-8364156): [CRaC] Recompile methods decompiled during or shortly after C/R (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/251/head:pull/251` \
`$ git checkout pull/251`

`$ git checkout pull/251` \
`$ git pull https://git.openjdk.org/crac.git pull/251/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 251`

`$ git pr show -t 251`

</details>

<a href="https://git.openjdk.org/crac/pull/251.diff">https://git.openjdk.org/crac/pull/251.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/251#issuecomment-3125844765)
</details>


src/hotspot/share/runtime/cracRecompiler.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  const GrowableArrayCHeap<CompilationInfo *, MemTag::mtInternal> *decomps;
+  {
+    const MutexLocker ml(decompilations_lock, Mutex::_no_safepoint_check_flag);
+    precond(is_recording_decompilations());
+    decomps = const_cast<GrowableArrayCHeap<CompilationInfo *, MemTag::mtInternal> *>(decompilations);
+    assert(decomps != nullptr, "recording has not been started");
+    log_debug(crac)("Finishing recording decompilations and requesting %i recompilations", decomps->length());
+    // fence should allow the recorder threads to stop locking quicker
+    Atomic::release_store_fence(&decompilations, static_cast<decltype(decompilations)>(nullptr));
```
TimPushkin: I've realized that after this nothing is guarding the recorded methods from being deleted by `RedefineClasses`. Need to fix this before integrating.

src/hotspot/share/runtime/cracRecompiler.cpp
```diff
@@ -0,0 +1,221 @@
...
+  Atomic::release_store_fence(&decompilations, new GrowableArrayCHeap<CompilationInfo *, MemTag::mtInternal>());
+  postcond(is_recording_decompilations());
+}
+
+void CRaCRecompiler::record_decompilation(const nmethod &nmethod) {
+  if (!is_recording_decompilations()) {
+    return; // Fast pass to not acquire a lock when no C/R occurs (i.e. most of the time)
+  }
+  const MutexLocker ml(decompilations_lock, Mutex::_no_safepoint_check_flag);
+  auto * const decomps = const_cast<GrowableArrayCHeap<CompilationInfo *, MemTag::mtInternal> *>(decompilations);
```
rvansa: Do you really need the `const_cast` here? I wonder why you use `GrowableArrayCHeap<...> * const` here and `const GrowableArrayCHeap<...> *` in the next function (or am I misinterpreting the `auto` use?)

test/jdk/jdk/crac/recompiler/NaturalDecompilationTest.java
```diff
@@ -0,0 +1,171 @@
...
+        final var whiteBox = WhiteBox.getWhiteBox();
+        final var testMethodRef = NaturalDecompilationTest.class.getDeclaredMethod(TEST_METHOD_NAME, int.class);
+
+        timedDoWhile("compilation", () -> {
+            for (int i = 0; i < 2000; i++) {
+                final var res = testMethod(TEST_ARG_EXPECTED);
+                blackhole(res);
+            }
+            try {
+                Thread.sleep(500); // Time to compile
```
rvansa, line 121: I think this has some potential for false positives in the testsuite. I guess there is no API in whitebox to wait until this is observed? Alternatively watching the output in `test()` and signaling the process somehow...

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
```diff
@@ -66,6 +67,8 @@ private static class FlagsHolder {
         private FlagsHolder() {}
         public static final boolean TRACE_STARTUP_TIME =
             Boolean.getBoolean("jdk.crac.trace-startup-time");
+        public static final long RECOMPILATION_DELAY_MS =
```
rvansa: I am thinking about an option for the 'old behaviour', at least for some testing. Could you interpret negative values as "don't disable the recompilations at all"?

And another thing: by putting this into a static variable you're fixing the value at checkpoint time. I believe it would be better if we could adjust this on restore (I think that at the point where you use this the system properties are already updated). Naturally if the recompilations were not disabled through a negative value this should have not effect but a warning being printed out.

test/jdk/jdk/crac/recompiler/RecompilationDelayTest.java
```diff
@@ -0,0 +1,133 @@
...
+        assertEquals(
+            TEST_METHOD_COMP_LEVEL, whiteBox.getMethodCompilationLevel(testMethodRef),
+            "Unexpected pre-C/R compilation level"
+        );
+
+        final var resource = new Resource() {
+            public long restoreFinishTimeMs = -1;
+
+            @Override
+            public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {}
```
rvansa: total nitpick: you can remove the `throws Exception`...

test/jdk/jdk/crac/recompiler/RecompilationDelayTest.java
```diff
@@ -0,0 +1,133 @@
...
+                    restoreFinishTimeMs = Math.floorDiv(System.nanoTime(), 1_000_000);
+                }
+            }
+        };
+        Core.getGlobalContext().register(resource);
+
+        Core.checkpointRestore();
+
+        if (delayMs > 0) {
+            assertEquals(
```
rvansa: 10 seconds sounds like enough, but ideally I would have tests as independent to timing as possible. Are there any timestamps in the compilation log so rather than checking the current time we could just read the logs and run assertions on those?

rvansa (commented):
Great work! We shall give some thought to the possible timing issues in the tests, as timing-related failures in the testsuite are really annoying.

src/hotspot/share/runtime/cracRecompiler.cpp
TimPushkin: That was actually to cast-away `volatile` because it should not be needed under the lock. I am now rewriting this while fixing https://github.com/openjdk/crac/pull/251#discussion_r2235008362 so it will disappear.

test/jdk/jdk/crac/recompiler/NaturalDecompilationTest.java
TimPushkin: I am not sure what kind of false positives can there be, as I see it, in worst case we will just wait longer then necessary.

WhiteBox API indeed does not provide a way to wait for a compilation. I could probably indeed make the parent process read `PrintCompilation` from stdout and print to stdin but I am not sure this is needed as it is more complicated.

rvansa: My bad, I misread this for failing when the method is not compiled but this is just a busy loop (and it won't fail if the first 2000 executions won't trigger recompilation). All is good, then.

src/hotspot/share/runtime/cracRecompiler.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+  if (!Atomic::load_acquire(&is_recording)) {
+    return; // Fast pass to not acquire a lock when no C/R occurs (i.e. most of the time)
+  }
+
+  const MutexLocker ml(decompilations_lock, Mutex::_no_safepoint_check_flag);
+  if (is_recording) { // Re-check under the lock to be safe from concurrent changes
+    assert(decompilations != nullptr, "unexpected state: is_recording = %s, decompilations = %p",
+           BOOL_TO_STR(is_recording), decompilations);
+    decompilations->append(new CompilationInfo(nmethod.method(),
```
rvansa, line 196: Would it be problematic to make `CompilationInfo` copy-able, rather than allocating each element individually?

test/jdk/jdk/crac/recompiler/RecompilationDelayTest.java
TimPushkin: I don't think this particular assert can be removed even if we would read logs.

It asserts that the machine had time to deoptimize the method before the delay expired. Because if we do it after the delay expires it is expected that there will be no recompilation. Reading logs won't free us from this assert.

The only thing I can come up with is to implement some mechanism to relaunch the test (or just the restore part if I implement the request above to make recompilation delay restore-settable) with a higher delay when the test fails this way.

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
TimPushkin: There are two properties now:
- `jdk.crac.enable-recompilation` — setting it to false enables the old behavior, it is static final because letting it change at runtime would complicate things and I don't think it is so much needed
- `jdk.crac.recompilation-delay-ms` — same as before but can be changed at runtime, a warning is printed if it is set when `jdk.crac.enable-recompilation` is false

src/hotspot/share/runtime/cracRecompiler.cpp
TimPushkin: It can be done but `_klass_holder` field complicates things. As I see it, it will be either inefficient (fill `_klass_holder` in the temporary, then do it again in `_klass_holder` — an excess allocation of a weak handle) or ugly (don't initiate `_klass_holder` by default, add a method to initiate it, call it on the non-temporary — until this method is called the object is in a weird semi-initialized state). I would prefer to just leave it as is.

It is sad that `GrowableArray` and HotSpot in general do not support move semantics...

test/jdk/jdk/crac/recompiler/RecompilationDelayTest.java
rvansa: So this is actually triggering the deoptimization (yes, the method sounds like that, but the error message sounds like it's just checking some statistics)? If the purpose of the test is to check that recompilations won't happen too soon, why is there a difference in the time when you trigger the deoptimization? (for delayMs == 0 this must be in the afterRestore, but why wouldn't you trigger it there even for delayMs > 0?)

TimPushkin: > So this is actually triggering the deoptimization

Yes, it triggers the deoptimization. It returns the amount of compilations actually deoptimized (can be zero if the method was not compiled; can be multiple for one method if there were OSR compilations with different entry byte codes, although I believe this overload skips OSRs).

> but why wouldn't you trigger it there even for delayMs > 0

Because then the test would succeed even if the delaying code gets removed, i.e. it won't actually be testing that the delay works.

TimPushkin: It would be easier to test if after-restore recompilations would happen immediately instead of after the delay expires (then I could set a very large delay but after triggering the decompilation just wait a bit until it gets recompiled instead of having to wait until the delay expires). But that turned out to not be easy to implement (threads triggering decompilations sometimes cannot enqueue compilations), so it is something to maybe implement in the future.

rvansa: > Because then the test would succeed even if the delaying code gets removed, i.e. it won't actually be testing that the delay works.

And can't you test `whiteBox.getMethodCompilationLevel()` and ensure that the method is still deoptimized? That'd be prone to timing as well, but you could check timestamp after the level check and invalidate it if it's too late...

TimPushkin: IIUC that would be essentially the same thing just tested the other way: we are still checking that the delay is long enough for this machine. And we would still need the timestamp check to check that the check itself did not happen too late, so there will still be two checks.

Also I consider optimizing the implementation in the future to recompile things recorded during C/R right after restore and start a new recording for the delay (this would recompile during-C/R decompilations earlier potentially improving the performance) — such implementation would require the test to be as it is now.

PR #255, opened 2025-08-01
8364538: [CRaC] Test CRaCIgnoreRestoreIfUnavailable

TimPushkin:
Adds tests for `CRaCIgnoreRestoreIfUnavailable`.

The case for incompatible CPU features is not tested, although I believe it should work. Such test would be more complex (to the best of my knowledge, emulation is needed to be able to reduce the set of available features between checkpoint and restore) so it deserves a separate task. It would be something like what `test/jdk/jdk/crac/CPUFeatures/CPUFeatures.sh` does (have never run it, not sure if it works).

---------
### Progress

### Issue
 * [JDK-8364538](https://bugs.openjdk.org/browse/JDK-8364538): [CRaC] Test CRaCIgnoreRestoreIfUnavailable (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/255/head:pull/255` \
`$ git checkout pull/255`

`$ git checkout pull/255` \
`$ git pull https://git.openjdk.org/crac.git pull/255/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 255`

`$ git pr show -t 255`

</details>

<a href="https://git.openjdk.org/crac/pull/255.diff">https://git.openjdk.org/crac/pull/255.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/255#issuecomment-3144751951)
</details>

### DIFF
--- a/test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
+++ b/test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025, Azul Systems, Inc. All rights reserved.
+ * Copyright (c) 2025, Azul Systems, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
+++ b/test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
@@ -23,10 +23,11 @@
  * questions.
  */
 
-import java.nio.file.Files;
 import java.util.List;
 
+import jdk.crac.Core;
 import jdk.test.lib.crac.CracBuilder;
+import jdk.test.lib.crac.CracEngine;
 
 /**
  * @test
@@ -45,23 +46,30 @@ public class EngineFailureTest {
         // supposed to do that, JVM does it automatically when needed) and do
         // not provide the mandatory 'exec_location' option.
 
-        final var builder = new CracBuilder()
+        // Create an empty image which will pass VM's own checks. It may contain
+        // CPU features, for example.
+        new CracBuilder().engine(CracEngine.SIMULATE)
+            .main(Main.class).args("true")
+            .startCheckpoint().waitForSuccess();
+
+        new CracBuilder()
             .vmOption("-XX:CRaCEngine=crexec")
             .vmOption("-XX:+CRaCIgnoreRestoreIfUnavailable")
             .forwardClasspathOnRestore(true)
-            .captureOutput(true);
-
-        // Make VM's internal image checks pass
-        Files.createDirectory(builder.imageDir());
-
-        builder.startRestoreWithArgs(null, List.of(Main.class.getName())).outputAnalyzer()
+            .captureOutput(true)
+            .startRestoreWithArgs(null, List.of(Main.class.getName(), "false"))
+            .outputAnalyzer()
             .shouldContain("CRaC engine failed to restore")
             .shouldContain(MAIN_MSG);
     }
 
     public static class Main {
-        public static void main(String[] args) {
-            System.out.println(MAIN_MSG);
+        public static void main(String[] args) throws Exception {
+            if (Boolean.parseBoolean(args[0])) {
+                Core.checkpointRestore();
+            } else {
+                System.out.println(MAIN_MSG);
+            }
         }
     }
 }
--- a/test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
+++ b/test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2022, 2025, Azul Systems, Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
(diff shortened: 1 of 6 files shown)


test/jdk/jdk/crac/ignoreRestore/RestoreIfPossibleTest.java
```diff
@@ -0,0 +1,78 @@
...
+            final var checkpointProcess = builder.main(Main.class).args("true").startCheckpoint();
+            checkpointProcess.waitForCheckpointed();
+            final var out = checkpointProcess.outputAnalyzer();
+            out.stdoutShouldContain(WARMUP_MSG).stdoutShouldNotContain(MAIN_MSG);
+        } else {
+            FileUtils.deleteFileTreeWithRetry(builder.imageDir()); // Existance depends on the order of @run tags
+        }
+
+        final var out = builder.startRestoreWithArgs(null, List.of(Main.class.getName(), "false")).outputAnalyzer();
+        out.stdoutShouldNotContain(WARMUP_MSG).stdoutShouldContain(MAIN_MSG);
```
TimPushkin, line 65: Currently `MAIN_MSG` will actually be printed twice: once from newly launched main (because we provided new arguments on restore) and once from the original main.

Such behavior is inconvenient, especially when `-jar <jar-path>` is used (because new args feature does not support this yet and fails). It will be fixed once we add something like `-XX:+CRaCExecuteNewMain` required to execute new args:
- By default `CRaCExecuteNewMain` is false and restore fails if new args are provided (unless `-XX:+CRaCIgnoreRestoreIfUnavailable` is provided, then they should be ignored for restore but used if restore fails for another reason)
- With `-XX:+CRaCExecuteNewMain` new args are mandatory

With `-XX:+CRaCIgnoreRestoreIfUnavailable -XX:+CRaCExecuteNewMain` the behavior will be as it is now with just `-XX:+CRaCIgnoreRestoreIfUnavailable`.

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
```diff
@@ -45,23 +46,30 @@ public static void main(String[] args) throws Exception {
...
         // not provide the mandatory 'exec_location' option.
 
-        final var builder = new CracBuilder()
+        // Create an empty image which will pass VM's own checks. It may contain
+        // CPU features, for example.
+        new CracBuilder().engine(CracEngine.SIMULATE)
+            .main(Main.class).args("true")
+            .startCheckpoint().waitForSuccess();
+
+        new CracBuilder()
```
rvansa, line 55: The `crexec` option for `CRaCEngine` is weird; looks like a leftover from times when we had `CREngineLibrary` as well. Is it supposed to execute CRIU, or completely fail the restore, leading to 'unavailable'?

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
```diff
@@ -45,23 +46,30 @@ public static void main(String[] args) throws Exception {
         // supposed to do that, JVM does it automatically when needed) and do
         // not provide the mandatory 'exec_location' option.
 
-        final var builder = new CracBuilder()
+        // Create an empty image which will pass VM's own checks. It may contain
+        // CPU features, for example.
+        new CracBuilder().engine(CracEngine.SIMULATE)
```
rvansa, line 51: OK, this creates the empty directory (or rather one with just features), but what's the point? Just making the 'restore' below fail at a different point of time? If there are two ways of failure, both should be tested.

TimPushkin: Yes, if there are no CPU features — that is a different failure.

Did not want to add a test for that because I am not testing CPU features in this PR, but this particular one should not be hard to implement so I'll add it.

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
TimPushkin: There is a long comment above explaining why this is done like this. Basically we need the failure to happen before `exec`ing the engine executable because there is no return to the parent VM from there.

TimPushkin: If we had some library-based engine besides `crexec` I would use that

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
TimPushkin: Added  `NoCPUFeaturesTest`

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
rvansa: Could we use `-XX:CRaCEngineOptions=fail-with=unknown-option` instead? Or even `-XX:CRaCEngine=non-existent`?

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
rvansa: I was not implying that you should add the test for missing CPU features, but good to have one.

My point is that running `simengine` alone should be a no-op (yes, you're testing that simengine works, but that's not a point of this test). The comment says `Create an empty image` but simengine is not supposed to create an empty image, it is not supposed to do anything. As a side-effect, JVM creates the directory.

My understanding is that you are doing that to prevent checks before entering the engine to fail in a different manner than the test is supposed to do. I consider running the test with two different engines a confusing, obscure way to get the prerequisites; you could just `mkdir` and write a dummy CPU features file.

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
TimPushkin: No, these are both VM-side failures, I could add tests for those but they are all currently handled in the same way w.r.t. `CRaCIgnoreRestoreIfUnavailable` so I don't see a point.

This test is supposed to check a failure that is detected inside the `CRaCEngine::restore()` call, inside the engine library, and makes the call return.

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
TimPushkin: Writing a dummy CPU features file would make the test depend on the format of the file, i.e. both on crexec and CPU features implementation.

The only way I see to make the test less artificial is to create a library-based simengine.

test/jdk/jdk/crac/ignoreRestore/EngineFailureTest.java
rvansa: OK, if you're saying there's no better way let's keep it as is.

TimPushkin:
Thank you for the review.

I believe it is only `EngineFailureTest.java` you don't like but yes, I don't see ways to improve it until we add a library-based engine to OpenJDK.

PR #256, opened 2025-08-01
8364554: [CRaC] Improve CRaC VM options' docs

TimPushkin:
Besides some wording and formatting of VM options:
- Renames `CRaCAllowToSkipCheckpoint` to `CRaCSkipCheckpoint` since it makes the VM always skip checkpoints and not only in some cases
- Replaces `CRaCMaxHeapSizeBeforeCheckpoint == 0` with `FLAG_IS_DEFAULT(CRaCMaxHeapSizeBeforeCheckpoint)` because the docs 0 do not specify 0 as a special value

---------
### Progress

### Issue
 * [JDK-8364554](https://bugs.openjdk.org/browse/JDK-8364554): [CRaC] Improve CRaC VM options' docs (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/256/head:pull/256` \
`$ git checkout pull/256`

`$ git checkout pull/256` \
`$ git pull https://git.openjdk.org/crac.git pull/256/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 256`

`$ git pr show -t 256`

</details>

<a href="https://git.openjdk.org/crac/pull/256.diff">https://git.openjdk.org/crac/pull/256.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/256#issuecomment-3145560970)
</details>

### DIFF
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -577,6 +577,7 @@ static SpecialFlag const special_jvm_flags[] = {
   { "CRDoThrowCheckpointException", JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
   { "CRPauseOnCheckpointError",     JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
   { "CRTrace",                      JDK_Version::undefined(), JDK_Version::jdk(24), JDK_Version::undefined() },
+  { "CRaCAllowToSkipCheckpoint",    JDK_Version::jdk(25), JDK_Version::undefined(), JDK_Version::undefined() },
   { "CRaCDoThrowCheckpointException", JDK_Version::undefined(), JDK_Version::jdk(25), JDK_Version::undefined() },
 
   { nullptr, JDK_Version(0), JDK_Version(0) }
@@ -592,7 +593,7 @@ static AliasedFlag const aliased_jvm_flags[] = {
   { "CreateMinidumpOnCrash",    "CreateCoredumpOnCrash" },
   { "CREngine",                        "CRaCEngine" },
   { "CRAllowedOpenFilePrefixes",       "CRaCAllowedOpenFilePrefixes" },
-  { "CRAllowToSkipCheckpoint",         "CRaCSkipCheckpoint "},
+  { "CRAllowToSkipCheckpoint",         "CRaCSkipCheckpoint"},
   { "CRHeapDumpOnCheckpointException", "CRaCHeapDumpOnCheckpointException" },
   { "CRPrintResourcesOnCheckpoint",    "CRaCPrintResourcesOnCheckpoint" },
   { "CRTraceStartupTime",              "CRaCTraceStartupTime" },
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -568,17 +568,17 @@ static SpecialFlag const special_jvm_flags[] = {
   { "dup option",                   JDK_Version::jdk(9), JDK_Version::undefined(), JDK_Version::undefined() },
 #endif
 
-  { "CREngine",                     JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRAllowedOpenFilePrefixes",    JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRAllowToSkipCheckpoint",      JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRHeapDumpOnCheckpointException", JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRPrintResourcesOnCheckpoint", JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRTraceStartupTime",           JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRDoThrowCheckpointException", JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRPauseOnCheckpointError",     JDK_Version::jdk(24), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRTrace",                      JDK_Version::undefined(), JDK_Version::jdk(24), JDK_Version::undefined() },
-  { "CRaCAllowToSkipCheckpoint",    JDK_Version::jdk(25), JDK_Version::undefined(), JDK_Version::undefined() },
-  { "CRaCDoThrowCheckpointException", JDK_Version::undefined(), JDK_Version::jdk(25), JDK_Version::undefined() },
+  { "CREngine",                     JDK_Version::jdk(24), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRAllowedOpenFilePrefixes",    JDK_Version::jdk(24), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRAllowToSkipCheckpoint",      JDK_Version::jdk(24), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRHeapDumpOnCheckpointException", JDK_Version::jdk(24), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRPrintResourcesOnCheckpoint", JDK_Version::jdk(24), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRTraceStartupTime",           JDK_Version::jdk(24), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRDoThrowCheckpointException", JDK_Version::jdk(24), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRPauseOnCheckpointError",     JDK_Version::jdk(24), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRTrace",                      JDK_Version::undefined(), JDK_Version::jdk(24), JDK_Version::jdk(26) },
+  { "CRaCAllowToSkipCheckpoint",    JDK_Version::jdk(25), JDK_Version::jdk(26), JDK_Version::jdk(27) },
+  { "CRaCDoThrowCheckpointException", JDK_Version::undefined(), JDK_Version::jdk(25), JDK_Version::jdk(26) },
 
   { nullptr, JDK_Version(0), JDK_Version(0) }
 };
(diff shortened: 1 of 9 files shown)


src/java.base/share/man/java.md
```diff
@@ -1075,18 +1075,19 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
...
     created.
 
 `-XX:CRaCRestoreFrom=`*directory*
-:   Restores a snapshot created by `-XX:CRaCCheckpointTo=`*directory*.
+:   Restores an image created by `-XX:CRaCCheckpointTo`.
 
 `-XX:CRaCMinPid=`*value*
-:   A desired minimal PID value for checkpoint'ed process. Ignored on restore.
+:   A desired minimal PID value for checkpointed process. Used by the launcher,
+    only on POSIX-like platforms.
```
TimPushkin: BTW, I thought about removing `CRaCMinPid` from `globals.hpp` since it is not used by the VM at all but decided against it (too big of a change for a pretty general PR like this) and just documented that it is to be used by the launcher.

Since it is (or should be) possible to use JVM without the default launcher we could add a check in VM that the option is honored and at least warn if it is not.

Internally we've had similar problems with Tomcat launching JVM directly and thus not honoring options implemented purely in the launcher.

rvansa: I don't think we should remove; it should be normally listed with `-XX:PrintFlagsFinal` etc.

Regarding wording, I don't think that "launcher" is a term that we should use in user-facing documentation.

TimPushkin: This term is used a lot in this file, e.g. "To launch a class declared in a source file, run the `java` launcher in
source-file mode"

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1959,69 +1959,72 @@ const int ObjectAlignmentInBytes = 8;
              "fence. Add cleanliness checks.")                              \
                                                                             \
   product(ccstr, CRaCCheckpointTo, nullptr, RESTORE_SETTABLE,               \
-        "Path to checkpoint image directory")                               \
+          "Path to place the checkpoint image into")                        \
```
rvansa: I think it is still important to state that the path leads to a directory.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1959,69 +1959,72 @@ const int ObjectAlignmentInBytes = 8;
...
                                                                             \
   product(ccstr, CRaCRestoreFrom, nullptr, RESTORE_SETTABLE,                \
-      "Path to image for restore, replaces the initializing VM on success") \
+          "Path to the image to restore from. If restore succeeds, the "    \
```
rvansa: In case of CRIU it is not really replaced, the restored VM is a child process of launcher stub...

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1959,69 +1959,72 @@ const int ObjectAlignmentInBytes = 8;
...
+          "initializing VM is replaced with the restored one.")             \
                                                                             \
   product(uint, CRaCMinPid, 128,                                            \
-      "Mininal PID value for checkpoint'ed process")                        \
-      range(1, UINT_MAX)                                                    \
+          "Mininal PID value for checkpointed process (Unix only)")         \
```
rvansa: `Unix` -> `POSIX` (same as in java.md). Linux is not Unix.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1959,69 +1959,72 @@ const int ObjectAlignmentInBytes = 8;
...
-      "for files opened by Java code use -Djdk.crac.resource-policies=...") \
+          "prefixes for files that can be open during checkpoint. CRaC "    \
+          "will not error upon detecting these and will leave the handling "\
+          "up to the engine. This option applies only to files opened by "  \
+          "native code and is Linux only; for files opened by Java code "   \
+          "use -Djdk.crac.resource-policies=...")                           \
                                                                             \
-  product(bool, CRaCAllowToSkipCheckpoint, false, DIAGNOSTIC,               \
-      "Allow implementation to not call Checkpoint if helper not found")    \
+  product(bool, CRaCSkipCheckpoint, false, DIAGNOSTIC,                      \
```
rvansa, line 2008: Shouldn't we rather deprecate/obsolete the option altogether? Sounds like a predecessor to `simengine`, that is the preferred way. Looking into the implementation it seems that this will keep `jcmd` hanging if invoked from that, won't it?

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1959,69 +1959,72 @@ const int ObjectAlignmentInBytes = 8;
...
+          "Dump heap on CheckpointException thrown because of a low-level " \
+          "CRaC precondition failing")                                      \
                                                                             \
   product(bool, CRaCPrintResourcesOnCheckpoint, false, DIAGNOSTIC,          \
-      "Print resources to decide CheckpointException")                      \
+          "Print low-level resources checked by VM before checkpoint")      \
                                                                             \
   product(bool, CRaCTraceStartupTime, false, DIAGNOSTIC,                    \
-      "Trace startup time")                                                 \
+          "Trace restore startup time")                                     \
```
rvansa: So "restore" or "startup"? :-) What about

> Print timestamp after successful restore by CRaC engine

In general, I think that this should be turned into unified logging configuration option. It's also weird that it is not restore-settable, you have to set that on checkpoint...

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1959,69 +1959,72 @@ const int ObjectAlignmentInBytes = 8;
...
                                                                             \
   product(bool, CRaCDoThrowCheckpointException, true, EXPERIMENTAL,         \
-      "Throw CheckpointException if uncheckpointable resource handle found")\
+          "Throw CheckpointException if non-checkpointable low-level "      \
```
rvansa: I don't really see what's the purpose, and there's no test. Since it's experimental, can we obsolete it righ away?

src/java.base/share/man/java.md
```diff
@@ -1075,18 +1075,19 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
...
+    lower start-up and warm-up times compared to a full re-launch.
 
-    This option defines a path to the snapshot which is currently a directory. The
+    This option defines a path to the image which is currently a directory. The
     directory will be created if it does not exist, but no parent directories are
     created.
 
 `-XX:CRaCRestoreFrom=`*directory*
-:   Restores a snapshot created by `-XX:CRaCCheckpointTo=`*directory*.
+:   Restores an image created by `-XX:CRaCCheckpointTo`.
```
rvansa: The "by" is a bit out of place, since the option does not immediately trigger creation of the image. How about
> Restores an image written on checkpoint into `-XX:CRaCCheckpointTo` 

?

src/hotspot/share/runtime/globals.hpp
TimPushkin: I agree but I am following the style of the file. In this file "(Unix only)" is used for 3 other options while "posix" is not used at all.

src/hotspot/share/runtime/globals.hpp
TimPushkin: I used it recently when doing performance investigations: it allowed to rule out influence of the engine on checkpoint/restore performance. So it still can be useful. But I agree, some tests wouldn't harm.

> it seems that this will keep jcmd hanging if invoked from that, won't it?

No, the connection is closed after the JCMD operation is finished, as for any other JCMD operation. `report_ok_to_jcmd_if_any` is needed when we use an engine to close the socket before checkpointing.

src/hotspot/share/runtime/globals.hpp
TimPushkin: Will improve the wording. The name can also be improved but I'll leave it as is for now since maybe the option should be deprecated (see below).

> It's also weird that it is not restore-settable, you have to set that on checkpoint

We cannot really make it restore-settable because the timestamp is printed before restore-settable options are read. Unless we record the timestamp to print it later.

We also have `jdk.crac.trace-startup-time` property used for the same purpose on the Java side (also cannot be changed on restore, although it could be) — we should probably unify these deprecating one of them. If they get unified the VM-level one should continue without the unified logging so that it prints to the same destination as the Java-level one.

Anyway, I'll open a separate task for this as there are things to be thought through.

src/hotspot/share/runtime/arguments.cpp
TimPushkin, line 1: BTW maybe it is time to put expiration dates for all of CRaC's deprecated/obsoleted options? For each parsed argument there are linear searches through lists of special (deprecated/obsoleted/expired) and aliased flags `strcmp`-ing them which slows down argument parsing.

Since most of them have been deprecated since winter we can put 26 as the expiration version: it will be a year since the options were deprecated. For the ones deprecated/obsoleted today it can be 27.

src/hotspot/share/runtime/globals.hpp
rvansa: Understood, but that doesn't make it correct. Linux is still not Unix. Btw., should we move this to `globals_linux.hpp`?

TimPushkin: > Understood, but that doesn't make it correct. Linux is still not Unix.

Ok, will redact.

> Btw., should we move this to `globals_linux.hpp`?

No, it is used on all platforms but Windows, and there is no `globals_posix.hpp`.

src/hotspot/share/runtime/arguments.cpp
TimPushkin: Ended up expiring everything since 26: the options added to special in this PR are diagnostic/experimental and not widely used, so this is in accordance with OpenJDK's recommendations.

rvansa: Expiration by 26 is probably fine, but it seems that you're obsoleting retroactively? There is no 24-based version that would obsolete those, this will be ~25 and you're obsoleting on the spot. Except for `CRaCDoThrowCheckpointException` I would make these defunct (through expiration) only in 26.

TimPushkin: Oh, I really did not intend to do that. I was fighting version correctness asserts and misread the guidelines in `arguments.cpp`.

The guidelines state: "To change the name of an option, use the alias table as well as a 2-step model adding major release numbers to the deprecate and expire columns", but this seems to not be possible because asserts mandate that:
- it is not allowed to only fill deprecate and expire columns, obsolete must also be filled: "<flag> must be explicitly obsoleted before expired";
- it is not allowed to set obsolete and expire columns to the same version, expire must be strictly greater: "<flag> must be obsoleted before expired".

So only a 3-step model is possible. Since we do not want to obsolete on the spot for our aliases we can only either:
- If using only major versions, obsolete in 26 and expire in 27 (for LTS releases: April and October 2026)
- If using minor versions too, obsolete in 26 and expire in 26.0.2 (for LTS releases: April and July 2026)
   - We do not want to obsolete before 26 because that would be less than a year since the deprecation of frequently-used aliases
   - To correspond to LTS versions, we cannot obsolete in 26 and expire in 26.0.1 because in LTS that would be the same April 2026 release and we'll get assert failures because obsolete == expire

I do not have a preference: the first one looks cleaner, the second one makes the aliases expire sooner. Went with the first one just to choose something.

For `CRTrace` and `CRaCDoThrowCheckpointException` it is possible to expire in 26 so I did that.

rvansa: Let's stick to major versions, that's really cleaner. We don't really need to rush it.

PR #257, opened 2025-08-04
8364624: [CRaC] Resolve startup failures with static JDK build

dimitryc:
The patch fixes locating the `lib` folder and also fixes handling for `crexec` case for static JDK build.

It also fixes "setup CRIU" test step and ensures that criu is copied into static JDK for static JDK build. Another change in tests is to re-enable tests for static JDK and add some CRaC tests with known failures to the exclude list for static JDK (tests requiring jcmd and tests requiring a container with libX11 preinstalled).

### DIFF
--- a/test/jdk/ProblemList-StaticJdk.txt
+++ b/test/jdk/ProblemList-StaticJdk.txt
@@ -12,3 +12,16 @@ java/util/ResourceBundle/modules/unnamed/UnNamedTest.java                       
 
 # Require jps
 java/util/concurrent/locks/Lock/TimedAcquireLeak.java                            8346719 generic-all
+
+# Require jcmd
+jdk/crac/DaemonAfterRestore.java                                                 8346719 generic-all
+jdk/crac/fileDescriptors/SharedLibraryTest.java                                  8346719 generic-all
+jdk/crac/jfr/FlightRecorderJcmdTest.java                                         8346719 generic-all
+jdk/crac/PerfMemoryRestoreTest.java                                              8346719 generic-all
+
+# Require container with libX11
+jdk/crac/ContainerOOMETest.java                                                  8364624 generic-all
+jdk/crac/ContainerPidAdjustmentTest.java                                         8364624 generic-all
+jdk/crac/java/lang/System/NanoTimeTest.java                                      8364624 generic-all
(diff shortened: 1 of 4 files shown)


test/jdk/ProblemList-StaticJdk.txt
```diff
@@ -12,3 +12,16 @@ java/util/ResourceBundle/modules/unnamed/UnNamedTest.java
...
+jdk/crac/fileDescriptors/SharedLibraryTest.java                                  8346719 generic-all
+jdk/crac/jfr/FlightRecorderJcmdTest.java                                         8346719 generic-all
+jdk/crac/PerfMemoryRestoreTest.java                                              8346719 generic-all
+
+# Require container with libX11
+jdk/crac/ContainerOOMETest.java                                                  8364624 generic-all
+jdk/crac/ContainerPidAdjustmentTest.java                                         8364624 generic-all
+jdk/crac/java/lang/System/NanoTimeTest.java                                      8364624 generic-all
+jdk/crac/java/net/InetAddress/ResolveTest.java                                   8364624 generic-all
+jdk/crac/java/lang/System/TimedWaitingTest.java                                  8364624 generic-all
```
TimPushkin, line 27: Could you please explain what is the problem with these? The tests themselves should not require X11.

TimPushkin (approved):
LGTM overall, thanks, just one question

test/jdk/ProblemList-StaticJdk.txt
dimitryc: The tests fail with following error

> /jdk/bin/java: error while loading shared libraries: libX11.so.6: cannot open shared object file: No such file or directory 

It seems that `libawt_xawt.so` has the dependency on `libX11.so.6` and when it's statically linked, `java` executable has the dependency.

TimPushkin: I wonder why none of the tests from the mainline failed since some of these tests use the default base image which others should also use. Maybe tests using Docker are not included into the tiers run in CI...

dimitryc: Yes, the tests using Docker from mainline aren't included in tiers that run in CI. I haven't checked this but if they had been running, they most likely would have failed with a similar error.

PR #258, opened 2025-08-04
8364662: [CRaC] Process restore options in the common arguments parser

TimPushkin:
Options for restore are now processed during the common VM arguments parsing stage.

Notable improvements compared to the old code:
- Options from `-XX:Flags=<file>` and `-XX:VMOptionsFile=<file>` are now also considered
- Option aliasing is now handled correctly

**Behavioral change**. If `CRaCIgnoreRestoreIfUnavailable` is true it is now not an error to specify non-restore-settable options on restore — these will just be omitted from forwarding to the restored JVM. This makes it possible to specify non-restore-settable to be used after a failed restore.

---------
### Progress

### Issue
 * [JDK-8364662](https://bugs.openjdk.org/browse/JDK-8364662): [CRaC] Process restore options in the common arguments parser (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/258/head:pull/258` \
`$ git checkout pull/258`

`$ git checkout pull/258` \
`$ git pull https://git.openjdk.org/crac.git pull/258/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 258`

`$ git pr show -t 258`

</details>

<a href="https://git.openjdk.org/crac/pull/258.diff">https://git.openjdk.org/crac/pull/258.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/258#issuecomment-3151943142)
</details>

### DIFF
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -2945,8 +2945,8 @@ bool Arguments::process_flag_for_restore(const char *arg) {
 
   const char* name;
   size_t name_len;
-  bool ignored_plus_minus;
-  parse_argname(arg, &name, &name_len, &ignored_plus_minus);
+  bool has_plus_minus;
+  parse_argname(arg, &name, &name_len, &has_plus_minus);
 
   const JVMFlag* flag;
   {
@@ -2963,6 +2963,7 @@ bool Arguments::process_flag_for_restore(const char *arg) {
     flag = JVMFlag::find_declared_flag(real_flag_name(stripped_name));
     guarantee(flag != nullptr, "unknown JVM flag name: %s", name); // Should've been detected earlier
   }
+  assert(has_plus_minus == flag->is_bool(), "sanity check");
 
   if (flag->is_restore_settable()) {
     // Restored JVM will search for the flag using the name we record here so we
@@ -2970,12 +2971,14 @@ bool Arguments::process_flag_for_restore(const char *arg) {
     if (strncmp(name, flag->name(), name_len) == 0) {
       build_jvm_restore_flags(arg);
     } else {
-      const size_t real_tail_len = strlen(arg) + (strlen(flag->name()) - name_len);
-      char* const real_tail = AllocateHeap(real_tail_len + 1, MemTag::mtArguments);
-      const int ret = jio_snprintf(real_tail, real_tail_len + 1, "%s%s", flag->name(), arg + name_len);
-      guarantee(ret >= 0 && checked_cast<size_t>(ret) == real_tail_len, "snprintf failed: %i", ret);
-      build_jvm_restore_flags(real_tail);
-      FreeHeap(real_tail);
+      const size_t real_arg_len = strlen(arg) + (strlen(flag->name()) - name_len);
+      char* const real_arg = AllocateHeap(real_arg_len + 1, MemTag::mtArguments);
+      const int ret = has_plus_minus ?
+        jio_snprintf(real_arg, real_arg_len + 1, "%c%s", arg[0], flag->name()) :
+        jio_snprintf(real_arg, real_arg_len + 1, "%s%s", flag->name(), arg + name_len);
+      guarantee(ret >= 0 && checked_cast<size_t>(ret) == real_arg_len, "snprintf failed: %i", ret);
+      build_jvm_restore_flags(real_arg);
+      FreeHeap(real_arg);
     }
   } else if (!CRaCIgnoreRestoreIfUnavailable) {
     // Same message format as used in JVMFlag::get_locked_message() for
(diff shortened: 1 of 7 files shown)


src/hotspot/share/runtime/crac.cpp
```diff
@@ -632,6 +632,11 @@ bool CracRestoreParameters::read_from(int fd) {
       } else {
         *eq = '\0';
         char* value = eq + 1;
+        // A single ccstrlist flag can be specified multiple times meaning those
+        // should be concatenated. But with the current code the last occurence
+        // will just overwrite the previous ones.
+        assert(!JVMFlag::find_flag(cursor)->ccstr_accumulates(),
```
rvansa, line 638: Ouch! I think that using multiple `-XX:CRaCEngineOptions` would be pretty common (at least some of my scripts do that). Could we fix this?

And this error should be guarantee/another check that is applied in release build, too because it verifies user input, not some invariants in internal components.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -1041,6 +1043,10 @@ void Arguments::build_jvm_flags(const char* arg) {
   add_string(&_jvm_flags_array, &_num_jvm_flags, arg);
 }
 
+void Arguments::build_jvm_restore_flags(const char* arg) {
```
rvansa, line 1046: I would use `append_jvm_restore_flags` - `build` implies the final operation on a builder.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -3006,6 +2935,87 @@ void Arguments::fix_appclasspath() {
...
+
+  if (flag->is_restore_settable()) {
+    // Restored JVM will search for the flag using the name we record here so we
+    // must ensure the real one is recorded
+    if (strncmp(name, flag->name(), name_len) == 0) {
+      build_jvm_restore_flags(arg);
+    } else {
+      const size_t real_tail_len = strlen(arg) + (strlen(flag->name()) - name_len);
+      char* const real_tail = AllocateHeap(real_tail_len + 1, MemTag::mtArguments);
+      const int ret = jio_snprintf(real_tail, real_tail_len + 1, "%s%s", flag->name(), arg + name_len);
```
rvansa: Where are `+BoolOpt` flags handled?

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -3006,6 +2935,87 @@ void Arguments::fix_appclasspath() {
...
+      stripped_name = buf;
+    }
+    flag = JVMFlag::find_declared_flag(real_flag_name(stripped_name));
+    guarantee(flag != nullptr, "unknown JVM flag name: %s", name); // Should've been detected earlier
+  }
+
+  if (flag->is_restore_settable()) {
+    // Restored JVM will search for the flag using the name we record here so we
+    // must ensure the real one is recorded
+    if (strncmp(name, flag->name(), name_len) == 0) {
```
rvansa: If `name` is a prefix of `flag->name()` this will pass - check `name_len == strlen(flag->name())` first

src/hotspot/share/runtime/crac.cpp
TimPushkin: This is not an issue for `CRaCEngineOptions` because it is not applied in the restored process (you can see it is handled in another branch above).

Currently there are no `ccstrlist` restore-settables actually being applied in the restored process, that is why I just added this assert for now. Its purpose is to fail when we add such option.

src/hotspot/share/runtime/arguments.cpp
TimPushkin: I follow the style of the class: there are `Arguments::build_jvm_args` and `Arguments::build_jvm_flags` which do exactly the same thing, just for two different arrays.

src/hotspot/share/runtime/arguments.cpp
TimPushkin: Oh, actually there are no aliased restore-settable boolean options so they don't reach this branch and are not handled. But this is not intentional, I just forgot about them and there are no tests since such things don't exist... I should fix this anyway, thanks for pointing out.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -3006,6 +2935,90 @@ void Arguments::fix_appclasspath() {
...
+    }
+    flag = JVMFlag::find_declared_flag(real_flag_name(stripped_name));
+    guarantee(flag != nullptr, "unknown JVM flag name: %s", name); // Should've been detected earlier
+  }
+  assert(has_plus_minus == flag->is_bool(), "sanity check");
+
+  if (flag->is_restore_settable()) {
+    // Restored JVM will search for the flag using the name we record here so we
+    // must ensure the real one is recorded
+    if (strcmp(flag->name(), name) == 0) {
```
rvansa: IIUC the `name` can still contain the equals sign and value? Maybe if you used the `stripped_name`...

rvansa (approved):

PR #261, opened 2025-08-11
8365241: [CRaC] Ignore new arguments when CRaCIgnoreRestoreIfUnavailable is true

TimPushkin:
Arguments are not passed to the restored process when `-XX:+CRaCIgnoreRestoreIfUnavailable` is set on restore. The check is performed on the side of the restoring VM since in the restored VM `CRaCIgnoreRestoreIfUnavailable` could have been set before the current restore.

Documentation for the new arguments feature and `CRaCIgnoreRestoreIfUnavailable` is also added, explaining this behavior.

---------
### Progress

### Issue
 * [JDK-8365241](https://bugs.openjdk.org/browse/JDK-8365241): [CRaC] Ignore new arguments when CRaCIgnoreRestoreIfUnavailable is true (**Enhancement** - P1)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/261/head:pull/261` \
`$ git checkout pull/261`

`$ git checkout pull/261` \
`$ git pull https://git.openjdk.org/crac.git pull/261/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 261`

`$ git pr show -t 261`

</details>

<a href="https://git.openjdk.org/crac/pull/261.diff">https://git.openjdk.org/crac/pull/261.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/261#issuecomment-3174318184)
</details>

### DIFF
--- a/src/java.base/share/man/java.md
+++ b/src/java.base/share/man/java.md
@@ -1093,8 +1093,10 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
     of its existing call stack.
 
 `-XX:+CRaCIgnoreRestoreIfUnavailable`
-:   If restoration from a checkpoint image fails, continue with the normal startup
-    instead of failing.
+:   If the checkpoint image specified for restoration is identified as being unusable,
+    continue with the normal startup instead of failing. Enabling this option does
+    not guarantee that the normal startup will follow any failed restoration attempt;
+    it only covers an early image/environment verification.
 
     When restoring with this option, you should specify a main class and its
     arguments as usual. If the restoration succeeds, they are ignored, but if it
--- a/src/java.base/share/man/java.md
+++ b/src/java.base/share/man/java.md
@@ -1085,22 +1085,21 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
 `-XX:CRaCRestoreFrom=`*directory*
 :   Restores from the specified checkpoint image.
 
-    When using this option, there is no requirement to specify a main class and its
-    arguments because the restored execution is based on the ones specified before
-    the checkpoint. However, it is possible to pass a new main class and new
-    arguments. After a successful restoration, the new main method will be invoked
-    with the specified arguments in the thread that started the checkpoint, on top
-    of its existing call stack.
+    You can optionally pass the name of the new main class and its arguments when
+    using this option. In such a case, the new main method will be invoked with
+    the specified arguments by the thread that initiated the checkpoint, after all
+    CRaC resources have been successfully processed.
 
 `-XX:+CRaCIgnoreRestoreIfUnavailable`
-:   If the checkpoint image specified for restoration is identified as being unusable,
-    continue with the normal startup instead of failing. Enabling this option does
-    not guarantee that the normal startup will follow any failed restoration attempt;
-    it only covers an early image/environment verification.
+:   If the checkpoint image specified for restoration is identified as being
+    unusable, continue with the normal startup instead of failing.
 
-    When restoring with this option, you should specify a main class and its
-    arguments as usual. If the restoration succeeds, they are ignored, but if it
-    fails, they are used for the normal startup process.
+    When restoring with this option, you should specify the command-line arguments
+    for the normal startup process. If the restoration succeeds, they are ignored;
+    otherwise, they are used for starting up.
+
+    If an initial image verification passes, but restoration still fails, the JVM
+    will not attempt to start normally.
 
 `-XX:CRaCMinPid=`*value*
 :   A desired minimal PID value for checkpointed process. Applied by the launcher,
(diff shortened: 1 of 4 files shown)


TimPushkin:
The test failure is not related to this change

src/java.base/share/man/java.md
```diff
@@ -1085,6 +1085,21 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
...
+    the checkpoint. However, it is possible to pass a new main class and new
+    arguments. After a successful restoration, the new main method will be invoked
+    with the specified arguments in the thread that started the checkpoint, on top
+    of its existing call stack.
+
+`-XX:+CRaCIgnoreRestoreIfUnavailable`
+:   If restoration from a checkpoint image fails, continue with the normal startup
+    instead of failing.
+
+    When restoring with this option, you should specify a main class and its
```
rvansa: > you **should** specify a main class

I guess we shall make this mandatory -> you **must** sepecify a main class? Do we have a test for the case when you run plain `java -XX:CRaCRestoreFrom=... -XX:+CRaCIgnoreRestoreIfUnavailable` and the image is not present?

TimPushkin: No, there is no such test currently.

You do not always have to specify a main class since if the image exists the main won't be used. Initially I wrote "you are expected to specify" and made it shorter into this version.

jankratochvil:
Is there a use case when the passed new arguments/main *are* useful? Is there a s testcase for it? (I have been quickly searching it now but I haven't found it.)

TimPushkin:
@jankratochvil There is a demo `src/demo/share/crac/JavaCompilerCRaC` which you use as described here https://github.com/openjdk/crac/pull/16#issuecomment-1036140922.

src/java.base/share/man/java.md
```diff
@@ -1085,6 +1085,23 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
 `-XX:CRaCRestoreFrom=`*directory*
 :   Restores from the specified checkpoint image.
 
+    When using this option, there is no requirement to specify a main class and its
+    arguments because the restored execution is based on the ones specified before
+    the checkpoint. However, it is possible to pass a new main class and new
+    arguments. After a successful restoration, the new main method will be invoked
+    with the specified arguments in the thread that started the checkpoint, on top
+    of its existing call stack.
```
AntonKozlov: I'd suggest start with positive example, something like:

You can pass a new "main" class and arguments with this option. Then, a method matching `main(String[])` will be invoked then with the provided arguments after all `afterRestore` methods complete. The method will invoked on the thread which initiated the checkpoint.

src/java.base/share/man/java.md
AntonKozlov: "Should" seems to fit. It won't be a violation of the interface, JDK will handle missing arguments and will and provide the help message, for example.

src/java.base/share/man/java.md
```diff
@@ -1085,6 +1085,23 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
...
+
+`-XX:+CRaCIgnoreRestoreIfUnavailable`
+:   If the checkpoint image specified for restoration is identified as being unusable,
+    continue with the normal startup instead of failing. Enabling this option does
+    not guarantee that the normal startup will follow any failed restoration attempt;
+    it only covers an early image/environment verification.
+
+    When restoring with this option, you should specify a main class and its
+    arguments as usual. If the restoration succeeds, they are ignored, but if it
+    fails, they are used for the normal startup process.
```
AntonKozlov: Let's move the "may fail" note to the bottom. Suggested wording:

If the checkpoint image specified for restoration is identified as being unusable, continue with the normal startup instead of failing. When restoring with this option, you should specify command line arguments for a normal start-up process. If the restoration succeeds, arguments are ignored. Otherwise, they will be used for the JVM initialization. The JVM won't be attempt to start normally, if initial verification passes but restoration does not succeed.

test/jdk/jdk/crac/ignoreRestore/RestoreIfPossibleTest.java
```diff
@@ -57,13 +60,20 @@ public static void main(String[] args) throws Exception {
...
-            .stdoutShouldNotContain(WARMUP_MSG).stdoutShouldContain(MAIN_MSG);
+        final var out = builder.startRestoreWithArgs(null, List.of(Main.class.getName(), "false"))
+            .waitForSuccess().outputAnalyzer();
+        out.stdoutShouldNotContain(WARMUP_MSG);
+
+        // Check the count to ensure a new main is not launched on successful restore
+        final var mainMsgCount = Collections.frequency(out.stdoutAsLines(), MAIN_MSG);
+        if (mainMsgCount != 1) {
+            out.reportDiagnosticSummary();
+            assertEquals(1, mainMsgCount, "Main message should be printed exactly once");
```
AntonKozlov: This is effectively `assert(false)`, but written in a more complex way. Let's change this or move it out from the `if`?

TimPushkin: I use assert equals because it prints the expected and the actual values of the counter.

If we move this out of if the diagnostic info will be printed unconditionally (may slow down the test a bit if the output is large for some reason).

PR #263, opened 2025-08-13
8365484: [CRaC] Support IgnoreUnrecognizedVMOptions on restore

TimPushkin:
Makes `IgnoreUnrecognizedVMOptions` restore-settable.

`IgnoreUnrecognizedVMOptions` is not forwarded to the restored JVM since that does not make much sense (the options are filtered in the restoring VM).

---------
### Progress

### Issue
 * [JDK-8365484](https://bugs.openjdk.org/browse/JDK-8365484): [CRaC] Support IgnoreUnrecognizedVMOptions on restore (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/263/head:pull/263` \
`$ git checkout pull/263`

`$ git checkout pull/263` \
`$ git pull https://git.openjdk.org/crac.git pull/263/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 263`

`$ git pr show -t 263`

</details>

<a href="https://git.openjdk.org/crac/pull/263.diff">https://git.openjdk.org/crac/pull/263.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/263#issuecomment-3183741230)
</details>

### DIFF
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -2938,6 +2938,20 @@ void Arguments::fix_appclasspath() {
   }
 }
 
+// Some restore-settable flags should not be updated in the restored JVM.
+static bool should_record_for_restore(const JVMFlag& flag) {
+  precond(flag.is_restore_settable());
+  if (strncmp(flag.name(), "CRaCEngine", ARRAY_SIZE("CRaCEngine") - 1) == 0) {
+    assert(strcmp(flag.name(), "CRaCEngine") == 0 || strcmp(flag.name(), "CRaCEngineOptions") == 0,
+           "unexpected CRaCEngine* flag: %s", flag.name());
+    return false;
+  }
+  if (strcmp(flag.name(), "IgnoreUnrecognizedVMOptions") == 0) {
+    return false;
+  }
+  return true;
+}
+
 // arg is a JVM argument without the leading "-XX:", e.g. "+BoolOpt" or "StrOpt=str"
 bool Arguments::process_flag_for_restore(const char *arg) {
   // -XX:Flags and -XX:VMOptionsFile do not have corresponding JVMFlags
@@ -2951,15 +2965,14 @@ bool Arguments::process_flag_for_restore(const char *arg) {
   bool has_plus_minus;
   parse_argname(arg, &name, &name_len, &has_plus_minus);
 
-  const JVMFlag* flag;
+  const JVMFlag* flag = nullptr;
   bool is_name_real; // false if the given name is an alias for a real flag name
-  {
+  if (name_len <= BUFLEN) {
     char buf[BUFLEN + 1];
     const char* stripped_name;
     if (name[name_len] == '\0') {
       stripped_name = name;
     } else {
-      guarantee(name_len <= BUFLEN, "argument name too long: %s", name); // Should've been detected earlier
       strncpy(buf, name, name_len);
       buf[name_len] = '\0';
       stripped_name = buf;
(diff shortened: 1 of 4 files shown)


test/jdk/jdk/crac/VMOptionsTest.java
```diff
@@ -47,121 +49,154 @@
...
-        builder.vmOption("-XX:+UnlockExperimentalVMOptions");
+        setVmOptions(builder, OPTIONS_RESTORE);
+        builder.vmOption("-XX:CREngine=criuengine"); // Deprecated alias
         checkRestoreOutput(builder.doRestore());
 
-        // 5) Same as (1) but options come from a settings file => should succeed
+        // Only restore-settable options coming from a settings file => should succeed
         builder.clearVmOptions();
-        builder.vmOption("-XX:Flags=" + createSettingsFile(enginePath));
+        builder.vmOption("-XX:Flags=" + createSettingsFile(OPTIONS_RESTORE));
```
AntonKozlov, line 139: It seems we replace full path to the engine with the engine name, which is actually a relative path. I'm not sure, why the path was in the first place?

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -2938,6 +2938,20 @@ void Arguments::fix_appclasspath() {
...
 
+// Some restore-settable flags should not be updated in the restored JVM.
+static bool should_record_for_restore(const JVMFlag& flag) {
+  precond(flag.is_restore_settable());
+  if (strncmp(flag.name(), "CRaCEngine", ARRAY_SIZE("CRaCEngine") - 1) == 0) {
+    assert(strcmp(flag.name(), "CRaCEngine") == 0 || strcmp(flag.name(), "CRaCEngineOptions") == 0,
+           "unexpected CRaCEngine* flag: %s", flag.name());
+    return false;
+  }
+  if (strcmp(flag.name(), "IgnoreUnrecognizedVMOptions") == 0) {
```
AntonKozlov, line 2949: I see the test is consistent with this, but why do we don't record IgnoreUnrecognizedVMOptions?

test/jdk/jdk/crac/VMOptionsTest.java
TimPushkin: We need a different value of `CRaCEngine` VM option on checkpoint and on restore to check that on restore it does not change. It does not matter what values are used unless they make JVM use the same engine.

We can check that just the origin of the option has not changed but checking the value too covers more errors.

src/hotspot/share/runtime/arguments.cpp
TimPushkin: Not doing this should not lead to any errors in the current JVM code but I decided to do this because updating `IgnoreUnrecognizedVMOptions` may make the state of the restored VM somewhat inconsistent.

Suppose `IgnoreUnrecognizedVMOptions` was true on checkpoint and an unrecognized option was used in the command line. Then suppose `IgnoreUnrecognizedVMOptions` was set to false on restore. In such case after restore the command line (which I believe can be read via management or JVM TI) will contain unrecognized options while `IgnoreUnrecognizedVMOptions` will be false.

TimPushkin: I actually was not sure if this is needed or not. If you are against this, I'll remove it (i.e. make the option forwardable).

AntonKozlov: I see, thanks for the example. Right now, it also has an inconsistency. Suppose you enable it for the restore, and provide nonsense argument. The IgnoreUnrecognizedVMOptions took effect, you can sense unrecognized argument, but you check IgnoreUnrecognizedVMOptions and it's false.

I would prefer to remove special handling for the IgnoreUnrecognizedVMOptions, it will make things a bit more straightforward.

test/jdk/jdk/crac/VMOptionsTest.java
AntonKozlov: I see, thanks.

src/hotspot/share/runtime/arguments.cpp
TimPushkin: > Right now, it also has an inconsistency. Suppose you enable it for the restore, and provide nonsense argument. The IgnoreUnrecognizedVMOptions took effect, you can sense unrecognized argument, but you check IgnoreUnrecognizedVMOptions and it's false.

The unrecognized option will not be forwarded to the restored JVM so there is no inconsistency in this case.

But ok, I'll remove this special handling.

TimPushkin: Removed, although I am still not sure what is the best thing to do. Forwarding `IgnoreUnrecognizedVMOptions` seems more consistent, but the ignoring is fully handled by the restoring VM so forwarding it does not make much sense.

AntonKozlov: My bad. I mis-interpret the stuff. I see, there is no way "to sense" unrecognized options in the being restored VM. I agree that once all unrecognized are filtered out in restoring VM, there is no reason to forward IgnoreUnrecognizedVMOptions further to the being restored VM. The initial version of the change was good. Could you revert to that? Sorry for this back and forth.

PR #264, opened 2025-09-18
8367975: [CRaC] Pattern in CRaCCheckpointTo

rvansa:
Add support for pattern in `CRaCCheckpointTo` VM option.

---------
### Progress

### Issue
 * [JDK-8367975](https://bugs.openjdk.org/browse/JDK-8367975): [CRaC] Pattern in CRaCCheckpointTo (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/264/head:pull/264` \
`$ git checkout pull/264`

`$ git checkout pull/264` \
`$ git pull https://git.openjdk.org/crac.git pull/264/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 264`

`$ git pr show -t 264`

</details>

<a href="https://git.openjdk.org/crac/pull/264.diff">https://git.openjdk.org/crac/pull/264.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/264#issuecomment-3306466111)
</details>

### DIFF
--- a/src/hotspot/cpu/x86/vm_version_x86.hpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.hpp
@@ -601,20 +601,27 @@ protected:
       const char *format = hexonly ? UINT64_FORMAT_0 : UINT64_FORMAT_X;
       apply_to_all_features([&](uint64_t u, int idx) {
         int res = jio_snprintf(buf, buflen, format, u);
+        if (res < 0) {
+          buflen = 0;
+          return;
+        }
         buf += res;
         buflen -= res;
-        assert(res > 0 && buflen >= 1, "not enough temporary space allocated");
-        if (!hexonly && idx + 1 < features_bitmap_element_count()) {
+        if (!hexonly && idx + 1 < features_bitmap_element_count() && buflen > 0) {
           *buf++ = ',';
           --buflen;
         }
       });
+      if (buflen == 0) {
+        return -1;
+      }
       *buf = 0;
       return buf - buf_orig;
     }
 
     void print_numbers_and_names(char *buf, size_t buflen) const {
       int res = print_numbers(buf, buflen);
+      assert(res >= 0, "buffer too short");
       buf += res;
       buflen -= res;
       assert(buflen >= 3, "not enough temporary space allocated");
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -147,10 +147,10 @@ static int append_size(char *buf, size_t buflen, bool zero_pad, int width, size_
 
 #define check_retval(statement) do { \
     int ret = statement; \
-    if (ret < 0) { \
+    if ((size_t) ret > buflen) { \
       log_error(crac)("Error interpolating CRaCCheckpointTo=%s (too long)", CRaCCheckpointTo); \
       return false; \
-    } else if ((size_t) ret > buflen) { \
+    } else if (ret < 0) { \
       log_error(crac)("Error interpolating CRaCCheckpointTo=%s", CRaCCheckpointTo); \
       return false; \
     } \
@@ -216,9 +216,7 @@ bool crac::interpolate_checkpoint_location(char *buf, size_t buflen, bool *fixed
         check_no_width_padding();
         VM_Version::VM_Features data;
         if (VM_Version::cpu_features_binary(&data)) {
-          int ret = data.print_numbers(buf, buflen, true);
-          buf += ret;
-          buflen -= ret;
+          check_retval(data.print_numbers(buf, buflen, true));
         } // otherwise just empty string
       }
       break;
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -1963,7 +1963,7 @@ const int ObjectAlignmentInBytes = 8;
           "image is a directory, the directory will be created if it does " \
           "not exist (parent directories are not created) or overwritten "  \
           "otherwise. The path can contain placeholders (e.g. %p for PID);" \
-          "check documentation for pattern format." )                       \
+          " check documentation for pattern format.")                       \
                                                                             \
   product(ccstr, CRaCRestoreFrom, nullptr, RESTORE_SETTABLE,                \
           "Path to the image to restore from.")                             \
--- a/src/java.base/share/man/java.md
+++ b/src/java.base/share/man/java.md
@@ -1085,10 +1085,11 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
     It is possible to use a pattern with automatically inferred values, using
     these placeholders:
     -   `%%`: single % character
-    -   `%a`: architecture (x86_64 or aarch64)
+    -   `%a`: architecture; the same value as system property 'os.arch'
     -   `%f`: CPU features hex string. Empty string if the architecture does not
               use optional CPU features.
     -   `%u`: UUID (version 4 = random)
+    -   `%g`: checkpoint generation (starting with 1, 0 is reserved for ‘unknown’)
     -   `%t`: checkpoint date & time in ISO-8601 in UTC, basic format (without
               separators) with second precision, e.g. `20250909T141711Z`
     -   `%T`: checkpoint epoch time (second precision)
@@ -1098,7 +1099,6 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
     -   `%p`: PID of checkpointed process
     -   `%c`: number of CPU cores
     -   `%m`: max heap size (`-Xmx`) in a user-friendly format - using G or M suffix
-    -   `%g`: checkpoint generation (starting with 1, 0 is reserved for ‘unknown’)
 
     Numeric placeholders (`%T`, `%B`, `%R`, `%p`, `%c`, `%m` and `%g`) support
     an optional prefix with minimum width, padded with spaces or zeroes if the prefix
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -108,7 +108,7 @@ bool VM_Version::supports_clflush() {
   VM_Features flush;
   flush.set_feature(CPU_FLUSH);
   char buf[MAX_CPU_FEATURES];
-  flush.print_numbers(buf, sizeof(buf));
+  guarantee(flush.print_numbers(buf, sizeof(buf)) >= 0, "buffer too short");
   vm_exit_during_initialization(err_msg("-XX:CPUFeatures option requires FLUSH flag to be set: %s", buf));
   return false;
 }
@@ -1261,9 +1261,9 @@ bool VM_Version::glibc_not_using() {
   all_features.set_all_features();
   if (handled != all_features) {
     char buf_handled[MAX_CPU_FEATURES];
-    handled.print_numbers(buf_handled, sizeof(buf_handled));
+    guarantee(handled.print_numbers(buf_handled, sizeof(buf_handled)) >= 0, "buffer too short");
     char buf_all_features[MAX_CPU_FEATURES];
-    all_features.print_numbers(buf_all_features, sizeof(buf_all_features));
+    guarantee(all_features.print_numbers(buf_all_features, sizeof(buf_all_features)) >= 0, "buffer too short");
     vm_exit_during_initialization(err_msg("internal error: Unsupported disabling of some CPU_* %s != full %s", buf_handled, buf_all_features));
   }
 #endif // ASSERT
@@ -1282,7 +1282,7 @@ void VM_Version::print_using_features_cr() {
     tty->print_cr("CPU features are being kept intact as requested by -XX:CPUFeatures=ignore");
   } else {
     char buf[MAX_CPU_FEATURES];
-    _features.print_numbers(buf, sizeof(buf));
+    guarantee(_features.print_numbers(buf, sizeof(buf)) >= 0, "buffer too short");
     tty->print_cr("CPU features being used are: -XX:CPUFeatures=%s", buf);
   }
 }
@@ -1322,7 +1322,7 @@ void VM_Version::get_processor_features_hardware() {
 
   if (ShowCPUFeatures) {
     char buf[MAX_CPU_FEATURES];
-    _features.print_numbers(buf, sizeof(buf));
+    guarantee(_features.print_numbers(buf, sizeof(buf)) >= 0, "buffer too short");
     tty->print_cr("This machine's CPU features are: -XX:CPUFeatures=%s", buf);
   }
 }
@@ -1334,7 +1334,7 @@ void VM_Version::get_processor_features_hotspot() {
       VM_Features sse2;
       sse2.set_feature(CPU_SSE2);
       char buf[MAX_CPU_FEATURES];
-      sse2.print_numbers(buf, sizeof(buf));
+      guarantee(sse2.print_numbers(buf, sizeof(buf)) >= 0, "buffer too short");
       vm_exit_during_initialization(err_msg("-XX:CPUFeatures option requires SSE2 flag to be set: %s", buf));
     }
     vm_exit_during_initialization("Unknown x64 processor: SSE2 not supported");
@@ -2581,9 +2581,9 @@ bool VM_Version::cpu_features_binary_check(const VM_Version::VM_Features *data_p
 
   if (!features_missing.empty()) {
     char buf_use[MAX_CPU_FEATURES];
-    (data & _features).print_numbers(buf_use, sizeof(buf_use));
+    guarantee((data & _features).print_numbers(buf_use, sizeof(buf_use)) >= 0, "buffer too short");
     char buf_have[MAX_CPU_FEATURES];
-    data.print_numbers(buf_have, sizeof(buf_have));
+    guarantee(data.print_numbers(buf_have, sizeof(buf_have)) >= 0, "buffer too short");
     tty->print("You have to specify -XX:CPUFeatures=%s together with -XX:CRaCCheckpointTo when making a checkpoint file"
                "; specified -XX:CRaCRestoreFrom file contains CPU features %s",
                buf_use, buf_have);
@@ -2671,9 +2671,9 @@ void VM_Version::initialize() {
 
   if (!features_missing.empty()) {
     char buf_CPUFeatures_parsed[MAX_CPU_FEATURES];
-    CPUFeatures_parsed.print_numbers(buf_CPUFeatures_parsed, sizeof(buf_CPUFeatures_parsed));
+    guarantee(CPUFeatures_parsed.print_numbers(buf_CPUFeatures_parsed, sizeof(buf_CPUFeatures_parsed)) >= 0, "buffer too short");
     char buf_features[MAX_CPU_FEATURES];
-    _features.print_numbers(buf_features, sizeof(buf_features));
+    guarantee(_features.print_numbers(buf_features, sizeof(buf_features)) >= 0, "buffer too short");
     tty->print("Specified -XX:CPUFeatures=%s; this machine's CPU features are %s", buf_CPUFeatures_parsed, buf_features);
     features_missing.print_missing_features();
     vm_exit_during_initialization();


src/hotspot/cpu/x86/vm_version_x86.hpp
```diff
@@ -612,6 +612,18 @@ class VM_Version : public Abstract_VM_Version {
       return buf - buf_orig;
     }
 
+    int print_numbers_hexonly(char *buf_orig, size_t buflen) const {
```
TimPushkin: Can be merged into `print_numbers` with a parameter like `bool hex_only` (false by default, true where this was used), but I don't insist

src/hotspot/share/runtime/crac_engine.hpp
```diff
@@ -39,7 +39,7 @@
 // CRaC engine library wrapper.
 class CracEngine : public CHeapObj<mtInternal> {
 public:
-  explicit CracEngine(const char *image_location = nullptr);
+  explicit CracEngine();
```
TimPushkin: `explicit` is not needed now. It was there to adhere to https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-explicit.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
 
+static int append_time(char *buf, size_t buflen, bool iso8601, bool zero_pad, int width, long timeMillis) {
+  if (iso8601) {
+    if (width >= 0 || zero_pad) {
+      log_warning(crac)("CRaCCheckpointTo=%s cannot set zero_pad or width for ISO-8601 time", CRaCCheckpointTo);
+    }
+    // os::iso8601_time formats with dashes and colons, we want the basic version
+    time_t time = timeMillis / 1000;
+    struct tm tms;
+    if (gmtime_r(&time, &tms) == NULL) {
```
TimPushkin: There is `os::gmtime_pd` which is `gmtime_r` on POSIX and `gmtime` on Windows. Not sure why it exists but I would suggest to use it just in case.

```suggestion
    if (gmtime_r(&time, &tms) == nullptr) {
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+  }
+}
+
+static int append_size(char *buf, size_t buflen, bool zero_pad, int width, size_t size) {
+  if (zero_pad) {
+    return snprintf(buf, buflen, "%0*zu", width, size);
+  } else if (width >= 0) {
+    return snprintf(buf, buflen, "%*zu", width, size);
+  } else {
+    const char *suffixes[] = { "k", "M", "G" };
```
TimPushkin: I believe without  `static` this is put on stack
```suggestion
    static constexpr const char *suffixes[] = { "k", "M", "G" };
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
 }
 #endif //__APPLE__ || _WINDOWS
 
+static int append_time(char *buf, size_t buflen, bool iso8601, bool zero_pad, int width, long timeMillis) {
+  if (iso8601) {
+    if (width >= 0 || zero_pad) {
+      log_warning(crac)("CRaCCheckpointTo=%s cannot set zero_pad or width for ISO-8601 time", CRaCCheckpointTo);
```
TimPushkin: `zero_pad` is a name of a variable not known to the user.
```suggestion
      log_warning(crac)("CRaCCheckpointTo=%s cannot zero-pad or set width for ISO-8601 time", CRaCCheckpointTo);
```

I would also write "Cannot \<do something> in CRaCCheckpointTo=%s" here and in other similar places below but that is a personal preference.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+#define check_no_width() do { \
+    if (width >= 0) { \
+      log_warning(crac)("CRaCCheckpointTo=%s cannot set width for %%%c", CRaCCheckpointTo, c); \
+    } \
+  } while (false)
+
+#define check_retval(statement) do { \
+    int ret = statement; \
+    if (ret < 0 || (size_t) ret > buflen) { \
+      log_error(crac)("Error interpolating CRaCCheckpointTo=%s (@%zu, buffer size %zu)", CRaCCheckpointTo, si, buflen); \
```
TimPushkin: To me it is not completely obvious what @ means here, if it is the position in the interpreted string then I would prefer "position"/"location"

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+      return false; \
+    } \
+    buf += ret; \
+    buflen -= ret; \
+  } while (false)
+
+bool crac::resolve_image_location(char *buf, size_t buflen, bool *fixed) {
```
TimPushkin: I expect "resolve" to access the filesystem somehow. Also it is only meant for the checkpoint location. So how about  `interpolate_checkpoint_image_location`?

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+  *fixed = true;
+  for (size_t si = 0; ; si++) {
+    char c = CRaCCheckpointTo[si];
+    if (c == '%') {
```
TimPushkin: I would suggest handling the simple case first with continue to reduce the nesting:

```c
    if (c != '%') {
      *(buf++) = c;
      --buflen;
      if (c == '\0') {
        break;
      } else {
        continue;
      }
    }
    // handle % with reduce nesting
```

src/hotspot/share/runtime/crac.cpp
TimPushkin: `buflen` is not checked when we write to the buffer directly, not via stdlib functions

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+      while (CRaCCheckpointTo[si] >= '1' && CRaCCheckpointTo[si] <= '9') {
+        ++si;
+      }
+      int width = -1;
+      if (zero_pad && width_start == si) {
+        log_error(crac)("CRaCCheckpointTo=%s contains a pattern with zero padding but no length", CRaCCheckpointTo);
+        return false;
+      } else if (si > width_start) {
+        width = atoi(&CRaCCheckpointTo[width_start]);
+      }
```
TimPushkin: Since width is not set in the first case:
```suggestion
      if (zero_pad && width_start == si) {
        log_error(crac)("CRaCCheckpointTo=%s contains a pattern with zero padding but no length", CRaCCheckpointTo);
        return false;
      }
      const int width = si > width_start ? atoi(&CRaCCheckpointTo[width_start]) : -1;
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+    if (c == '%') {
+      si++;
+      c = CRaCCheckpointTo[si];
+      bool zero_pad = false;
+      if (c == '0') {
+        zero_pad = true;
+        si++;
+      }
+      size_t width_start = si;
+      while (CRaCCheckpointTo[si] >= '1' && CRaCCheckpointTo[si] <= '9') {
```
TimPushkin: Since 0 is not allowed, I believe currently it is not possible to specify width 10, for example?

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+      switch (c) {
+      case '%':
+        check_no_width();
+        *(buf++) = '%';
+        --buflen;
+        break;
+      case 'a': // CPU architecture; matches system property "os.arch"
+          check_no_width();
+#ifndef ARCHPROPNAME // defined by build scripts
+# define ARCHPROPNAME "unknown"
```
TimPushkin: Why not assert it is defined?
```suggestion
#error ARCHPROPNAME macro must be defined
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+#endif
+        check_retval(snprintf(buf, buflen, "%s", ARCHPROPNAME));
+        break;
+      case 'f': { // CPU features
+          struct VM_Version::VM_Features data;
+          if (VM_Version::cpu_features_binary(&data)) {
+            int ret = data.print_numbers_hexonly(buf, buflen);
+            buf += ret;
+            buflen -= ret;
+          } // otherwise just empty string
```
TimPushkin: I think we should print a warning otherwise

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+        --buflen;
+        break;
+      case 'a': // CPU architecture; matches system property "os.arch"
+          check_no_width();
+#ifndef ARCHPROPNAME // defined by build scripts
+# define ARCHPROPNAME "unknown"
+#endif
+        check_retval(snprintf(buf, buflen, "%s", ARCHPROPNAME));
+        break;
+      case 'f': { // CPU features
```
TimPushkin: Should check no width/padding

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+    const char *suffix = "";
+    for (size_t i = 0; i < ARRAY_SIZE(suffixes) && size != 0 && (size & 1023) == 0; ++i) {
+      suffix = suffixes[i];
+      size = size >> 10;
+    }
+    return snprintf(buf, buflen, "%zu%s", size, suffix);
+  }
+}
+
+#define check_no_width() do { \
```
```suggestion
// Also implies no zero-padding
#define check_no_width() do { \
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+        }
+        break;
+      case 'u': { // Random UUID (v4)
+          check_no_width();
+          *fixed = false; // FIXME?
+          u4 time_mid_high = static_cast<u4>(os::random());
+          u4 seq_and_node_low = static_cast<u4>(os::random());
+          check_retval(snprintf(buf, buflen, "%08x-%04x-4%03x-a%03x-%04x%08x",
+            static_cast<u4>(os::random()), time_mid_high >> 16, time_mid_high & 0xFFF,
+            seq_and_node_low & 0xFFF, seq_and_node_low >> 16, static_cast<u4>(os::random())));
```
TimPushkin: Not that it matters much, but I believe the 4th part does not always have to start with "a" according to the spec

```suggestion
          check_retval(snprintf(buf, buflen, "%08x-%04x-4%03x-%04x-%04x%08x",
            static_cast<u4>(os::random()), time_mid_high >> 16, time_mid_high & 0xFFF,
            (seq_and_node_low & 0x3FFF | 0x8000), seq_and_node_low >> 16, static_cast<u4>(os::random())));
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+          if (VM_Version::cpu_features_binary(&data)) {
+            int ret = data.print_numbers_hexonly(buf, buflen);
+            buf += ret;
+            buflen -= ret;
+          } // otherwise just empty string
+        }
+        break;
+      case 'u': { // Random UUID (v4)
+          check_no_width();
+          *fixed = false; // FIXME?
```
TimPushkin: Looks correct to me, what would you want to fix?

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+      case 'b': // boot time
+      case 'B':
+        check_retval(append_time(buf, buflen, c == 'b', zero_pad, width,
+            os::javaTimeMillis() - 1000 * (os::elapsed_counter() / os::elapsed_frequency())));
+        break;
+      case 'r': // last restore time
+      case 'R':
+        check_retval(append_time(buf, buflen, c == 'r', zero_pad, width,
+            os::javaTimeMillis() - 1000 * (os::elapsed_counter_since_restore() / os::elapsed_frequency())));
+        break;
```
TimPushkin: Shouldn't we use the native implementations of `RuntimeMXBean.getStartTime()` and `CRaCMXBean.getRestoreTime()` respectively? I think it is reasonable to expect these to be consistent. But if there are reasons not to do that it is ok with me to leave this as is, we just won't document them to be consistent in such case.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+      }
+    }
+  }
+  return true;
+}
+
+#undef check_no_width
+#undef check_retval
+
+static bool ensure_image_location(const char *path, bool rm) {
```
TimPushkin: Since this is checkpoint-specific:
```suggestion
static bool ensure_checkpoint_image_location(const char *path, bool rm) {
```

Or maybe make it also handle the restore case have this in one place.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,195 @@ static char * strchrnul(char * str, char c) {
...
+      !ensure_image_location(image_location, false)) {
+    return JVM_CHECKPOINT_ERROR;
+  }
   // Note that CRaCEngine and CRaCEngineOptions are not updated (as documented)
   // so we don't need to re-init the whole engine handle.
-  if (restore_start_time() != -1 && // A way to detect we've restored at least once
-      !_engine->configure_image_location(CRaCCheckpointTo)) {
+  if (!_engine->configure_image_location(image_location)) {
     return JVM_CHECKPOINT_ERROR;
   }
```
TimPushkin: Is there a reason we do this and CPU features recording below so late instead of at the beginning `crac::checkpoint`, for example? The only reason I see is that the timestamps in `crac::record_time_before_checkpoint` and `%t` would be further away in that case (BTW, if we leave this as is we could use the recorded time for the pattern for consistency).

If we did not clear the recorded user data on checkpoint, CPU features could be recorded just once on VM initialization instead of on every checkpoint (though not in `prepare_checkpoint` — that is too early, I believe).

And just to have it more compact since the calls are very closely related
```suggestion
  // Note that CRaCEngine and CRaCEngineOptions are not updated (as documented)
  // so we don't need to re-init the whole engine handle.
  if (!resolve_image_location(image_location, sizeof(image_location), &ignored) ||
      !ensure_image_location(image_location, false) ||
      !_engine->configure_image_location(image_location)) {
    return JVM_CHECKPOINT_ERROR;
}
```

test/jdk/jdk/crac/PathPatternTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+    @Override
+    public void test() throws Exception {
+        //noinspection ResultOfMethodCallIgnored
+        new File("foo").mkdirs();
+
+        // Fixed fields only, no-timestamps
+        long pid = runCheckpoints("foo/cr_%%_%a_%p_%05c_%3m_%m_%g", false);
+        File f = new File(String.format("foo/cr_%%_%s_%d_%05d_%d_1G_1",
+                Platform.getOsArch(), pid, Runtime.getRuntime().availableProcessors(), 1024 * 1024 * 1024));
```
TimPushkin: FYI there's `jdk.test.lib.Unit.G` which has `size == 1024 * 1024 * 1024`, and there's other common units

test/jdk/jdk/crac/PathPatternTest.java
```diff
@@ -0,0 +1,170 @@
...
+        assertTrue(f.exists(), "Expect" + f);
```
TimPushkin:
```suggestion
        assertTrue(f.exists(), "Expect " + f);
```

test/jdk/jdk/crac/PathPatternTest.java
```diff
@@ -0,0 +1,170 @@
...
+            assertEquals("1", matcher1.group(7));
+            assertEquals("2", matcher2.group(7));
+        }
+    }
+
+    private static void checkDateEquals(String str, long ts) {
+        assertEquals(str, DATE_FORMAT.format(new Date(TimeUnit.SECONDS.toMillis(ts))));
+    }
+
+    private static void deleteDir(String dir) throws IOException {
```
TimPushkin: `clearDir` would fit better. Also it would be simpler to use `jdk.test.lib.util.FileUtils.deleteFileTreeWithRetry(Path)` and re-create the directory itself.

test/jdk/jdk/crac/PathPatternTest.java
```diff
@@ -0,0 +1,170 @@
...
+        // Do two checkpoint-restores to ensure code behaves correctly after repeated execution
+        Core.checkpointRestore();
+        if (Boolean.getBoolean("test.pause")) {
+            Thread.sleep(1000);
+        }
+        Core.checkpointRestore();
+    }
+
+
+}
```
TimPushkin, line 166:
```suggestion
    }
}
```

src/hotspot/share/runtime/crac.cpp
rvansa: I had the `define` here mostly to avoid IDE errors, but OK...

src/hotspot/share/runtime/crac.cpp
rvansa: I don't think so; `cpu_features_binary` will return false on `aarch64`, and I don't think the user has to use different patterns to checkpoint on different architectures.

src/hotspot/share/runtime/crac.cpp
rvansa: The UUID could be generated once on the first pattern interpolation. Probably needs to be reset on restore, then, to keep the semantic (UUID should change on each checkpoint - that shall be documented, too).

src/hotspot/share/runtime/crac.cpp
rvansa: We're using RFC 4122 variant, with Msb0 = 1 and Msb1 = 0: https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.1

src/hotspot/share/runtime/crac.cpp
TimPushkin: From the other hand, it can be confusing to get absolutely nothing in the place of this pattern. But if we explicitly document that this is possible, I am OK with leaving it as is.

src/hotspot/share/runtime/crac.cpp
TimPushkin: Yes, and IIUC it fixes only two bits, allowing the other two to change: 1000 = 8, 1001 = 9, 1010 = a, 1011 = b. Or am I reading it wrong?

src/hotspot/share/runtime/crac.cpp
rvansa: I don't follow the second part - restore should not attempt to create the directory.

src/hotspot/share/runtime/crac.cpp
rvansa: In this PR I have merely reused the place; I don't see much incentive to move it earlier on. I think this is up to preference between
1) doing some (reversible) work on checkpoint and then (possibly) failing due to no access to the directory, vs. 
2) creating the directory, (possibly) failing during checkpoint and leaving the filesystem tainted (or implementing cleanup)
I don't have a strong preference, but I agree that we can collapse the two `if`s.

src/hotspot/share/runtime/crac.cpp
TimPushkin: Yes, we would need to add a parameters like `bool is_checkpoint` which will define whether we need to create a directory or just check for existence

src/hotspot/share/runtime/crac.cpp
rvansa: Sorry, you're right. My mind flipped binary `10` to hexadecimal `a` as if it was decimal `10`.

src/hotspot/share/runtime/crac.cpp
TimPushkin: Even with the current code we can still fail after creating the directory, without writing anything and AFAIK without cleaning up, so I do not think moving it earlier will make this much worse.

But yes, for now let's leave it here.

rvansa:
Looks like when I've created the PR from a branch in this repository the GHA with tests is not executed. Pushed the code to my repo, too, tests running here: https://github.com/rvansa/crac/actions/runs/18130706409

src/hotspot/share/runtime/crac.cpp
```diff
@@ -136,125 +136,140 @@ static int append_size(char *buf, size_t buflen, bool zero_pad, int width, size_
...
 #define check_retval(statement) do { \
     int ret = statement; \
-    if (ret < 0 || (size_t) ret > buflen) { \
-      log_error(crac)("Error interpolating CRaCCheckpointTo=%s (@%zu, buffer size %zu)", CRaCCheckpointTo, si, buflen); \
+    if (ret < 0) { \
+      log_error(crac)("Error interpolating CRaCCheckpointTo=%s (too long)", CRaCCheckpointTo); \
+      return false; \
+    } else if ((size_t) ret > buflen) { \
+      log_error(crac)("Error interpolating CRaCCheckpointTo=%s", CRaCCheckpointTo); \
       return false; \
```
TimPushkin: Shouldn't the error messages be swapped? For snprintf, if `ret > buflen` then the buffer was too small, if `ret < 0` then it is some encoding error. Also I think it should be `ret >= buflen` since `ret` does not count the nul terminator.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,208 @@ static char * strchrnul(char * str, char c) {
...
+#endif
+      check_retval(snprintf(buf, buflen, "%s", ARCHPROPNAME));
+      break;
+    case 'f': { // CPU features
+        check_no_width_padding();
+        struct VM_Version::VM_Features data;
+        if (VM_Version::cpu_features_binary(&data)) {
+          int ret = data.print_numbers(buf, buflen, true);
+          buf += ret;
+          buflen -= ret;
```
TimPushkin: We only check that there is enough space for a single character in the buffer while `print_numbers` asserts that there is enough space for all the features to fit. I.e. I believe it is possible to go past the end of the buffer here.

Although not a part of this change, but the assert in `print_numbers` should be before subtracting from `buflen`, not after.

src/java.base/share/man/java.md
```diff
@@ -1082,6 +1082,28 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
     directory will be created if it does not exist, but no parent directories are
     created.
 
+    It is possible to use a pattern with automatically inferred values, using
+    these placeholders:
+    -   `%%`: single % character
+    -   `%a`: architecture (x86_64 or aarch64)
```
TimPushkin: Are you sure it is always one of these? [This code](https://github.com/openjdk/crac/blob/5abb5dac8e9785ad0c38e238e526775f98617aec/make/autoconf/platform.m4#L475) suggests that it can be i386 or amd64 for example. Also the shared part of CRaC is probably compilable for other arches too. Maybe better to say it is the same as what `os.arch` returns?

src/java.base/share/man/java.md
```diff
@@ -1082,6 +1082,28 @@ These `java` options control the runtime behavior of the Java HotSpot VM.
...
+    -   `%f`: CPU features hex string. Empty string if the architecture does not
+              use optional CPU features.
+    -   `%u`: UUID (version 4 = random)
+    -   `%t`: checkpoint date & time in ISO-8601 in UTC, basic format (without
+              separators) with second precision, e.g. `20250909T141711Z`
+    -   `%T`: checkpoint epoch time (second precision)
+    -   `%b` and `%B`: process boot time (generation 1), same format as `%t` or `%T`
```
TimPushkin, line 1096: I suggest putting the description of `%g` before this (between `%u` and `%t`, for example) so that the definition of a "generation" is more clear when the reader reaches this point

src/hotspot/share/runtime/crac.cpp
```diff
@@ -99,17 +101,208 @@ static char * strchrnul(char * str, char c) {
...
 
-  // CRaCCheckpointTo can be changed on restore so we need to update the conf
-  // to account for that.
+  // CRaCCheckpointTo can be changed on restore, and if this contains a pattern
+  // it might not have been configured => we need to update the conf.
   // Note that CRaCEngine and CRaCEngineOptions are not updated (as documented)
   // so we don't need to re-init the whole engine handle.
-  if (restore_start_time() != -1 && // A way to detect we've restored at least once
-      !_engine->configure_image_location(CRaCCheckpointTo)) {
+    char image_location[PATH_MAX];
```
```suggestion
  char image_location[PATH_MAX];
```

src/hotspot/share/runtime/globals.hpp
```diff
@@ -1962,7 +1962,8 @@ const int ObjectAlignmentInBytes = 8;
           "Path where the checkpoint image should be placed. Currently an " \
           "image is a directory, the directory will be created if it does " \
           "not exist (parent directories are not created) or overwritten "  \
-          "otherwise.")                                                     \
+          "otherwise. The path can contain placeholders (e.g. %p for PID);" \
```
TimPushkin, line 1965: Whitespace is missing after ";"

rvansa:
@TimPushkin I hope I have addressed these, and a green build in CI is here: https://github.com/rvansa/crac/actions/runs/18275647535

src/hotspot/share/runtime/crac.cpp
```diff
@@ -289,6 +289,10 @@ static bool ensure_checkpoint_dir(const char *path, bool rm) {
   return true;
 }
 
+#ifndef PATH_MAX
+# define PATH_MAX 1024
+#endif
```
TimPushkin, line 292: Not critical, but we could also try `MAX_PATH` defined in Windows before falling back to a hard-coded value

TimPushkin (approved):
LGTM

PR #265, opened 2025-09-24
8368570: [CRaC] Improve JdwpTransportTest

wkia:
JdwpTransportTest sometimes hangs, so it'd be good to dump Java threads, and create a corefile if possible.

Also improved reading from output stream.

---------
### Progress

### Issue
 * [JDK-8368570](https://bugs.openjdk.org/browse/JDK-8368570): [CRaC] Improve JdwpTransportTest (**Task** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/265/head:pull/265` \
`$ git checkout pull/265`

`$ git checkout pull/265` \
`$ git pull https://git.openjdk.org/crac.git pull/265/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 265`

`$ git pr show -t 265`

</details>

<a href="https://git.openjdk.org/crac/pull/265.diff">https://git.openjdk.org/crac/pull/265.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/265#issuecomment-3329363312)
</details>

### DIFF
--- a/test/lib/jdk/test/lib/crac/CracProcess.java
+++ b/test/lib/jdk/test/lib/crac/CracProcess.java
@@ -234,7 +234,7 @@ public class CracProcess {
         }
     }
 
-    public void dumpProcess() throws IOException, InterruptedException {
+    public void dumpProcess() {
         // For gcore, it's required 'sudo sysctl -w kernel.yama.ptrace_scope=0'
         // For kill, it's required 'ulimit -c unlimited && echo core.%p | sudo tee /proc/sys/kernel/core_pattern'
         final long pid = this.pid();


test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, Azul Systems, Inc. All rights reserved.
+ * Copyright (c) 2023-2025, Azul Systems, Inc. All rights reserved.
```
TimPushkin: I believe the convention is to write it like this, not sure if there are any legally significant differences
```suggestion
 * Copyright (c) 2023, 2025, Azul Systems, Inc. All rights reserved.
```

test/lib/jdk/test/lib/crac/AsyncStreamReader.java
```diff
@@ -0,0 +1,42 @@
+package jdk.test.lib.crac;
```
TimPushkin, line 24: Missing copyright header

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
```diff
@@ -104,9 +108,9 @@ public void test() throws Exception {
         builder.engine(CracEngine.SIMULATE);
 
         CracProcess process = builder.captureOutput(true).startCheckpoint();
-        var errReader = new BufferedReader(new InputStreamReader(process.errOutput()));
+        var errReader = new AsyncStreamReader(process.errOutput());
+        var reader = new AsyncStreamReader(process.output());
```
TimPushkin, line 112: Not that it matters much in this test, but I believe `AsyncStreamReader` should be made `AutoCloseable` (on close, close the buffer, maybe also stop the reading thread) and these should be treated as resources in the try below

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
```diff
@@ -140,21 +144,31 @@ public void test() throws Exception {
...
+            }
+            process.destroyForcibly();
         }
     }
 
     @Override
     public void exec() throws Exception {
-        System.out.println(STARTED);
+        System.out.println(STARTED + ", pid=" + ProcessHandle.current().pid());
+        System.out.flush();
```
TimPushkin: AFAIK `System.out` is auto-flushed when `\n` is written, so `println` should be sufficient

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
```diff
@@ -140,21 +144,31 @@ public void test() throws Exception {
             process.input().flush();
 
             process.waitForSuccess();
+        } catch (TimeoutException e) {
+            System.err.println("reader.isRunning()=" + reader.isRunning());
+            CracProcess.printThreadDump(process.pid());
+            CracProcess.dumpProcess(process.pid());
```
TimPushkin: If we end up always taking the PID from `CracProcess` itself why making these two static?

test/lib/jdk/test/lib/crac/CracProcess.java
```diff
@@ -192,4 +194,71 @@ public void sendNewline() throws IOException {
...
+                System.out.println("gcore is NOT available.");
+                return false;
+            }
+        } catch (IOException | InterruptedException e) {
+            System.out.println("Could not run 'which gcore' or was interrupted");
+            return false;
+        }
+    }
+
+    public static void dumpProcess(long pid) throws IOException, InterruptedException {
```
TimPushkin: `JdwpTransportTest` runs on all platforms and I believe this is not going to work on non-Linux

test/lib/jdk/test/lib/crac/CracProcess.java
```diff
@@ -192,4 +194,71 @@ public void sendNewline() throws IOException {
...
+            int exitCode = process.waitFor();
+            try {
+                while (true) {
+                    System.out.println("dumpProcess: " + reader.readLine(100));
+                }
+            } catch (Exception e) {
+                // do nothing
+            }
+            if (exitCode == 0) {
+                System.out.println("Core dump seems created successfully for pid=" + pid);
```
TimPushkin:
```suggestion
                System.out.println("Core dump seems to be created successfully for pid=" + pid);
```

test/lib/jdk/test/lib/crac/AsyncStreamReader.java
wkia: There is no copyright header for all files in in `test/jdk/jdk/crac/` folder, so I thought it's intended to be so.

TimPushkin: I think that is an oversight as other files in `test/lib` have them. We have an internal issue to bring all the missing license headers to CRaC code.

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
wkia: For this test, yes. But these two methods may be useful for tests when Java changes its PID on restore. 

I will change this, if you doesn't think so.

test/lib/jdk/test/lib/crac/CracProcess.java
wkia: I think it should work on linuxes, macos, and cygwin. cygwin doesn't produce a core file of course. For other cases, the command will fail with a message. I don't think this is important as it will be triggered when process hangs only.

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
TimPushkin: I'd make it non-static for now and change to static when/if we need that but I don't insist

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
wkia: Autocloseable is ok for me, but these two cannot be used in catch and finally blocks in case they are resources.

TimPushkin: Oh, yes, then there should be either another wrapping try-with-resources or explicit `close` calls in the existing `finally`

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
wkia: done

test/lib/jdk/test/lib/crac/AsyncStreamReader.java
wkia: done

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
wkia: done

test/jdk/jdk/crac/jdwp/JdwpTransportTest.java
wkia: OK, I've done this.
However, in this particular case, it seems like `process.destroyForcibly()` should guarantee the streams are closed from the prosess' side.

test/lib/jdk/test/lib/crac/CracProcess.java
TimPushkin: I believe it does not actually throw these
```suggestion
    public static void dumpProcess(long pid) {
```

wkia: done

TimPushkin:
`jdk/crac/RemoteJmxTest` failed — I doubt it is related but please check and file an issue if not related

PR #266, opened 2025-09-30
8368929: [CRaC] Move CPUFeatures check to C/R engine

rvansa:
Right now the logic checking if CPU features used before checkpoint match current CPU features is in VM code. VM stores and retrieves CPU features through C/R API's user_data extension. This is convenient when we have a single image that can be either accepted or rejected, but does not offer the flexibility for C/R engine to select the best image for current execution environment.

The goal of this issue is to move to a declarative API that will express the requirements using more abstract means, labels (for CPU architecture) and bitmaps (for CPU features).


rvansa:
@jankratochvil Please give this a review, since you've authored most of the modified code.

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,224 @@
...
+    char *nl = strchr(eq + 1, '\n');
+    if (eq == nullptr || nl == nullptr) {
+      fprintf(stderr, CREXEC "Invalid format of tags file: %s\n", line);
+      return RESTORE_ERROR_INVALID;
+    }
+    *eq = 0;
+    *nl = 0;
+    assert(eq < nl);
+    if (!strncmp(line, LABEL_PREFIX, strlen(LABEL_PREFIX))) {
+      Tag& t = tags.add({
```
jankratochvil:
```
/home/runner/work/crac/crac/src/java.base/share/native/libcrexec/image_constraints.cpp:155:12: error: unused variable ‘t’ [-Werror=unused-variable]
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -520,17 +520,22 @@ void crac::restore(crac_restore_data& restore_data) {
     return;
   }
 
-  if (!VM_Version::ignore_cpu_features()) {
-    switch (engine.prepare_user_data_api()) {
+  // Previously IgnoreCPUFeatures didn't disable the check completely; the difference
+  // was printed out but continued even despite features not being satisfied.
+  // Since the check itself is delegated to the C/R Engine we will simply
+  // skip the check here (or prevent storing the features in the image).
+#ifdef __x86_64__
```
jankratochvil: One should not use `#ifdef` as it may lead to compilation errors on other arches.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -579,7 +584,17 @@ void crac::restore(crac_restore_data& restore_data) {
   }
 
   const int ret = engine.restore();
-  if (ret != 0) {
+  if (ret == RESTORE_ERROR_INVALID) {
+    log_error(crac)("CRaC engine failed to restore from %s: the image is either corrupted or does not match current CPU", CRaCRestoreFrom);
+#ifdef __x86_64__
```
jankratochvil: Likewise one should not use `#ifdef` as it may lead to compilation errors on other arches.

src/hotspot/share/include/crlib/crlib.h
```diff
@@ -76,6 +76,14 @@ struct crlib_api {
 };
 typedef const struct crlib_api crlib_api_t;
 
+// Error codes returned from crlib_api.restore():
+#define RESTORE_ERROR_UNKNOWN   -1 // generic error
+#define RESTORE_ERROR_NOT_FOUND -2 // image location does not contain an image
+#define RESTORE_ERROR_NO_ACCESS -3 // image cannot be accessed/retrieved (permissions or I/O issue)
+#define RESTORE_ERROR_INVALID   -4 // image is not suitable (e.g. corruption or wrong architecture)
+#define RESTORE_ERROR_MEMORY    -5 // memory allocation failure during restore
+#define RESTORE_ERROR_PROCINFO  -6 // the process cannot fetch information about itself
```
jankratochvil: Cannot it be an enum?

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+#include "image_constraints.hpp"
+#include "hashtable.hpp"
+
+#define CREXEC "crexec: "
+#define LABEL_PREFIX "label:"
+#define BITMAP_PREFIX "bitmap:"
+
+static FILE *open_tags(const char *image_location, const char *mode) {
+   char fname[PATH_MAX];
```
jankratochvil: whitespace

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,224 @@
...
+bool ImageConstraints::persist(const char *image_location) const {
+  FILE *f = open_tags(image_location, "w");
+  if (f == nullptr) {
+    return false;
+  }
+  _tags.foreach([&](const Tag &tag){
+    if (tag.type == LABEL) {
+        fprintf(f, LABEL_PREFIX "%s=%s\n", tag.name, static_cast<const char *>(tag.data));
+    } else {
+        fprintf(f, BITMAP_PREFIX "%s=", tag.name);
```
jankratochvil: indentation 2 vs. 4

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,224 @@
...
+  });
+  if (fclose(f)) {
+    fprintf(stderr, CREXEC "cannot close %s/tags: %s\n", image_location, strerror(errno));
+    return false;
+  }
+  return true;
+}
+
+static inline unsigned char from_hex(char c, bool *err) {
+    if (c >= '0' && c <= '9') {
```
jankratochvil: again changed indentation, more occurences in this file

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,224 @@
...
+  if (keys == nullptr) {
+    fprintf(stderr, CREXEC "Insufficient memory\n");
+    return RESTORE_ERROR_MEMORY;
+  }
+  int counter = 0;
+  tags.foreach([&](const Tag &t) {
+    keys[counter++] = t.name;
+  });
+  Hashtable<Tag> ht(keys, tags.size());
+  delete keys;
```
jankratochvil:
```
src/java.base/share/native/libcrexec/image_constraints.cpp:198:3: warning: 'delete' applied to a pointer that was allocated with 'new[]'; did you mean 'delete[]'? [-Wmismatched-new-delete]
  198 |   delete keys;
      |   ^
      |         []
src/java.base/share/native/libcrexec/image_constraints.cpp:188:23: note: allocated with 'new[]' here
  188 |   const char **keys = new(std::nothrow) const char*[tags.size()];
      |                       ^
```

src/hotspot/share/include/crlib/crlib.h
rvansa: The API already defines return value as `int`, we don't want to change that part.

src/hotspot/share/runtime/crac.cpp
rvansa: How is this different from creating an arch-specific method with different implementations? To me that look just like a boilerplate (and the presence is only checked for currently compiled arch - so I won't make sure that the code work on all archs).

jankratochvil: It is true there are no target-specific files/classes. But there still can be:
```
class crac {
static bool is_x86_64() {
#ifdef __x86_64__
  return true;
#else
  return false;
#endif
  };
};
if (is_x86_64()) {
  ...
}
```
Then all arch-specific code is compile-time checked on all arches.

rvansa: The ifdef in here is doing exactly that - guarding a VM option that is defined only on x86_64... though it looks like the CI is compiling against the 'zero' architecture (which lacks the VM option despite the macro is defined).

jankratochvil: I do not mind how it is done, I was trying to suggest more compile-time checks. If you disagree then let's just close this thread.

src/hotspot/cpu/x86/vm_version_x86.hpp
```diff
@@ -612,6 +612,10 @@ class VM_Version : public Abstract_VM_Version {
       return buf - buf_orig;
     }
 
+    static constexpr size_t print_buffer_length() {
+      return MAX_CPU_FEATURES;
+    }
```
TimPushkin: I believe by the time the corresponding printing function is used the resource area is already initialized. It thus should be possible to make it allocate a resource array internally instead of making the callers use this to allocate on stack. Or there can be two functions: one that uses a provided buffer and its wrapper that allocates a resource array.

src/java.base/share/native/libcrexec/linkedlist.hpp
```diff
@@ -0,0 +1,81 @@
...
+    Node *node = _head;
+    while (node) {
+      Node *next = node->_next;
+      delete node;
+      node = next;
+    }
+  }
+
+  T& add(T&& item) {
+    Node *node = new(std::nothrow) Node();
```
TimPushkin: Should do a null check after this

src/java.base/share/native/libcrexec/linkedlist.hpp
```diff
@@ -0,0 +1,81 @@
...
+  struct Node {
+    T _item;
+    Node *_next;
+  };
+
+  Node *_head;
+  Node *_tail;
+  size_t _size;
+public:
+  LinkedList(): _head(nullptr), _tail(nullptr), _size(0) {}
```
TimPushkin:
```suggestion
  Node *_head = nullptr;
  Node *_tail = nullptr;
  size_t _size = 0;
public:
```

src/java.base/share/native/libcrexec/linkedlist.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+  template<typename Func> void foreach(Func f) const {
+    Node *node = _head;
+    while (node) {
+      f(node->_item);
+      node = node->_next;
+    }
+  }
+
+  size_t size() {
```
TimPushkin:
```suggestion
  size_t size() const {
```

src/java.base/share/native/libcrexec/linkedlist.hpp
```diff
@@ -0,0 +1,81 @@
...
+    return node->_item;
+  }
+
+  template<typename Func> void foreach(Func f) const {
+    Node *node = _head;
+    while (node) {
+      f(node->_item);
+      node = node->_next;
+    }
+  }
```
TimPushkin, line 85: I don't insist but implementing an `std::iterator` does not seem too cumbersome and would allow to use the normal for loop

src/java.base/share/native/libcrexec/hashtable.hpp
```diff
@@ -129,12 +133,12 @@ T *Hashtable<T>::get(const char *key) const {
...
-bool Hashtable<T>::put(const char *key, T value) {
+bool Hashtable<T>::put(const char *key, T&& value) {
   T * const value_ptr = get(key);
   if (value_ptr == nullptr) {
     return false;
   }
-  *value_ptr = value;
+  *value_ptr = std::forward<T>(value);
   return true;
 }
```
TimPushkin: I believe this usage of `std::forward` is incorrect as `T` is deduced from the hashtable class and not from the `value` parameter, so with this change it is now not possible to pass lvalues as `value`. I think the best would be to leave `T value` as it was and replace `std::forward` with `std::move` — lvalues will be copied then moved, rvalues will be just moved.

Same for `LinkedList::add`.

src/java.base/share/native/libcrexec/linkedlist.hpp
```diff
@@ -0,0 +1,81 @@
...
+  ~LinkedList() {
+    Node *node = _head;
+    while (node) {
+      Node *next = node->_next;
+      delete node;
+      node = next;
+    }
+  }
+
+  T& add(T&& item) {
```
TimPushkin: With the current style this should be:
```suggestion
  T &add(T &&item) {
```

src/java.base/share/native/libcrexec/image_constraints.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+#include <cstring>
+#include <cstdlib>
+
+#include "crlib/crlib_image_constraints.h"
+#include "linkedlist.hpp"
+
+class ImageConstraints {
+private:
+  enum TagType {
```
TimPushkin: It can have a smaller base type, and also making it a class is more idiomatic I believe
```suggestion
  enum class TagType : std::uint8_t {
```

src/java.base/share/native/libcrexec/image_constraints.hpp
```diff
@@ -0,0 +1,167 @@
...
+    bool compare_bitmaps(const unsigned char *bitmap, size_t length) const;
+  };
+
+  LinkedList<Tag> _tags;
+  LinkedList<Constraint> _constraints;
+
+  const size_t _max_name_length = 256;
+  const size_t _max_value_length = 256;
+
+  bool check_tag(const char *name) {
```
TimPushkin: "check" has too broad meaning IMO
```suggestion
  bool has_tag(const char *name) const {
```

src/java.base/share/native/libcrexec/image_constraints.hpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    bool compare_bitmaps(const unsigned char *bitmap, size_t length) const;
+  };
+
+  LinkedList<Tag> _tags;
+  LinkedList<Constraint> _constraints;
+
+  const size_t _max_name_length = 256;
+  const size_t _max_value_length = 256;
```
TimPushkin:
```suggestion
  static constexpr size_t _MAX_NAME_LENGTH = 256;
  static constexpr size_t _MAX_VALUE_LENGTH = 256;
```

TimPushkin: A nitpick, but in the case of labels, where both names and values are strings it may be error-prone that the first refers to a length excluding the nul-char and the second refers to the full size including the nul-char. I usually use "length" for the first and "size" for the second.

src/java.base/share/native/libcrexec/image_constraints.hpp
```diff
@@ -0,0 +1,167 @@
...
+    size_t value_length = strlen(value) + 1;
+    if (strlen(name) >= _max_name_length || value_length >= _max_value_length) {
+      return false;
+    }
+    _tags.add({
+      .type = LABEL,
+      .name = strdup(name),
+      .data = strdup(value),
+      .data_length = value_length,
+    });
```
TimPushkin: `add` and `strdup`s can fail — this should be checked

TimPushkin: Error-checking in the functions below is also lacking

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+#include "image_constraints.hpp"
+#include "hashtable.hpp"
+
+#define CREXEC "crexec: "
+#define LABEL_PREFIX "label:"
+#define BITMAP_PREFIX "bitmap:"
+
+static FILE *open_tags(const char *image_location, const char *mode) {
+  char fname[PATH_MAX];
```
TimPushkin, line 39: In `crexec.cpp` we for some reason have our own conditional definition of `PATH_MAX`, I guess it caused problems at some point... Maybe create some `common.hpp` or `crexec.hpp` where `PATH_MAX` and `CREXEC` would be defined?

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,223 @@
...
+  if (snprintf(fname, sizeof(fname), "%s/tags", image_location) >= (int) sizeof(fname) - 1) {
+    fprintf(stderr, CREXEC "filename too long: %s/tags\n", image_location);
+    return nullptr;
+  }
+  FILE *f = fopen(fname, mode);
+  if (f == nullptr) {
+    fprintf(stderr, CREXEC "cannot open(%s) %s: %s\n", fname, mode, strerror(errno));
```
TimPushkin:
```suggestion
    fprintf(stderr, CREXEC "cannot open %s in %s mode: %s\n", fname, mode, strerror(errno));
```

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,223 @@
...
+int ImageConstraints::validate(const char *image_location) const {
+  FILE *f = open_tags(image_location, "r");
+  if (f == nullptr) {
+    return errno == ENOENT ? RESTORE_ERROR_NOT_FOUND : RESTORE_ERROR_NO_ACCESS;
+  }
+  char line[strlen(BITMAP_PREFIX) + _max_name_length + 1 + _max_value_length + 2];
+  LinkedList<Tag> tags;
+  while (fgets(line, sizeof(line), f)) {
+    char *eq = strchr(line, '=');
+    char *nl = strchr(eq + 1, '\n');
```
TimPushkin: I believe parsing will be broken if `=` is encountered in the tag name itself. Same for `\n` although that is less likely.

src/java.base/share/native/libcrexec/image_constraints.cpp
TimPushkin: Yep, probably this is the reason: https://github.com/rvansa/crac/actions/runs/18203262060/job/51827234536#step:10:8170

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,223 @@
...
+  }
+  fputc('\n', stderr);
+}
+
+int ImageConstraints::validate(const char *image_location) const {
+  FILE *f = open_tags(image_location, "r");
+  if (f == nullptr) {
+    return errno == ENOENT ? RESTORE_ERROR_NOT_FOUND : RESTORE_ERROR_NO_ACCESS;
+  }
+  char line[strlen(BITMAP_PREFIX) + _max_name_length + 1 + _max_value_length + 2];
```
TimPushkin: AFAIR MSVC does not support arrays of non-constexpr size

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,223 @@
...
+    *eq = 0;
+    *nl = 0;
+    assert(eq < nl);
+    if (!strncmp(line, LABEL_PREFIX, strlen(LABEL_PREFIX))) {
+      tags.add({
+        .type = LABEL,
+        .name = strdup(line + strlen(LABEL_PREFIX)),
+        .data = strdup(eq + 1),
+        .data_length = (size_t) (nl - eq),
+      });
```
TimPushkin: Missing error checking here and below (`add`, `strdup`, `malloc`)

src/java.base/share/native/libcrexec/crexec.cpp
```diff
@@ -118,9 +125,21 @@ static crlib_user_data_t user_data_extension = {
...
+  set_label,
+  set_bitmap,
+  require_label,
+  require_bitmap,
+};
+
 static const crlib_extension_t *extensions[] = {
   &restore_data_extension.header,
   &user_data_extension.header,
+  &image_constraints_extension.header,
```
TimPushkin: Since we replace the direct use of user data with this new extension it should be a bit more efficient to put the new extension above the old one

src/java.base/share/native/libcrexec/crexec.cpp
```diff
@@ -777,20 +821,25 @@ static int restore(crlib_conf_t *conf) {
...
       !env.append("CRAC_NEW_ARGS_ID", restore_data_str) ||
       (!conf->direct_map().value && !env.add_criu_option("--no-mmap-page-image"))) {
-    return -1;
+    return RESTORE_ERROR_UNKNOWN;
+  }
+
+  int error = conf->image_constraints().validate(conf->argv()[ARGV_IMAGE_LOCATION]);
+  if (error) {
+    return error;
   }
```
TimPushkin: Nitpick: let's move this before initializing `restore_data_str` as this is a part of argument validation while `restore_data_str` and below is preparation for exec-ing

src/java.base/share/native/libcrexec/hashtable.hpp
rvansa: This should be a idiomatic use of perfect forwarding, since `T` is deduced the `T &&` is a universal reference and is treated as lvalue, depending on its use. The current code copies the argument; for `std::move` I would have to pass rvalue reference (therefore the method wouldn't be usable with const value).
Sources:
* https://blog.vero.site/post/rvalue-references 
* https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers

TimPushkin: I understand what was the intention here, and thanks for the links. But I believe in this case `value` is not a universal reference because `T` gets deduced from the `Hashtable<T>` object, not from the `value` parameter.

Looks like this is discussed in your second link:
> Sometimes you can see `T&&` in a function template declaration where `T` is a template parameter, yet there’s still no type deduction.  Consider this `push_back` function in `std::vector`:
> 
> ```
> template <class T, class Allocator = allocator<T> >
> class vector {
> public:
>     ...
>     void push_back(T&& x);       // fully specified parameter type ⇒ no type deduction;
>     ...                          // && ≡ rvalue reference
> };
> ```
> Here, `T` is a template parameter, and `push_back` takes a `T&&`, yet the parameter is not a universal reference!  How can that be?
> 
> The answer becomes apparent if we look at how `push_back` would be declared outside the class. I’m going to pretend that `std::vector`’s `Allocator` parameter doesn’t exist, because it’s irrelevant to the discussion, and it just clutters up the code.  With that in mind, here’s the declaration for this version of `std::vector::push_back`:
> 
> ```
> template <class T>
> void vector<T>::push_back(T&& x);
> ```
> `push_back` can’t exist without the class `std::vector<T>` that contains it.  But if we have a class `std::vector<T>`, we already know what `T` is, so there’s no need to deduce it.
> 
> An example will help.  If I write
> 
> ```
> Widget makeWidget();             // factory function for Widget
> std::vector<Widget> vw;
> ...
> Widget w;
> vw.push_back(makeWidget());      // create Widget from factory, add it to vw
> ```
> my use of `push_back` will cause the compiler to instantiate that function for the class `std::vector<Widget>`. The declaration for that `push_back` looks like this:
> ```
> void std::vector<Widget>::push_back(Widget&& x);
> ```
> See?  Once we know that the class is `std::vector<Widget>`, the type of `push_back`’s parameter is fully determined:  it’s `Widget&&`.  There’s no role here for type deduction.

I think something like [this](https://stackoverflow.com/a/70435367/16360266) is needed to make it a universal reference.

TimPushkin: > for std::move I would have to pass rvalue reference (therefore the method wouldn't be usable with const value).

To clarify, my suggestion was to leave the parameter as `T value` so that we can safely use `std::move` to treat it as an rvalue reference and move from it. An lvalue argument will be copied into `value` and then the copy will be moved into `*value_ptr`. An rvalue argument will be moved into `value` and then moved into `*value_ptr` --- probably the compiler optimizes these moves but I haven't researched this enough.

jankratochvil: > This should be a idiomatic use of perfect forwarding, since `T` is deduced the `T &&` is a universal reference and is treated as lvalue, depending on its use. The current code copies the argument; for `std::move` I would have to pass rvalue reference (therefore the method wouldn't be usable with const value). Sources:

One can try it does not compile:
```
#include "hashtable.hpp"
int main() {
  Hashtable<int> h(nullptr, 10);
  int i = 42;
  h.put("foo", i);
}
h.cpp:5:16: error: rvalue reference to type 'int' cannot bind to lvalue of type 'int'
    5 |   h.put("foo", i);
      |                ^
src/java.base/share/native/libcrexec/hashtable.hpp:47:33: note: passing argument to parameter 'value' here
   47 |   bool put(const char *key, T&& value);
      |                                 ^
1 error generated.
```
Couldn't we use a new template parameter for the method?
[tu.patch](https://github.com/user-attachments/files/22741639/tu.patch)

rvansa: You are right, it didn't work when I've tried to pass lvalue references in a test (and thanks to testing separately I found a bug in the list impl itself...). While this is not critical for CRaC itself I took the opportunity to improve my C++ knowledge and do the structures right :) Here is a standalone check to ensure that the containers work as expected: https://gist.github.com/rvansa/5f4cb76ddb4f694d2feb9a1ffb09d630
Note that I use two methods in `linkedlist.hpp` rather than universal reference in the end, as the universal reference template would consider `{ ... }` a braced initializer list rather than a struct initializer.

src/java.base/share/native/libcrexec/linkedlist.hpp
rvansa: I personally find the lambda more convenient.

src/java.base/share/native/libcrexec/hashtable.hpp
rvansa: With the last change I can't use aggregate initializer anyway as the move ctor is defined :-/ Good ol' Java.

src/java.base/share/native/libcrexec/image_constraints.hpp
rvansa: I kept the `check_tag` name because it's checking more things in the end...

src/java.base/share/native/libcrexec/image_constraints.cpp
rvansa: Added extra checks when the tag is added.

src/java.base/share/native/libcrexec/hashtable.hpp
jankratochvil: It is an initializer list, one should not call it with an initializer list but rather just with the number itself. Up to you what calling conventions you want to support.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -102,23 +103,27 @@ static char * strchrnul(char * str, char c) {
 int crac::checkpoint_restore(int *shmid) {
   guarantee(_engine != nullptr, "CRaC engine is not initialized");
 
+  // If this is a second checkpoint we should use a clear configuration
+  if (_generation != 0 && !_engine->reset_conf()) {
```
TimPushkin: Generation numbering starts from 1, so it should be `_generation > 1`

TimPushkin: Why do we need to reset the configuration now? Is it because user data is auto-reset while the tags are not? Ideally they should act the same to make the behavior more predictable. IMO it would be better if both were kept until explicitly cleared by the user, if that is possible. But that would require updating the user data API probably...

Anyway, shouldn't the CPU features stay the same between checkpoints? Then it would be more efficient to set them only on the first checkpoint instead of resetting the whole configuration each time.

src/java.base/share/native/libcrexec/linkedlist.hpp
```diff
@@ -0,0 +1,94 @@
...
+    if (node == nullptr) {
+      return false;
+    }
+    node->_next = nullptr;
+    if (_tail) {
+      _tail->_next = node;
+      _tail = node;
+    } else {
+      _tail = node;
+    }
```
TimPushkin, line 52:
```suggestion
    if (_tail) {
      _tail->_next = node;
    }
    _tail = node;
```

src/java.base/share/native/libcrexec/hashtable.hpp
```diff
@@ -132,13 +132,13 @@ T *Hashtable<T>::get(const char *key) const {
   return nullptr;
 }
 
-template<class T>
-bool Hashtable<T>::put(const char *key, T&& value) {
+template<class T> template<typename TT>
+bool Hashtable<T>::put(const char* key, TT&& value) {
```
TimPushkin, line 136: A total nitpick, but now we have `const char *key` in the declaration and `const char* key` in the definition. When I was rewriting crexec into C++ I used the `type *x`, `type * const x` style everywhere. But since there is no linter for this in CI it seems that there is no sense in trying to keep it in a single style, it will diverge with time anyway...

src/java.base/share/native/libcrexec/image_constraints.hpp
```diff
@@ -0,0 +1,143 @@
...
+    return _constraints.add(Constraint(TagType::LABEL, strdup(name), strdup(value), strlen(value) + 1, EQUALS));
+  }
+
+  bool require_bitmap(const char* name, const unsigned char* value, size_t length_bytes, crlib_bitmap_comparison_t comparison) {
+    void* copy = malloc(length_bytes);
+    memcpy(copy, value, length_bytes);
+    return _constraints.add(Constraint(TagType::BITMAP, strdup(name), copy, length_bytes, comparison));
+  }
+
+  bool is_failed(const char* name) {
```
TimPushkin:
```suggestion
  bool is_failed(const char* name) const {
```

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,294 @@
...
+  });
+  if (present) {
+    fprintf(stderr, CREXEC "%s %s is already set\n", type, name);
+    return false;
+  } else if (strpbrk(name, "=\n")) {
+    fprintf(stderr, CREXEC "%s name must not contain '=' or newline\n", type);
+    return false;
+  }
+  if (strlen(name) >= _MAX_NAME_SIZE) {
+    fprintf(stderr, CREXEC "%s %s is too long\n", type, name);
```
TimPushkin: Would be a better UX if we specify how many symbols is allowed
```suggestion
    fprintf(stderr, CREXEC "%s %s name is too long, at most %zu chars allowed\n", type, name, _MAX_NAME_SIZE - 1;
```

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,294 @@
...
+    return false;
+  }
+  if (value_size >= _MAX_VALUE_SIZE) {
+      fprintf(stderr, CREXEC "%s %s value is too long (%zu bytes)\n", type, name, value_size);
```
TimPushkin:
```suggestion
      fprintf(stderr, CREXEC "%s %s value is too long: %zu bytes > %zu allowed\n", type, name, value_size, _MAX_VALUE_SIZE);
```

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,294 @@
...
+    return false;
+  } else if (strpbrk(name, "=\n")) {
+    fprintf(stderr, CREXEC "%s name must not contain '=' or newline\n", type);
+    return false;
+  }
+  if (strlen(name) >= _MAX_NAME_SIZE) {
+    fprintf(stderr, CREXEC "%s %s is too long\n", type, name);
+    return false;
+  }
+  if (value_size >= _MAX_VALUE_SIZE) {
```
TimPushkin: I think it should be:
```suggestion
  if (value_size > _MAX_VALUE_SIZE) {
```

src/java.base/share/native/libcrexec/image_constraints.cpp
```diff
@@ -0,0 +1,294 @@
...
+    }
+  });
+  if (fclose(f)) {
+    fprintf(stderr, CREXEC "cannot close %s/tags: %s\n", image_location, strerror(errno));
+    return false;
+  }
+  return true;
+}
+
+static inline unsigned char from_hex(char c, bool& err) {
```
TimPushkin: I'd say it's a better style to take a pointer when you plan to write into the argument since that becomes more obvious at the call site, but this is opinionated

src/hotspot/share/runtime/crac.cpp
rvansa: Hmm, the reset was there to prevent errors coming from setting the CPU features second time, but you're right that rather than that we could just avoid setting them in the second checkpoint.

src/java.base/share/native/libcrexec/hashtable.hpp
rvansa: To me, it makes little sense to use `T&& x` and `T *x`, and we should be consistent at least on the file level. And yes, without linting I'll always forget it somewhere...

While I personally prefer `T *x`, from some recent mainline JDK PRs I think that the project prefers `T* x`, so let's use that.

PR #295, opened 2026-02-10
8377562: [CRaC] Add better logging to RemoteJmxTest

TimPushkin:
~~Makes `RemoteJmxTest` log stderr of the child process~~. Refactors CRaC tests to make coordination with exec process simpler and more robust.

---------
### Progress

### Issue
 * [JDK-8377562](https://bugs.openjdk.org/browse/JDK-8377562): [CRaC] Add better logging to RemoteJmxTest (**Task** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/295/head:pull/295` \
`$ git checkout pull/295`

`$ git checkout pull/295` \
`$ git pull https://git.openjdk.org/crac.git pull/295/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 295`

`$ git pr show -t 295`

</details>

<a href="https://git.openjdk.org/crac/pull/295.diff">https://git.openjdk.org/crac/pull/295.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/295#issuecomment-3877702028)
</details>


test/jdk/jdk/crac/RemoteJmxTest.java
```diff
@@ -70,37 +67,44 @@ public class RemoteJmxTest implements CracTest {
...
         try {
-            var reader = new BufferedReader(new InputStreamReader(process.output()));
-            waitForString(reader, BOOTED);
+            final var bootedFuture = new CompletableFuture<>();
+            final var restoredFuture = new CompletableFuture<>();
+            process.watch(
+                    (line) -> {
+                        System.out.println("[CHILD STDOUT] " + line);
+                        if (!bootedFuture.isDone()) {
+                            if (BOOTED.equals(line)) bootedFuture.complete(null);
```
rvansa: This code looks like as if you expect that the messages could be printed multiple times; if that happens this should fail the test (for simplicity I wouldn't mind if it fails by timeout in the end, though, as long as the reason is printed earlier in the log).

I wonder if this could be consolidated with `CracProcess.waitForStdout()` (the least the lines are prefixed here. Maybe the best interface would be
```
CountDownLatch<Void> watch(String line, boolean positiveMatch);
```
(called repeatedly to get multiple latches), and `waitForStdout` would use that.

TimPushkin: I wanted to use `waitForStdout` initially as well but with the existing implementation it is not possible to use multiple calls to `watch`/`waitForStdout`/`outputAnalyzer` — only the first of such calls will receive the output.

I've refactored our test code now (+ did some cleanup) but since it is pretty big it'll probably require a few fixups so I'm converting this to draft for now.

test/lib/jdk/test/lib/crac/CracProcess.java
```diff
@@ -1,66 +1,171 @@
...
+                        lines.add(line);
+                        lines.notifyAll();
+                    }
+                }
+            }).process();
+        }
+
+        @Override
+        public void close() {
+            future.cancel(true);
```
rvansa, line 81: You should probably call `lines.notifyAll()` in here and throw an `InterruptedException` from  `waitForLine`.

test/lib/jdk/test/lib/crac/CracBuilderBase.java
```diff
@@ -45,7 +45,14 @@ public abstract class CracBuilderBase<T extends CracBuilderBase<T>> {
     // when the whole suite is executed.
     private static final Class<CracTestArg> dummyWorkaround = CracTestArg.class;
 
-    boolean verbose = true;
+    static void log(String fmt, Object... args) {
```
rvansa: I am not so sure how useful is this wrapper in the end, the code could probably just `System.err.print...` .
But if we keep this, it should probably move to `CracProcess`.

test/lib/jdk/test/lib/crac/CracProcess.java
```diff
@@ -101,72 +206,46 @@ private void waitForFileCreated(WatchService watcher, Path dir, Predicate<Path>
...
-        assertTrue(builder.captureOutput, "Output must be captured with .captureOutput(true)");
-        pump(process.getInputStream(), outputConsumer);
-        pump(process.getErrorStream(), errorConsumer);
-        return this;
+        CracBuilderBase.log("Process %d completed with exit value %d", process.pid(), exitValue);
     }
 
-    public void waitForStdout(String str) throws InterruptedException {
-        waitForStdout(str, true);
+    public void waitForStdout(Predicate<String> predicate, long timeoutSec) throws InterruptedException, EOFException, TimeoutException {
```
rvansa, line 223: Do we really need to pass the timeout, or would some common-sense default suffice?

rvansa (commented):
If it were not for the default check of the exit status I would go with 3) (return OA from `doRestore()` & friends)  but while the `toAnalyze` suffix still seems awkward to me ('analysis' is rather pompous word for string match) I won't have objections.

Minor comments below.

rvansa (approved):

PR #300, opened 2026-03-03
Merge jdk:jdk-27+11

TimPushkin:
Merges `jdk-27+11`.

Main conflicts (see resolutions in https://github.com/TimPushkin/crac/commit/2c17e065fce0da346410d09d354603805a1d9749):
- [JDK-8374170](https://bugs.openjdk.org/browse/JDK-8374170) — refactoring of `sun.nio.ch.Poller` and its implementations. It adds FD-closing code to the mainline so I reused that instead of what we had in our fork. Overall, I didn't exactly copy the old implementation so please pay attention to the related parts.
- [JDK-8377507](https://bugs.openjdk.org/browse/JDK-8377507) — trivial conflicts in `vm_version_x86.hpp`, but I want to bring attention to Leyden now recording and checking CPU features as well, so there is some duplication between their code and ours. Maybe we'll be able to merge the implementations better at some point, currently they just exist completely separately.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/300/head:pull/300` \
`$ git checkout pull/300`

`$ git checkout pull/300` \
`$ git pull https://git.openjdk.org/crac.git pull/300/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 300`

`$ git pr show -t 300`

</details>

<a href="https://git.openjdk.org/crac/pull/300.diff">https://git.openjdk.org/crac/pull/300.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/300#issuecomment-3990878821)
</details>


src/java.base/share/classes/sun/nio/ch/Poller.java
```diff
@@ -24,68 +24,135 @@
...
+    // shutdown (if supported by poller group)
+    private volatile boolean shutdown;
+
+    private static final JDKResource resource = new JDKResource() {
+        @Override
+        public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+            // TODO: handle existing/concurrently-starting polling. Currently this is left for the
+            //  user to handle, and if not handled, some polling threads wake up (possibly
+            //  prematurely), some remain parked forever. But ideally new polling attempts should
+            //  fail and existing ones should abort the checkpoint.
```
TimPushkin, line 80: Our old code had the same issue so fixing this does not belong to a merge, but we should address this at some point

src/java.base/share/classes/sun/nio/ch/Poller.java
```diff
@@ -24,68 +24,135 @@
...
+            POLLER_GROUP = null;
+        }
+
+        @Override
+        public void afterRestore(Context<? extends Resource> context) {
+            POLLER_GROUP = createPollerGroup();
+        }
+    };
+    static {
+        Core.Priority.NORMAL.getContext().register(resource);
```
TimPushkin, line 91: The old code used this priority but to me it seems that pollers should be closed after sockets and whatever else in lower-than-normal priorities can use them. But I don't think we should change such things during a merge.

rvansa: Your reasoning might be correct, but there might be always some chicken-egg problem in the implementation between these resources. So far I've added resources to the `NORMAL` priority unless there was a specific need to place them elsewhere; I always think twice before introducing another priority level.

rvansa:
TBH I wouldn't hurry into unifying Leyden and ours code for the CPU features; we already have rather painful divergence across LTS branches in the downstream. Unless this results in reducing the divergence, not worth the effort, IMO.

rvansa (approved):

PR #309, opened 2026-04-16
8382337: [CRaC] Add CPU features nicknames for microarch levels

antonvoznia:
x86_64 defines microarchitecture levels v1-v4: https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels

Let’s allow users to set -XX:CPUFeatures=x86-64-vX in addition to the current options (generic, native, ignore and bitmask). 

---------

---------
### Progress

### Issue
 * [JDK-8382337](https://bugs.openjdk.org/browse/JDK-8382337): [CRaC] Add CPU features nicknames for microarch levels (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/309/head:pull/309` \
`$ git checkout pull/309`

`$ git checkout pull/309` \
`$ git pull https://git.openjdk.org/crac.git pull/309/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 309`

`$ git pr show -t 309`

</details>

<a href="https://git.openjdk.org/crac/pull/309.diff">https://git.openjdk.org/crac/pull/309.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/309#issuecomment-4260110827)
</details>

### DIFF
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -964,19 +964,17 @@ VM_Version::VM_Features VM_Version::CPUFeatures_parse(const char *str) {
   return {};
 #else // LINUX
 
-  VM_Version::VM_Features retval;
 #ifdef AMD64
   const char *microarch_prefix="x86-64-v";
   const size_t microarch_prefix_len = strlen(microarch_prefix);
   if (strncmp(str, microarch_prefix, strlen(microarch_prefix)) == 0 &&
     (strlen(str) == microarch_prefix_len + 1)) {
-      set_microarch_features(str[microarch_prefix_len], retval);
-
-      return retval;
+      return get_microarch_features(str[microarch_prefix_len]);
   }
 #endif
 
   int count = VM_Version::VM_Features::features_bitmap_element_count();
+  VM_Version::VM_Features retval;
   const char *str_orig = str;
   for (int idx = 0;; ++idx) {
     static_assert(sizeof(uint64_t) == sizeof(unsigned long long), "unexpected arch");
@@ -1007,7 +1005,8 @@ VM_Version::VM_Features VM_Version::CPUFeatures_parse(const char *str) {
 }
 
 #if defined(LINUX) && defined(AMD64)
-void VM_Version::set_microarch_features(const char microarch_level, VM_Version::VM_Features &features) {
+void VM_Version::get_microarch_features(const char microarch_level) {
+  VM_Version::VM_Features features;
   switch(microarch_level) {
     case '4':
       features.set_feature(CPU_AVX512F);
@@ -1049,6 +1048,8 @@ void VM_Version::set_microarch_features(const char microarch_level, VM_Version::
       vm_exit_during_initialization(err_msg("VM option 'CPUFeatures=x86-64-v%c' is invalid: "
         "supported levels are x86-64-v1 through x86-64-v4", microarch_level));
   }
+
+  return features;
 }
 #endif
 
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -1007,7 +1007,7 @@ VM_Version::VM_Features VM_Version::CPUFeatures_parse(const char *str) {
 }
 
 #if defined(LINUX) && defined(AMD64)
-static void VM_Version::set_michroarch_features(const char microarch_level, VM_Version::VM_Features &features) {
+void VM_Version::set_michroarch_features(const char microarch_level, VM_Version::VM_Features &features) {
   switch(microarch_level) {
     case '4':
     case '3':
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -944,17 +944,7 @@ VM_Version::VM_Features VM_Version::CPUFeatures_parse(const char *str) {
     // 32-bit x86 cannot rely on anything.
     VM_Version::VM_Features retval;
 #ifdef AMD64
-      // The following options are all in /proc/cpuinfo of one of the first 64-bit CPUs - Atom D2700 (and Opteron 1352): https://superuser.com/q/1572306/1015048
-    retval.set_feature(CPU_SSE); // enabled in 'gcc -Q --help=target', used by OpenJDK
-    retval.set_feature(CPU_SSE2); // enabled in 'gcc -Q --help=target', required by OpenJDK
-    retval.set_feature(CPU_FXSR); // enabled in 'gcc -Q --help=target', not used by OpenJDK
-    retval.set_feature(CPU_MMX); // enabled in 'gcc -Q --help=target', used only by 32-bit x86 OpenJDK
-    retval.set_feature(CPU_TSC); // not used by gcc, used by OpenJDK
-    retval.set_feature(CPU_CX8); // gcc detects it to set cpu "pentium" (=32-bit only), used by OpenJDK
-    retval.set_feature(CPU_CMOV); // gcc detects it to set cpu "pentiumpro" (=32-bit only), used by OpenJDK
-    retval.set_feature(CPU_FLUSH); // ="clflush" in cpuinfo, not used by gcc, required by OpenJDK
-    // CPU_MOVBE is disabled in 'gcc -Q --help=target' and for example i7-720QM does not support it
-    // CPU_LAHFSAHF is disabled in 'gcc -Q --help=target' and "Early Intel Pentium 4 CPUs with Intel 64 support ... lacked the LAHF and SAHF instructions"
+    set_generic_features(retval);
 #endif // AMD64
     return retval;
 #endif // LINUX
@@ -1036,13 +1026,7 @@ void VM_Version::get_microarch_features(const char microarch_level) {
       features.set_feature(CPU_SSSE3);
       /* FALLTHROUGH */
     case '1':
-      features.set_feature(CPU_CMOV);
-      features.set_feature(CPU_CX8);
-      features.set_feature(CPU_FPU);
-      features.set_feature(CPU_FXSR);
-      features.set_feature(CPU_MMX);
-      features.set_feature(CPU_SSE);
-      features.set_feature(CPU_SSE2);
+      set_generic_features(features);
       break;
     default:
       vm_exit_during_initialization(err_msg("VM option 'CPUFeatures=x86-64-v%c' is invalid: "
@@ -3530,9 +3514,6 @@ VM_Version::VM_Features VM_Version::CpuidInfo::feature_flags() const {
     vm_features.set_feature(CPU_CMOV);
   if (std_cpuid1_edx.bits.clflush != 0)
     vm_features.set_feature(CPU_FLUSH);
-  if (std_cpuid1_edx.bits.fpu != 0 || (is_amd_family() &&
-    ext_cpuid1_edx.bits.fpu != 0))
-    vm_features.set_feature(CPU_FPU);
   // clflush should always be available on x86_64
   // if not we are in real trouble because we rely on it
   // to flush the code cache.
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -1015,6 +1015,7 @@ void VM_Version::set_microarch_features(const char microarch_level, VM_Version::
       features.set_feature(CPU_AVX512CD);
       features.set_feature(CPU_AVX512DQ);
       features.set_feature(CPU_AVX512VL);
+      /* FALLTHROUGH */
     case '3':
       features.set_feature(CPU_AVX);
       features.set_feature(CPU_AVX2);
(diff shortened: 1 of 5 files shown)


src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -963,8 +963,20 @@ VM_Version::VM_Features VM_Version::CPUFeatures_parse(const char *str) {
   vm_exit_during_initialization("This OS does not support any arch-specific -XX:CPUFeatures options");
   return {};
 #else // LINUX
-  int count = VM_Version::VM_Features::features_bitmap_element_count();
+
   VM_Version::VM_Features retval;
+#ifdef AMD64
+  const char *microarch_prefix="x86-64-v";
+  const size_t microarch_prefix_len = strlen(microarch_prefix);
+  if (strncmp(str, microarch_prefix, strlen(microarch_prefix)) == 0 &&
```
rvansa: replace the `strlen(...)` in here

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -994,6 +1006,52 @@ VM_Version::VM_Features VM_Version::CPUFeatures_parse(const char *str) {
 #endif // LINUX
 }
 
+#if defined(LINUX) && defined(AMD64)
+void VM_Version::set_microarch_features(const char microarch_level, VM_Version::VM_Features &features) {
```
rvansa: Could this return `VM_Features` rather than just modify reference?

src/hotspot/cpu/x86/vm_version_x86.hpp
```diff
@@ -105,7 +105,8 @@ class VM_Version : public Abstract_VM_Version {
   union StdCpuid1Edx {
     uint32_t value;
     struct {
-      uint32_t          : 4,
+      uint32_t fpu      : 1,
```
rvansa: Could you add a link to the struct with some reference saying where the bit position comes from?

rvansa (commented):
I see that you've followed the information on wiki, but from a practical point of you I am not sure if the FPU is present on modern CPUs. I will check AWS instance types and report my findings.
Also, I am somewhat concerned that `v1` is not strictly superset of `generic`; we shall check out the FLUSH flag. Maybe these two could be equivalent?

Please add a test for the parsing; you can extend `CheckCPUFeaturesTest` or create a standalone one.

antonvoznia:
@rvansa , Thank you for reviewing!

> I see that you've followed the information on wiki, but from a practical point of you I am not sure if the FPU is present on modern CPUs. I will check AWS instance types and report my findings. 

I verified on my x86 vm
```
$ cat /proc/cpuinfo | grep -i fpu
fpu             : yes
fpu_exception   : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr....
```

> Also, I am somewhat concerned that `v1` is not strictly superset of `generic`; we shall check out the FLUSH flag. Maybe these two could be equivalent?

I followed this pages
https://en.opensuse.org/X86-64_microarchitecture_levels
and
https://gitlab.com/x86-psABIs/x86-64-ABI ,
there is no FLUSH mentioned.
Do you think we need to check it out for some reason?

I also didn't include flags OSFXSR and SCE. To have them, we need priveledge access to registers CR4 and MRS
CR4
https://stackoverflow.com/a/10186910
MRS
https://wiki.osdev.org/Model_Specific_Registers

> Please add a test for the parsing; you can extend `CheckCPUFeaturesTest` or create a standalone one.

I'll add this

rvansa:
I have verified that all Intel/AMD CPUs used for AWS instance types have the `FPU` flag in `lscpu`, so that should be good.

@jankratochvil Could you comment why is the FLUSH (clflush) "required by OpenJDK"?

rvansa:
Actually the reason is documented in `VM_Version::supports_clflush()` - let's just add it to the required set, please.

src/hotspot/cpu/x86/vm_version_x86.hpp
```diff
@@ -181,7 +182,8 @@ class VM_Version : public Abstract_VM_Version {
   union ExtCpuid1Edx {
     uint32_t value;
     struct {
-      uint32_t           : 22,
```
jankratochvil, line 184: `fpu` should not be added anywhere. It is only set, but nothing ever reads or queries it. The only code that queries it is already disabled.
```
// FPU is always present on i686+: _features.supports_feature(CPU_FPU) &&
```

jankratochvil:
There is an existing message in the code:
```
      log_error(crac)("Restore failed due to incompatible or missing CPU features, try using -XX:CPUFeatures=%s on checkpoint.", image_features.print_numbers());
```
The question is whether it should not say (which would make it a bit long and complicated suggestion for the user):
```
      log_error(crac)("Restore failed due to incompatible or missing CPU features, try using -XX:CPUFeatures=x86-64-v%c or -XX:CPUFeatures=%s on checkpoint.", image_features.print_level_TBD(), image_features.print_numbers());
```
or even just (which would make the performance sub-optimal due to some CPU features possibly missing):
```
      log_error(crac)("Restore failed due to incompatible or missing CPU features, try using -XX:CPUFeatures=x86-64-v%c on checkpoint.", image_features.print_level_TBD());
```
As otherwise I have some doubts anyone will ever find the `x86-64-v%c` possibility exists (nobody reads the docs) and which level is his/her CPU.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -994,6 +1006,52 @@ VM_Version::VM_Features VM_Version::CPUFeatures_parse(const char *str) {
...
+      /* FALLTHROUGH */
+    case '1':
+      features.set_feature(CPU_CMOV);
+      features.set_feature(CPU_CX8);
+      features.set_feature(CPU_FPU);
+      features.set_feature(CPU_FXSR);
+      features.set_feature(CPU_MMX);
+      features.set_feature(CPU_SSE);
+      features.set_feature(CPU_SSE2);
+      break;
```
jankratochvil, line 1031:
```suggestion
      // The following options are all in /proc/cpuinfo of one of the first 64-bit CPUs - Atom D2700 (and Opteron 1352): https://superuser.com/q/1572306/1015048
      features.set_feature(CPU_FLUSH);
      features.set_feature(CPU_TSC);
      break;
```
I do not know why they are not in the `x86_64-v1` official set. I have already listed them for `"generic"` as they are required or useful for OpenJDK.

rvansa:
The more I dig into this, the more I am worried that the selected set might not be the best choice, as we might unintentionally prohibit use of an instruction set that has significant performance impact.

Also, I've found that the `CPU_OSXSAVE` bit is not reported on AWS instances, even though these are otherwise `v3`. I could not find a reasonable explanation beyond AWS Nitro virtualization quirk (AVX shall still work just fine). @jankratochvil could correct me but I wonder if forcing that bit to be set in mask wouldn't fail deployment in practice.

src/hotspot/cpu/x86/vm_version_x86.cpp
antonvoznia: done

src/hotspot/cpu/x86/vm_version_x86.hpp
antonvoznia: Eventually, I removed the FPU as @jankratochvil suggested.

src/hotspot/cpu/x86/vm_version_x86.hpp
antonvoznia: done

src/hotspot/cpu/x86/vm_version_x86.cpp
antonvoznia: Done, but a bit in different way.
I used common flags for v1 set and generic.

jankratochvil:
> Also, I've found that the `CPU_OSXSAVE` bit is not reported on AWS instances, even though these are otherwise `v3`. I could not find a reasonable explanation beyond AWS Nitro virtualization quirk (AVX shall still work just fine).

Missing `OSXSAVE` means a userland program must not use any AVX instructions. This is well defined. It may be an AWS virtualization bug, but that would only affect the performance of their virtual machines.

> @jankratochvil could correct me but I wonder if forcing that bit to be set in mask wouldn't fail deployment in practice.

Yes, any bit which is set by -XX:CPUFeatures=foobar and which is not present in cpuid of that process will be rejected:
```
$ java -XX:CPUFeatures=0,1 
Specified -XX:CPUFeatures=0x0,0x1; this machine's CPU features are 0x7ff3fff9dfdfbff,0xfcc; missing features of this CPU are 0x0,0x1 = avx10_2
If you are sure it will not crash you can override this check by -XX:+UnlockExperimentalVMOptions -XX:+IgnoreCPUFeatures .
```

jankratochvil:
> The more I dig into this, the more I am worried that the selected set might not be the best choice, as we might unintentionally prohibit use of an instruction set that has significant performance impact.

It looks to me the set of flags from ABI for v1..v4 is pretty small compared to what real processors offers.

Interesting is that it looks to me GCC also does not support more:

```
gcc -march=x86-64-v3 -Q --help=target | grep enabled
  -m128bit-long-double        		[enabled]
  -m64                        		[enabled]
  -m80387                     		[enabled]
...
```
Not all these flags can be mapped to CPUID bits (for example `-m64`). I haven't seen there any GCC flag which would map to a CPUID bit which is not a part of the v1..v4 ABI CPUID bits. But I did not go bit by bit.

Maybe we should add some bits like `FLUSH` and `TSC` even to v2..v4 as those bits are always set on real world CPUs despite not required by the v2..v4 ABI.

And at the end one should benchmark whether the possibly missing bits really matter. It looks all as much more work.

antonvoznia:
@jankratochvil 
> Maybe we should add some bits like FLUSH and TSC even to v2..v4 as those bits are always set on real world CPUs despite not required by the v2..v4 ABI.

Just FLUSH and TSC, or are there others you have in mind?

Because FLUSH and TSC are already in v1...v4, from the spec
`The levels are cumulative in the sense that features from previous levels are implicitly included in later levels.`

jankratochvil:
@antonvoznia 
> Just FLUSH and TSC, or are there others you have in mind?

I have been searching now and I have found for example `CLMUL`.
- It is not contained even in `x86-64-v4`.
- It improves code performance.
- All newer processors (I think since `x86-64-v3`) do have `CLMUL`.

But I had some difficulty finding such CPU feature which is:
- present in all recent CPUs
- not listed by any `x86-64-v?`
- having some impact on the JVM JIT functionality/performance

OTOH the performance impact of `CLMUL` is probably minimal so it would be nice to benchmark the current `x86-64-v4` vs. `native` if there is some real difference.

> Because FLUSH and TSC are already in v1...v4, from the spec `The levels are cumulative in the sense that features from previous levels are implicitly included in later levels.`

`FLUSH` and `TSC` are not in `x86-64-v?` specification. I have added them to `x86-64-v1` (let's call it `x86-64-v1-CRaC`) because they are in real world supported by any `x86-64-v1` CPU. This way I think we should add `CLMUL` to `x86-64-v3-CRaC`. And maybe some other features to some `x86-64-v?-CRaC` levels.

But it is some work. OTOH it is safe, if we add it by mistake it will just print an error refusing to run on such CPU missing our wrongly chosen CPU Feature. The question is how much of such work is worth of such performance improvement, if measurable.

test/jdk/jdk/crac/engine/CheckCPUFeaturesTest.java
```diff
@@ -61,11 +61,6 @@
  * @run driver jdk.test.lib.crac.CracTest x86-64-v3 exact           fail-x86
  * @run driver jdk.test.lib.crac.CracTest x86-64-v3 skip            fail
  * @run driver jdk.test.lib.crac.CracTest x86-64-v3 skip-experimental pass
- * @run driver jdk.test.lib.crac.CracTest x86-64-v4 compatible      pass
- * @run driver jdk.test.lib.crac.CracTest x86-64-v4 exact           fail-x86
```
jankratochvil: Could you comment why is it removed?

antonvoznia: My test environment and GH actions VM doesn't support this instructions, the test fails.
I just removed them.
On some CPUs it will work.

jankratochvil: It will not test v4 and on v2 CPUs it will falsely FAIL.
Couldn't you rather check for the string "`missing features of this CPU are`" and make it PASS in such case? There is no easy way for "NOT RUN" with Jtreg unless you want to code it as `test/jtreg-ext/requires/VMProps.java` and `@requires` such as `jdk.static` etc.

antonvoznia: Done.
@jankratochvil , I added it as separate test with @requires as you suggested.
I tested it in AWS with amd M5
```
Directory "JTwork" not found: creating
Directory "JTreport" not found: creating
Test results: passed: 2
Report written to /home/ubuntu/crac/JTreport/html/report.html
Results written to /home/ubuntu/crac/JTwork
```

antonvoznia:
Thank you @rvansa  and @jankratochvil for your help and review on this PR.
The current CPU features don't show any visible performance effect.
We've decided to explore another combination of CPU flags that could impact performance, which will be tracked as a follow-up task.

Closing this PR.

PR #311, opened 2026-04-29
8383556: [CRaC] CRaCHeapErgonomics VM option

rvansa:
The heap size is be determined in the environment where the application is booted and checkpointed. When CRaC is used to produce a ‘generic’ image where we don’t know the target needs & constraints, we cannot set the heap size ahead.

One solution for this is to size the heap reasonably big (e.g. assuming that a microservice won’t be run with more than 32 GB of heap, also to fit withing CompressedOops) and apply a secondary limit after restore, similar to -XX:CRaCMaxHeapSizeAfterRestore applied before checkpoint. The value of the limit shall be determined from the new environment; the secondary limit will be triggered using -XX:+CRaCHeapErgonomics.

We can set this option on by default. In the more common example (same checkpoint and restore host) it has no effect. If the restore host is upsized this won't have any effect either. If the restore host is smaller, this can prevent OOM kill through container/OS. With the default conservative `MaxRAMPercentage=25` this could lead to OOME in the application sooner than expected - this is something that can be diagnosed without too much effort.

---------

---------
### Progress

### Issue
 * [JDK-8383556](https://bugs.openjdk.org/browse/JDK-8383556): [CRaC] CRaCHeapErgonomics VM option (**New Feature** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/311/head:pull/311` \
`$ git checkout pull/311`

`$ git checkout pull/311` \
`$ git pull https://git.openjdk.org/crac.git pull/311/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 311`

`$ git pr show -t 311`

</details>

<a href="https://git.openjdk.org/crac/pull/311.diff">https://git.openjdk.org/crac/pull/311.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/311#issuecomment-4342033708)
</details>


rvansa:
I have a feeling that the limit is not logged extensively enough - looking forward to suggestions.

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2005,6 +2005,9 @@ const int ObjectAlignmentInBytes = 8;
   product(size_t, CRaCMaxHeapSizeBeforeCheckpoint, 0, "Maximum size "       \
           "of heap before checkpoint. By default equals to -Xmx.")          \
                                                                             \
+  product(bool, CRaCHeapErgonomics, true, RESTORE_SETTABLE, "Recalculate "  \
+          "heap limit after restore.")                                      \
```
TimPushkin: Maybe these two fit gc_globals.hpp better?

Or even `g1_globals.hpp` since currently only G1 implements these.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -1617,9 +1623,9 @@ void Arguments::set_heap_size() {
       if (reasonable_max > max_coop_heap) {
         if (FLAG_IS_ERGO(UseCompressedOops) && has_ram_limit) {
           log_debug(gc, heap, coops)("UseCompressedOops disabled due to "
-                                     "max heap %zu > compressed oop heap %zu. "
-                                     "Please check the setting of MaxRAMPercentage %5.2f.",
-                                     reasonable_max, (size_t)max_coop_heap, MaxRAMPercentage);
+                                      "max heap %zu > compressed oop heap %zu. "
+                                      "Please check the setting of MaxRAMPercentage %5.2f.",
+                                      reasonable_max, (size_t)max_coop_heap, MaxRAMPercentage);
```
TimPushkin: Let's not change whitespaces of the mainline code — to minimize the difference with the mainline

test/jdk/jdk/crac/ContainerOOMETest.java
```diff
@@ -40,11 +41,29 @@
  * @library /test/lib
  * @modules java.base/jdk.internal.platform
  * @build ContainerOOMETest
- * @run driver/timeout=120 jdk.test.lib.crac.CracTest
+ * @run driver/timeout=120 jdk.test.lib.crac.CracTest    -  -  -  -
+ * @run driver/timeout=120 jdk.test.lib.crac.CracTest 128M  -  -  -
+ * @run driver/timeout=120 jdk.test.lib.crac.CracTest 128M  - 1G  -
+ * @run driver/timeout=120 jdk.test.lib.crac.CracTest 128M  - 4G  -
+ * @run driver/timeout=120 jdk.test.lib.crac.CracTest 128M  - 1G  -
```
TimPushkin: Runs on lines 46 and 48 are duplicated

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -1540,6 +1540,43 @@ static size_t clamp_by_size_t_max(uint64_t value) {
...
 
+size_t Arguments::calculate_default_heap_size(size_t avail_mem) {
+  uint64_t min_memory = (uint64_t)(((double)avail_mem * MinRAMPercentage) / 100);
+  uint64_t max_memory = (uint64_t)(((double)avail_mem * MaxRAMPercentage) / 100);
+
+  const size_t reasonable_min = clamp_by_size_t_max(min_memory);
+  size_t reasonable_max = clamp_by_size_t_max(max_memory);
+
+  // There is no straightforward API to get the default value of MaxHeapSize
+  const size_t default_max_heap_size = ScaleForWordSize(96*M);
```
TimPushkin: We could put this into `gc_globals.hpp` as a separate constant and make `MaxHeapSize` use that as the default. `globals.hpp` already has a few `*StringTableSize` constants like this.

src/hotspot/share/runtime/globals.hpp
rvansa: Let's go for `gc_globals.hpp`, I hope that some day this could extends to other GCs as well.

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -1552,49 +1587,18 @@ void Arguments::set_heap_size() {
...
           // matches compressed oops printing flags
           log_debug(gc, heap, coops)("HeapBaseMinAddress must be at least %zu "
-                                     "(%zuG) which is greater than value given %zu",
-                                     DefaultHeapBaseMinAddress,
-                                     DefaultHeapBaseMinAddress/G,
-                                     HeapBaseMinAddress);
+                                      "(%zuG) which is greater than value given %zu",
+                                      DefaultHeapBaseMinAddress,
+                                      DefaultHeapBaseMinAddress/G,
+                                      HeapBaseMinAddress);
```
TimPushkin:
```suggestion
                                     "(%zuG) which is greater than value given %zu",
                                     DefaultHeapBaseMinAddress,
                                     DefaultHeapBaseMinAddress/G,
                                     HeapBaseMinAddress);
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -505,6 +505,9 @@ void VM_Crac::doit() {
   }
 
   memory_restore();
+  if (CRaCHeapErgonomics) {
+
+  }
```
```suggestion
```

TimPushkin (changes requested):

rvansa:
The rebase had me move the code from `arguments.[c|h]pp` to `gcArguments.[c|h]pp` but no significant changes.

TimPushkin (approved):

rvansa:
Ignoring the infra failures on Windows...

PR #312, opened 2026-05-04
8383789: [CRaC] Add reading to Java-side score API

TimPushkin:
Adds new methods to `Score`:
- `getScore` – read the current score at any moment;
- `addScoreProvider` — subscribe to score reading to update the score just before that; needed because the previous pattern of updating the score in `beforeCheckpoint` is not sufficient now (score can be read not only on checkpoint).

Replaces `resetAll()` with `removeScore(String metric)` — the new method is more traditional and can be more efficient in some cases, the old behavior can be achieved by removing all metrics returned by `getScore`.

---------

---------
### Progress

### Issue
 * [JDK-8383789](https://bugs.openjdk.org/browse/JDK-8383789): [CRaC] Add reading to Java-side score API (**Enhancement** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/312/head:pull/312` \
`$ git checkout pull/312`

`$ git checkout pull/312` \
`$ git pull https://git.openjdk.org/crac.git pull/312/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 312`

`$ git pr show -t 312`

</details>

<a href="https://git.openjdk.org/crac/pull/312.diff">https://git.openjdk.org/crac/pull/312.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/312#issuecomment-4370147856)
</details>

### DIFF
--- a/src/hotspot/share/prims/jvm.cpp
+++ b/src/hotspot/share/prims/jvm.cpp
@@ -3676,27 +3676,27 @@ JVM_ENTRY(jboolean, JVM_IsCRaCScoreSupported(JNIEnv *env))
   return crac::is_image_score_supported();
 JVM_END
 
-JVM_ENTRY(jobjectArray, JVM_GetCRaCScore(JNIEnv *env))
+JVM_ENTRY(jobjectArray, JVM_GetCRaCScores(JNIEnv *env))
   ResourceMark rm;
-  const GrowableArray<crac::score> score = crac::collect_image_score_from_jvm();
-  const objArrayOop score_flat_oop = oopFactory::new_objectArray(2 * score.length(), CHECK_NULL);
-  const objArrayHandle score_flat(THREAD, score_flat_oop);
-  for (int i = 0; i < score.length(); i++) {
-    const crac::score &score_i = score.at(i);
-    Handle metric = java_lang_String::create_from_str(score_i.metric, CHECK_NULL);
+  const GrowableArray<crac::score> scores = crac::get_image_scores_from_jvm();
+  const objArrayOop scores_flat_oop = oopFactory::new_objectArray(2 * scores.length(), CHECK_NULL);
+  const objArrayHandle scores_flat(THREAD, scores_flat_oop);
+  for (int i = 0; i < scores.length(); i++) {
+    const crac::score &score = scores.at(i);
+    Handle metric = java_lang_String::create_from_str(score.metric, CHECK_NULL);
     jvalue value_jval;
-    value_jval.d = score_i.value;
+    value_jval.d = score.value;
     oop value = java_lang_boxing_object::create(T_DOUBLE, &value_jval, CHECK_NULL);
-    score_flat->obj_at_put(2 * i, metric());
-    score_flat->obj_at_put(2 * i + 1, value);
+    scores_flat->obj_at_put(2 * i, metric());
+    scores_flat->obj_at_put(2 * i + 1, value);
   }
-  return checked_cast<jobjectArray>(JNIHandles::make_local(THREAD, score_flat()));
+  return checked_cast<jobjectArray>(JNIHandles::make_local(THREAD, scores_flat()));
 JVM_END
 
-JVM_ENTRY(void, JVM_RecordCRaCScore(JNIEnv *env, jobjectArray metrics, jdoubleArray values))
+JVM_ENTRY(void, JVM_RecordCRaCScores(JNIEnv *env, jobjectArray metrics, jdoubleArray values))
   // Not supported is ok: this is used internally, helps avoid another JNI call checking the support
-  if (crac::is_image_score_supported() && !crac::record_image_score(metrics, values)) {
-    THROW_MSG(vmSymbols::java_lang_RuntimeException(), err_msg("CRaC engine failed to record score"));
+  if (crac::is_image_score_supported() && !crac::record_image_scores(metrics, values)) {
+    THROW_MSG(vmSymbols::java_lang_RuntimeException(), err_msg("CRaC engine failed to record scores"));
   }
 JVM_END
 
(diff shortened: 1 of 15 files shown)


src/hotspot/share/prims/jvm.cpp
```diff
@@ -3676,8 +3676,30 @@ JVM_ENTRY(jboolean, JVM_IsCRaCScoreSupported(JNIEnv *env))
...
+  const GrowableArray<crac::score> score = crac::collect_image_score_from_jvm();
+  const objArrayHandle score_pairs = oopFactory::new_objArray_handle(Universe::objectArrayKlass(), score.length(), CHECK_NULL);
+  for (int i = 0; i < score.length(); i++) {
+    const crac::score &score_i = score.at(i);
+    Handle metric = java_lang_String::create_from_str(score_i.metric, CHECK_NULL);
+    jvalue value_jval;
+    value_jval.d = score_i.value;
+    oop value_oop = java_lang_boxing_object::create(T_DOUBLE, &value_jval, CHECK_NULL);
+    Handle value(THREAD, value_oop);
+    objArrayOop score_pair = oopFactory::new_objectArray(2, CHECK_NULL);
```
rvansa: Not sure if it's worth to create individual arrays working as a tuple; since you return a generic object array anyway I would return this in alternating pairs.

src/java.base/share/classes/jdk/internal/crac/Score.java
```diff
@@ -57,63 +78,93 @@ public void afterRestore(Context<? extends Resource> context) {
...
+     * read to provide an up-to-date value of some metrics.
+     * <p>
+     * The provider is expected to add or update the score by invoking
+     * {@link #setScore(String, double)}, possibly multiple times.
+     * <p>
+     * The provider is recorded through a weak reference. The caller is
+     * responsible for storing a strong reference while the provider functions.
+     * <p>
+     * The order in which the providers are called is indeterminate and can
+     * change.
```
rvansa: What happens if the provider throws?

src/java.base/share/classes/jdk/internal/crac/Core.java
```diff
@@ -33,6 +33,18 @@
...
     private static ClaimedFDs claimedFDs;
     private static final JfrResource jfrResource = new JfrResource();
+    private static final Runnable internalResourceCounter = () -> {
+        int resources = 0;
+        for (var p : Priority.values()) {
+            if (p.getContext() instanceof OrderedContext<?> octx) {
+                resources += octx.size();
+            }
+        }
+        Score.setScore("jdk.crac.internalResources", resources);
```
rvansa: I wonder if we should call it `jdk.crac.internalContext.size` for symmetry with `globalContext`

src/hotspot/share/services/diagnosticCommand.cpp
```diff
@@ -1113,13 +1113,9 @@ void CheckpointDCmd::execute(DCmdSource source, TRAPS) {
       // with '.' as the decimal point (some other locales use ',')
       LocaleGuard lg;
       if (metrics[0] == '@') {
-        if (!parse_pairs_from_file("metric", metrics + 1, CheckpointDCmd::accept_metric)) {
-          return;
-        }
+        parse_pairs_from_file("metric", metrics + 1, CheckpointDCmd::accept_metric, CHECK);
```
rvansa: I like that you made the error handling more idiomatic with the rest of JVM code. I realized, though, that we are missing a test for any of this.

rvansa (commented):
I've been inconsistent in naming: score was meant to represent the final (scalar) value qualifying the image, in extension a set of all metrics. Metric was supposed to be the single key-value pair. But due to `crlib_image_score_t::set_score` I think it's OK to accept that score is an individual pair, and metric is only the name (key).
Following that I wonder if we should use plural form (scores) in some of the methods.

src/java.base/share/classes/jdk/internal/crac/Score.java
TimPushkin: I actually did not think of this. I think we should catch `Exception`s and proceed — the failing provider can catch and handle these better if it wants. Other `Throwable`s (i.e. most likely `Error`s) should be propagated.

src/java.base/share/classes/jdk/internal/crac/Core.java
TimPushkin: Not sure, they are calculated differently. `jdk.crac.globalContext.size` is literally `size()` of the "user" global context, while this does some recursion. On the other hand, these are implementation details... Ok, let's rename it.

src/hotspot/share/services/diagnosticCommand.cpp
TimPushkin: Since this is not trivial and a pre-existing problem I propose to solve it separately

TimPushkin:
Windows jobs fail to install MSVC — unrelated issue visible also in a parallel PR

src/hotspot/share/services/diagnosticCommand.cpp
rvansa: OK. Before integrating the change please declare that you've done some testing of the error-handling manually.

rvansa (approved):

src/hotspot/share/services/diagnosticCommand.cpp
TimPushkin: Works well:
```
> $JAVA_HOME/bin/jcmd 76468 JDK.checkpoint scores=
76468:
java.lang.IllegalArgumentException: Empty key in scores

> $JAVA_HOME/bin/jcmd 76468 JDK.checkpoint scores=asfaf
76468:
java.lang.IllegalArgumentException: Missing value for key 'asfaf' in scores

> $JAVA_HOME/bin/jcmd 76468 JDK.checkpoint labels=@non-existent
76468:
java.io.IOException: Cannot open non-existent: No such file or directory

> $JAVA_HOME/bin/jcmd 76468 JDK.checkpoint labels=@Waiter.java
76468:
java.lang.IllegalArgumentException: Invalid line 1 in Waiter.java (no '=' separator)
```

TimPushkin: Filed the task internally

PR #314, opened 2026-05-12
8384412: [CRaC] Customizable error message on CPU features check failure

rvansa:
The author of app distribution embedding a CRaC image can provide further hints when CPU features check does not succeed, for example recommend an alternative app distribution (think e.g. Docker image). 

---------

---------
### Progress

### Issue
 * [JDK-8384412](https://bugs.openjdk.org/browse/JDK-8384412): [CRaC] Customizable error message on CPU features check failure (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/314/head:pull/314` \
`$ git checkout pull/314`

`$ git checkout pull/314` \
`$ git pull https://git.openjdk.org/crac.git pull/314/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 314`

`$ git pr show -t 314`

</details>

<a href="https://git.openjdk.org/crac/pull/314.diff">https://git.openjdk.org/crac/pull/314.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/314#issuecomment-4429222137)
</details>

### DIFF
--- a/src/hotspot/cpu/x86/vm_version_x86.cpp
+++ b/src/hotspot/cpu/x86/vm_version_x86.cpp
@@ -1355,7 +1355,10 @@ void VM_Version::print_using_features_cr() {
     tty->print_raw_cr("CPU features are being kept intact as requested by -XX:CPUFeatures=ignore");
   } else {
     tty->print_raw("CPU features being used are: -XX:CPUFeatures=");
-    _features.print_numbers(*tty);
+    // Workaround JDK-8311164: CPU_HT is set randomly on hybrid CPUs like Alder Lake.
+    VM_Features features = _features;
+    features.clear_feature(CPU_HT);
+    features.print_numbers(*tty);
     tty->cr();
   }
 }
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -919,11 +919,11 @@ void crac::restore(crac_restore_data& restore_data) {
   }
 
   const int ret = engine.restore();
-  if (ret != 0 && (CheckCPUFeaturesMessage == nullptr || strcmp("quiet", CheckCPUFeaturesMessage))) {
+  if (ret != 0) {
     log_error(crac)("CRaC engine failed to restore from %s: error %d", CRaCRestoreFrom, ret);
     VM_Version::VM_Features current_features;
     VM_Version::cpu_features_binary(&current_features); // ignore return value
-    engine.check_cpuinfo(&current_features, exact, CheckCPUFeaturesMessage);
+    engine.check_cpuinfo(&current_features, exact);
   }
 }
 
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -919,11 +919,11 @@ void crac::restore(crac_restore_data& restore_data) {
   }
 
   const int ret = engine.restore();
-  if (ret != 0) {
+  if (ret != 0 && (CheckCPUFeaturesMessage == nullptr || strcmp("quiet", CheckCPUFeaturesMessage))) {
     log_error(crac)("CRaC engine failed to restore from %s: error %d", CRaCRestoreFrom, ret);
     VM_Version::VM_Features current_features;
     VM_Version::cpu_features_binary(&current_features); // ignore return value
-    engine.check_cpuinfo(&current_features, exact);
+    engine.check_cpuinfo(&current_features, exact, CheckCPUFeaturesMessage);
   }
 }
 
--- a/src/hotspot/share/runtime/globals.hpp
+++ b/src/hotspot/share/runtime/globals.hpp
@@ -2031,6 +2031,13 @@ const int ObjectAlignmentInBytes = 8;
           "skip: don't check features at all (requires "                    \
           "-XX:+UnlockExperimentalVMOptions).")                             \
                                                                             \
+  product(ccstr, CheckCPUFeaturesMessage, nullptr, RESTORE_SETTABLE,        \
+          "Message reported when CPU features checks fails. This string "   \
+          "can contain placeholders '%c', '%s' or '%m' for current CPU "    \
+          "features, features stored in the image or common (intersection " \
+          "of current & stored) features respectively. A special value "    \
+          "'quiet' suppresses the failure message entirely.")               \
+                                                                            \
   product(bool, UseObjectMonitorTable, true, DIAGNOSTIC,                    \
           "Use a table to record inflated monitors rather than the first "  \
           "word of the object.")                                            \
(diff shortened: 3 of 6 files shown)


rvansa:
`ResolveTest` failure is something docker-related
`InterlockTest` also fails on PID conflict, it is known to fail spuriously.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -1355,7 +1355,10 @@ void VM_Version::print_using_features_cr() {
     tty->print_raw_cr("CPU features are being kept intact as requested by -XX:CPUFeatures=ignore");
   } else {
     tty->print_raw("CPU features being used are: -XX:CPUFeatures=");
-    _features.print_numbers(*tty);
+    // Workaround JDK-8311164: CPU_HT is set randomly on hybrid CPUs like Alder Lake.
+    VM_Features features = _features;
+    features.clear_feature(CPU_HT);
```
TimPushkin: This is the third place with such workaround, I wonder if we could do this somewhere centrally

src/hotspot/share/runtime/crac.cpp
```diff
@@ -919,11 +919,11 @@ void crac::restore(crac_restore_data& restore_data) {
   }
 
   const int ret = engine.restore();
-  if (ret != 0) {
+  if (ret != 0 && (CheckCPUFeaturesMessage == nullptr || strcmp("quiet", CheckCPUFeaturesMessage))) {
     log_error(crac)("CRaC engine failed to restore from %s: error %d", CRaCRestoreFrom, ret);
```
TimPushkin: An option named "CheckCPUFeaturesMessage" should not silence the whole restore error message, only the CPU-features part of it.

I'd expect `engine.check_cpuinfo` to use `CheckCPUFeaturesMessage` internally.

test/jdk/jdk/crac/engine/CheckCPUFeaturesMessageTest.java
```diff
@@ -0,0 +1,103 @@
...
+                .doRestoreToAnalyze()
+                .shouldHaveExitValue(1)
+                .shouldNotContain("quiet")
+                .shouldNotContain("CRaC engine failed to restore");
+    }
+
+    private CracBuilder mustFail(CracBuilder builder) {
+        return builder.clearVmOptions().vmOption("-XX:CheckCPUFeatures=exact");
+    }
+    @Override
```
TimPushkin, line 78:
```suggestion
    }

    @Override
```

src/hotspot/share/runtime/crac.cpp
rvansa: I anticipate that the user want a quiet failure for transparent restore or startup, that's why I wanted to silence the message here as well. Though, probably you can achieve that with `-Xlog:crac=fatal`... okay then, I shall add a test case.

Good point about silencing criuengine and simengine as well, though that should land through a different PR.

src/hotspot/cpu/x86/vm_version_x86.cpp
rvansa: Do you mean just `static suppressNonInstructionFeatures(...)` method that would be called from all the three places, or a bigger overhaul?

TimPushkin: I did not have anything in particular in mind, just noted in case you'll have some good idea. If nothing better than what already exists comes to mind it can be left as is.

src/hotspot/share/runtime/crac.cpp
TimPushkin: Generally for silencing logs we should indeed use what the logging facility provides: introduce a more fine-grained logging, e.g. `(crac, checkpoint)`, `(crac, restore)`, maybe `(crac, restore, cpu)` for CPU features, and let the user silence what they do not need. Transparent restore/startup should then auto-silence`(crac, restore)` logs.

But my main point is that an option with a CPU-features-centric name should only affect CPU-features-related things, not the general restore-failed log.

> Good point about silencing criuengine and simengine as well, though that should land through a different PR.

TBH I did not put that into my message but yeah, we should do this at some point. I just meant that `CheckCPUFeaturesMessage` could probably be used by `engine.check_cpuinfo` function internally, not as a parameter.

src/hotspot/cpu/x86/vm_version_x86.cpp
rvansa: This place and `VM_Version::cpu_features_binary` could have it replaced by call into `aot_code_cache_features()`. That place could be generalized, but this would mean more conflicts with JDK upstream and/or inconsistency across CRaC backports. In this instance I would prefer KISS over DRY.

TimPushkin: Ok, then we'll hopefully do it when eventually merging with Leyden's CPU feature handling.

rvansa:
Actually after merging with Jan's changes the AOT code masking HT is used anyway (note `cpu_features_binary()` method above the modification).

rvansa:
I've also removed the `quiet` option; absence of reporting should be achieved using `-Xlog:crac=off`.

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -2225,6 +2225,16 @@ bool VM_Version::cpu_features_binary(VM_Version::VM_Features *data) {
   return true;
 }
 
+void VM_Version::print_using_features_cr() {
+  if (_ignore_glibc_not_using) {
+    tty->print_raw_cr("CPU features are being kept intact as requested by -XX:CPUFeatures=ignore");
+  } else {
+    tty->print_raw("CPU features being used are: -XX:CPUFeatures=");
+    _features.aot_code_cache_features().print_numbers(*tty);
```
TimPushkin, line 2233: I'd make a platform-dependent function that does just this and reuse it in the unified `VM_Version::print_using_features_cr()`

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2028,6 +2028,12 @@ const int ObjectAlignmentInBytes = 8;
           "skip: don't check features at all (requires "                    \
           "-XX:+UnlockExperimentalVMOptions).")                             \
                                                                             \
+  product(ccstr, CheckCPUFeaturesMessage, nullptr, RESTORE_SETTABLE,        \
+          "Message reported when CPU features checks fails. This string "   \
+          "can contain placeholders '%c', '%s' or '%m' for current CPU "    \
+          "features, features stored in the image or common (intersection " \
+          "of current & stored) features respectively.")               \
```
TimPushkin, line 2035:
```suggestion
          "of current & stored) features respectively.")                    \
```

TimPushkin (approved):
LGTM, the two minor things can be fixed now or at some later point

PR #316, opened 2026-05-21
8385158: [CRaC] Make it possible to specify engine options only for checkpoint or restore

TimPushkin:
Adds `CRaCCheckpointEngineOptions` and `CRaCRestoreEngineOptions` appended to `CRaCEngineOptions` only on checkpoint or restore respectively.

---------

---------
### Progress

### Issue
 * [JDK-8385158](https://bugs.openjdk.org/browse/JDK-8385158): [CRaC] Make it possible to specify engine options only for checkpoint or restore (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/316/head:pull/316` \
`$ git checkout pull/316`

`$ git checkout pull/316` \
`$ git pull https://git.openjdk.org/crac.git pull/316/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 316`

`$ git pr show -t 316`

</details>

<a href="https://git.openjdk.org/crac/pull/316.diff">https://git.openjdk.org/crac/pull/316.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/316#issuecomment-4508899615)
</details>

### DIFF
--- a/src/hotspot/share/runtime/arguments.cpp
+++ b/src/hotspot/share/runtime/arguments.cpp
@@ -1629,10 +1629,18 @@ bool Arguments::check_vm_args_consistency() {
   }
 #endif
 
-  if (CheckCPUFeatures != nullptr && !strcmp(CheckCPUFeatures, "skip") && !UnlockExperimentalVMOptions) {
-    jio_fprintf(defaultStream::error_stream(),
-                "-XX:CheckCPUFeatures=skip is allowed only with -XX:+UnlockExperimentalVMOptions\n");
-    return false;
+  if (CheckCPUFeatures != nullptr && strcmp(CheckCPUFeatures, "skip") == 0 && !UnlockExperimentalVMOptions) {
+    jio_fprintf(defaultStream::error_stream(), "-XX:CheckCPUFeatures=skip requires -XX:+UnlockExperimentalVMOptions\n");
+    status = false;
+  }
+
+  if (CRaCCheckpointEngineOptions != nullptr && CRaCCheckpointTo == nullptr) {
+    jio_fprintf(defaultStream::error_stream(), "-XX:CRaCCheckpointEngineOptions requires -XX:CRaCCheckpointTo\n");
+    status = false;
+  }
+  if (CRaCRestoreEngineOptions != nullptr && CRaCRestoreFrom == nullptr) {
+    jio_fprintf(defaultStream::error_stream(), "-XX:CRaCRestoreEngineOptions requires -XX:CRaCRestoreFrom\n");
+    status = false;
   }
 
   return status;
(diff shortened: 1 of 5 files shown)


TimPushkin:
CI failures aren't related:
- MSVC failed to install in all Windows jobs — I wonder if this is a problem unique to my fork since others do not seem to get this, but clearing GHA caches did not help
- `ContainerOOMETest` and `InterlockTest` failed on Linux — they have been failing before

src/hotspot/share/runtime/arguments.cpp
```diff
@@ -2703,7 +2711,9 @@ static bool should_record_for_restore(const JVMFlag& flag) {
            "unexpected CRaCEngine* flag: %s", flag.name());
     return false;
   }
-  if (strcmp(flag.name(), "IgnoreUnrecognizedVMOptions") == 0) {
+  if (strcmp(flag.name(), "IgnoreUnrecognizedVMOptions") == 0 ||
+      strcmp(flag.name(), "CRaCCheckpointEngineOptions") == 0 ||
```
rvansa: I think that `CRaCCheckpointEngineOptions` should be sent over, for a possible second checkpoint (if the engine supports that). Not sure if we relay `CRaCCheckpointTo` - also if you're passing that during restore without `CRaCCheckpointTo` it would fail in the check above?

rvansa (commented):
Mostly LGTM.

src/hotspot/share/runtime/arguments.cpp
TimPushkin: We relay `CRaCCheckpointTo` but not `CRaCEngineOptions`. Relaying the options means we need to re-configure the engine on restore slowing the restore down. If we delay that until the next checkpoint it would be asymmetrical with how we parse options on startup and not just before checkpoint (because we want to fail fast on mis-configuration).

TimPushkin: I should remove this particular line anyway since `CRaCCheckpointEngineOptions` is not restore-settable and thus already filtered-out

rvansa: We're not relaying `CRaCEngineOptions` because these are meant to be scoped to current (restore) operation implicitly. `CRaCCheckpointEngineOptions` wouldn't make any sense without relaying... so I think that this being non-restore-settable is OK, nothing is silently ignored.
If the user needs to reconfigure the engine for second checkpoint, let's do that through a management operation (`jcmd` / `CRaCMxBean`). Not in this PR ofc :)

TimPushkin:
The test failures are unrelated

rvansa (approved):

PR #317, opened 2026-05-22
8385310: [CRaC] CRaC should fail gracefully on a second checkpoint attempt

antonvoznia:
Some CRaC engines don't support the second checkpoint after a restore, or cannot do it right now.
The changes are focused to "improve" the failure mechanism.
This PR introduces 3 checkpointable status. 
```
    never_after_restore - engine doesn't support the second
    ready_late          - engine cannot commit the checkpoint right now, but could do it later
    ready               - engine can commit the checkpoint
```

The statuses are managed by engine API library.

Testing locally:
* criuengine
* simengine (actually cannot fail the checkpoint)
* custom engine
---------

---------
### Progress

### Issue
 * [JDK-8385310](https://bugs.openjdk.org/browse/JDK-8385310): [CRaC] CRaC should fail gracefully on a second checkpoint attempt (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/317/head:pull/317` \
`$ git checkout pull/317`

`$ git checkout pull/317` \
`$ git pull https://git.openjdk.org/crac.git pull/317/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 317`

`$ git pr show -t 317`

</details>

<a href="https://git.openjdk.org/crac/pull/317.diff">https://git.openjdk.org/crac/pull/317.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/317#issuecomment-4518549652)
</details>


antonvoznia:
Could I ask you for review @rvansa @TimPushkin ?
Thanks.

src/hotspot/share/include/crlib/crlib_checkpointable_data.h
```diff
@@ -0,0 +1,54 @@
...
+
+#define CRLIB_EXTENSION_CHECKPOINTABLE_DATA_NAME "checkpointable data"
+#define CRLIB_EXTENSION_CHECKPOINTABLE(api) \
+  CRLIB_EXTENSION(api, crlib_checkpointable_data_t, CRLIB_EXTENSION_CHECKPOINTABLE_DATA_NAME)
+
+typedef enum checkpointable_status {
+    never_after_restore,     // it's not able to commit a new checkpoint after restore
+    ready_late,              // at some point it could become checkpointable
+    ready                    // checkpointable
+} checkpointable_status_t;
```
TimPushkin: Defines should be used instead of enums, see [this discussion](https://github.com/openjdk/crac/pull/286/changes/cb8d49c86e3ea5c08e15d94a411c421a1fa36482#r2703846454)

src/hotspot/share/include/crlib/crlib_checkpointable_data.h
```diff
@@ -0,0 +1,54 @@
...
+
+// API for obtaining information about chackpointable status.
+struct crlib_checkpointable_data {
```
TimPushkin: I think "checkpointable data" is not a well fitting name. You probably took inspiration from "restore data" but that one is about passing data from restoring JVM (the one you launch with -XX:CRaCRestoreFrom) to restored JVM (the one actually restored from the checkpoint). But here there's no data involved, it's just about checkpoint status.

Maybe "checkpoint possibility", for example? The method would be `get_checkpoint_possibility`, it would return these constants:
```
CRLIB_CHECKPOINT_POSSIBLE               // Yes
CRLIB_CHECKPOINT_TEMPORARILY_IMPOSSIBLE  // Not now, may be later
CRLIB_CHECKPOINT_PERMANENTLY_IMPOSSIBLE  // No, won't ever be
```
These names are pretty long though, so I'm open to other suggestions. I do not like "checkpoint state/status" since I expect that to also cover the checkpoint-in-progress status, for example (possible for non-JVM apps to see after the checkpoint has been triggered but before the engine has paused all app's threads). We could expand the coverage to report checkpoint-in-progress and restore-in-progress statuses and assume we should not checkpoint while in them, but I'm not sure if that would be correct for all engines (although we could leave it up to the user to decide...).

src/hotspot/share/include/crlib/crlib_checkpointable_data.h
```diff
@@ -0,0 +1,54 @@
...
+  crlib_extension_t header;
+
+  checkpointable_status_t (*get_checkpointable_status)(crlib_conf_t *);
```
TimPushkin: The method deserves some documentation. For example, is there a guarantee that status cannot go from possible to impossible without checkpoint being triggered? Without this the method seems not very useful since it means that a call to checkpoint may fail even if it was said to be possible, but maybe there could be engines for which such transition makes sense...

src/hotspot/share/include/crlib/crlib_checkpointable_data.h
```diff
@@ -0,0 +1,54 @@
...
+};
+typedef const struct crlib_checkpointable_data crlib_checkpointable_data_t;
+
+#ifdef __cplusplus
+} // extern "C
+#endif
+
+#endif // CRLIB_CHECKPOINTABLE_DATA_H
```
TimPushkin:
```suggestion
#endif // CRLIB_CHECKPOINTABLE_DATA_H

```

src/hotspot/share/runtime/crac.hpp
```diff
@@ -63,6 +57,7 @@ class crac: AllStatic {
 
   static jlong restore_start_time();
   static jlong uptime_since_restore();
+  static jint checkpointable_status();
```
TimPushkin, line 60: Maybe it should be put closer to `prepare_checkpoint` and `checkpoint`? Here it sits among time-related methods — these are related but I think it is closer to checkpoint-related methods.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -86,6 +80,16 @@ jlong crac::uptime_since_restore() {
   return os::javaTimeNanos() - _restore_start_nanos;
 }
 
+jint crac::checkpointable_status() {
+  bool is_checkpointable = true;
```
TimPushkin: Unused variable

src/hotspot/share/runtime/crac.cpp
```diff
@@ -86,6 +80,16 @@ jlong crac::uptime_since_restore() {
...
+  if ((crac::_generation > 1) &&
+      (crac::_engine->prepare_checkpointable_data_api() == CracEngine::ApiStatus::OK)) {
```
TimPushkin: While for our engines this is correct, in general engines should be allowed to not be ready even before the first checkpoint, e.g. maybe an engine performs some concurrent preparations.

src/java.base/linux/native/libcriuengine/criuengine.cpp
```diff
@@ -495,6 +496,13 @@ static char* get_relative_file(const char* rel) {
   return buf;
 }
 
+static checkpointable_status_t get_checkpointable_status(crlib_conf_t * conf) {
+  if (!conf->direct_map()) {
+    return ready;
+  }
+  return never_after_restore;
```
TimPushkin: It is worth it to leave a comment on why enabling direct-mapping forbids multiple checkpoints. I believe it's just because it is not practical, even though technically possible if all previous images are preserved?

src/java.base/linux/native/libcriuengine/criuengine.cpp
```diff
@@ -974,12 +982,21 @@ static crlib_user_data_t user_data_extension = {
...
+};
+
 static const crlib_extension_t* extensions[] = {
   &restore_data_extension.header,
   &image_constraints_extension.header,
   &image_score_extension.header,
   &user_data_extension.header,
   &description_extension.header,
+  &checkpointable_data_extension.header,
   nullptr
```
TimPushkin, line 1005: Extensions should be sorted by expected usage frequency and performance importance (from most to least) because they are searched linearly. We should have placed a comment above `extensions[]` explaining this.

Description should be the last one because usually when the user queries engine's docs they plan to just read them without actually using the engine yet, i.e. the performance doesn't matter. User data is currently unused IIRC, so it should also be below this new extension.

src/java.base/share/classes/jdk/internal/crac/mirror/CheckpointableStatus.java
```diff
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2026, Azul Systems, Inc. All rights reserved.
  * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
```
TimPushkin: I believe Oracle has not been directly involved here, and we've been dropping their mention in such cases

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
```diff
@@ -273,6 +274,25 @@ private static void checkpointRestore(long jcmdStream) throws
             CheckpointException,
             RestoreException {
         final List<String> newArguments;
+        int status_code = getCheckpointableStatus0();
```
TimPushkin: The new code does not use `newArguments` so the variable should be below the new code

src/java.base/share/native/libsimengine/simengine.cpp
```diff
@@ -288,11 +289,24 @@ static crlib_description_t description_extension = {
...
+  },
+  get_checkpointable_status,
+};
+
 static const crlib_extension_t *extensions[] = {
   &restore_data_extension.header,
   &image_constraints_extension.header,
   &image_score_extension.header,
   &description_extension.header,
+  &checkpointable_data_extension.header,
```
TimPushkin: I believe the extensions are searched linearly here too, so the same comment about the order of elements as in criuengine

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
```diff
@@ -273,6 +274,25 @@ private static void checkpointRestore(long jcmdStream) throws
...
+        switch (CheckpointableStatus.fromCode(status_code)) {
+            case NEVER_AFTER_RESTORE -> {
+                CheckpointException ex = new CheckpointException();
+                ex.addSuppressed(new Exception("Current engine doesn't support second checkpoint after restore."));
+                throw ex;
+            }
+            case READY_LATER -> {
+                CheckpointException ex = new CheckpointException();
+                ex.addSuppressed(new Exception("CRaC cannot commit checkpoint right now, try later."));
```
TimPushkin: The supressed exception should be `IllegalStateException`s

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
```diff
@@ -273,6 +274,25 @@ private static void checkpointRestore(long jcmdStream) throws
...
+                throw ex;
+            }
+            case READY -> {
+                // fall through to checkpoint logic below
+            }
+            default -> {
+                System.err.printf("Engine returned unknown checkpointable status code: %d. Proceeding with checkpoint.%n", status_code);
```
TimPushkin, line 292: I'd treat this as an error and throw

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
```diff
@@ -273,6 +274,25 @@ private static void checkpointRestore(long jcmdStream) throws
...
+            }
+        }
```
TimPushkin, line 294: I thought we wanted to make this a public interface? I'm not sure if we should completely block checkpoint based on the current status since it can change until we actually reach calling the engine: resource processing below can take a long time (resources are free to block the thread).

TimPushkin (commented):
I expected this to be an optional (but maybe recommended for some engines) user's responsibility to do this checking instead of us fully blocking checkpoint attempts based on it. It's possible that between the moment we do the check and the moment we reach calling the engine to checkpoint the status changes. It seems a good fit for `CRaCMXBean`.

Also some tests should be added.

src/hotspot/share/include/crlib/crlib_checkpointable_data.h
rvansa: The C/R API is intended as generic, and shouldn't enforce the particular transition. This API Anton is adding is informative, and inherently prone to races.

I am definitely in support of clarifying docs (possibly warning about that positive answer might be already stale), but it shouldn't be restrictive.

src/hotspot/share/runtime/crac.hpp
antonvoznia: I put the `checkpointable_status()` close to others JNI related functions.
Sure, it could go in your suggested place. But I find the current one reasonable as well.

src/java.base/share/classes/jdk/internal/crac/mirror/CheckpointableStatus.java
antonvoznia: Definitely it hasn't been.
This one was inspired by some others examples, for instance:
https://github.com/openjdk/crac/blob/crac/src/java.base/linux/native/libcriuengine/user_data.cpp#L3

I don't know if it was accidentally or intentionally, and I decided to put it also.
Perhaps, we have to clean up our copy-rights.

src/hotspot/share/runtime/crac.hpp
TimPushkin: I think methods should generally be grouped by what they do, not by where they are used, since more call sites may appear with time (maybe we'll want to check the status somewhere in the JVM as well). But this is minor, so I'll accept the current placing as well.

src/java.base/share/classes/jdk/internal/crac/mirror/CheckpointableStatus.java
TimPushkin: The origins of the linked file can be traced back to `src/java.base/unix/native/criu/criu.c` which was imported with Oracle's mention back in 2021 and does not exist anymore. Maybe at some point someone from Oracle was involved in creating the original code as the CRaC OpenJDK project has non-Azul members in the census.

I'm not sure if we are legally allowed to just remove such mentions from the license headers so it's better not to add them to begin with until there's a reason to.

src/hotspot/share/runtime/crac.cpp
antonvoznia: That's an open question what we are going to implement.
Currently, I considered the checkpointable status after restore only. The naming is not clear and good.
I completely agree about your point `"in general engines should be allowed to not be ready even before the first checkpoint"`.
What I can suggest there, it's functions renaming to `checkpointable_status_after_restore()` to avoid a confusion.
WDYT?

TimPushkin: I'd prefer not to make the API so specific:
- The first checkpoint should ideally not be any special w.r.t. to the other checkpoints
- I can imagine an engine doing some background preparations on startup and thus being not ready even before the first checkpoint

src/hotspot/share/include/crlib/crlib_checkpointable_data.h
antonvoznia: done

src/hotspot/share/include/crlib/crlib_checkpointable_data.h
antonvoznia: Done, please check if you would agree with following naming
```
#define CRLIB_CHECKPOINTABLE_NEVER   0
#define CRLIB_CHECKPOINTABLE_NOT_YET 1  
#define CRLIB_CHECKPOINTABLE_READY   2 
```

src/hotspot/share/runtime/crac.hpp
antonvoznia: I left it untouched

src/hotspot/share/runtime/crac.cpp
antonvoznia: done

src/hotspot/share/runtime/crac.cpp
antonvoznia: done

src/java.base/linux/native/libcriuengine/criuengine.cpp
antonvoznia: done

src/java.base/linux/native/libcriuengine/criuengine.cpp
antonvoznia: I put it after restore_data_extension.header.
It's a minor optimization, but it seems to be reasonable if we assume that restore is called more often.

src/java.base/share/classes/jdk/internal/crac/mirror/CheckpointableStatus.java
antonvoznia: done

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
antonvoznia: done

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
antonvoznia: I just don't want to block checkpoint attempt if there is incorrectly handled checkpointable status.
It already fails without these changes if cannot commit a checkpoint. Based on that, I also consider that the behavior has not been changed.

src/java.base/share/classes/jdk/internal/crac/mirror/Core.java
antonvoznia: > I thought we wanted to make this a public interface?

I'm not sure I caught the point.
If you are talking about CRaCMXBean, I didn't find a way to use it in the internal `src/java.base/share/classes/jdk/internal/crac/mirror/Core.java`.
These changes could go into `CRaCImpl.java` but then we have a problem with JCMD call.


> 'm not sure if we should completely block checkpoint based on the current status since it can change until we actually reach calling the engine

I could move this verification into synchronized block
`synchronized (checkpointRestoreLock) {`
But anyway, I would like to call it before the `beforeCheckpoint()` is handled.

TimPushkin: > I could move this verification into synchronized block `synchronized (checkpointRestoreLock) {`

The race is between the Java code and the engine so synchronizing only on Java side would not help.

I still think we should leave calling this up to the user, while `checkpointRestore` should just attempt to checkpoint and fail if that is not possible.

> If you are talking about CRaCMXBean, I didn't find a way to use it in the internal `src/java.base/share/classes/jdk/internal/crac/mirror/Core.java`.

`CRaCMXBean` could use `Core` if we want to call this in `Core`. Otherwise it should indeed be in `CRaCImpl`.

src/java.base/linux/native/libcriuengine/criuengine.cpp
```diff
@@ -974,8 +987,17 @@ static crlib_user_data_t user_data_extension = {
...
+    CRLIB_EXTENSION_CHECKPOINT_AVAILABILITY_NAME,
+    sizeof(checkpoint_availability_extension)
+  },
+  get_checkpointable_status,
+};
+
 static const crlib_extension_t* extensions[] = {
   &restore_data_extension.header,
+  &checkpoint_availability_extension.header,
   &image_constraints_extension.header,
```
TimPushkin, line 1001: Image constraints are always used on restore (which is expected to happen more often than checkpoint) so would be better to place the new extension beneath it. Same in `simengine`.

src/hotspot/share/include/crlib/crlib_checkpoint_availability.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef CRLIB_CHECKPOINT_AVAILABILITY_H
+#define CRLIB_CHECKPOINT_AVAILABILITY_H
+
+#include "crlib.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define CRLIB_EXTENSION_CHECKPOINT_AVAILABILITY_NAME "checkpointable data"
```
TimPushkin, line 32:
```suggestion
#define CRLIB_EXTENSION_CHECKPOINT_AVAILABILITY_NAME "checkpoint availability"
```

src/hotspot/share/include/crlib/crlib_checkpoint_availability.h
```diff
@@ -0,0 +1,54 @@
...
+#define CRLIB_EXTENSION_CHECKPOINTABLE(api) \
```
TimPushkin, line 33: Should be `CRLIB_EXTENSION_CHECKPOINT_AVAILABILITY` to follow naming traditions of other extensions

src/hotspot/share/include/crlib/crlib_checkpoint_availability.h
```diff
@@ -0,0 +1,54 @@
...
+  CRLIB_EXTENSION(api, crlib_checkpoint_availability_t, CRLIB_EXTENSION_CHECKPOINT_CRLIB_CHECKPOINT_AVAILABILITY_NAME)
```
TimPushkin, line 34: A smoke test for the feature would reveal typos like this
```suggestion
  CRLIB_EXTENSION(api, crlib_checkpoint_availability_t, CRLIB_EXTENSION_CHECKPOINT_AVAILABILITY_NAME)
```

src/hotspot/share/include/crlib/crlib_checkpoint_availability.h
```diff
@@ -0,0 +1,54 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+#ifndef CRLIB_CHECKPOINT_AVAILABILITY_H
+#define CRLIB_CHECKPOINT_AVAILABILITY_H
```
TimPushkin, line 24: For me "checkpoint is available" means that there is an existing checkpoint image from which we can restore, while the extension is for checking it is currently **possible**/**permitted** to create a checkpoint image.

src/hotspot/share/include/crlib/crlib_checkpoint_availability.h
```diff
@@ -0,0 +1,54 @@
...
+#endif
+
+#define CRLIB_EXTENSION_CHECKPOINT_AVAILABILITY_NAME "checkpointable data"
+#define CRLIB_EXTENSION_CHECKPOINTABLE(api) \
+  CRLIB_EXTENSION(api, crlib_checkpoint_availability_t, CRLIB_EXTENSION_CHECKPOINT_CRLIB_CHECKPOINT_AVAILABILITY_NAME)
+
+typedef int crlib_checkpointable_status_t;
+#define CRLIB_CHECKPOINTABLE_NEVER   0   // engine will never accept another checkpoint
+#define CRLIB_CHECKPOINTABLE_NOT_YET 1   // not now; may become ready later
+#define CRLIB_CHECKPOINTABLE_READY   2   // checkpoint can proceed
```
TimPushkin, line 39: Nitpicks, checkpoint can be accepted to restore from but that is not what was meant here
```suggestion
#define CRLIB_CHECKPOINTABLE_NEVER   0   // engine will never create another checkpoint
#define CRLIB_CHECKPOINTABLE_NOT_YET 1   // not now; may become ready later
#define CRLIB_CHECKPOINTABLE_READY   2   // checkpoint can be created
```

src/hotspot/share/include/crlib/crlib_checkpoint_availability.h
```diff
@@ -0,0 +1,54 @@
...
+
+
```
TimPushkin, line 41:
```suggestion

```

src/hotspot/share/include/crlib/crlib_checkpoint_availability.h
```diff
@@ -0,0 +1,54 @@
...
+  crlib_checkpointable_status_t (*get_checkpointable_status)(crlib_conf_t *);
+};
+typedef const struct crlib_checkpoint_availability crlib_checkpoint_availability_t;
+
+#ifdef __cplusplus
+} // extern "C
+#endif
+
+#endif // CRLIB_CHECKPOINT_AVAILABILITY_H
```
TimPushkin, line 54:
```suggestion
#endif // CRLIB_CHECKPOINT_AVAILABILITY_H

```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -86,6 +80,24 @@ jlong crac::uptime_since_restore() {
   return os::javaTimeNanos() - _restore_start_nanos;
 }
 
+jint crac::checkpointable_status() {
+  if (crac::_engine->prepare_checkpoint_availability_api() == CracEngine::ApiStatus::OK) {
+    crlib_checkpointable_status_t status = crac::_engine->get_checkpointable_status();
+    if (status == CRLIB_CHECKPOINTABLE_NEVER) {
+      if (crac::_generation > 1) {
```
TimPushkin, line 87: `crac::` prefix should not be needed in this method
```suggestion
  if (_engine->prepare_checkpoint_availability_api() == CracEngine::ApiStatus::OK) {
    crlib_checkpointable_status_t status = _engine->get_checkpointable_status();
    if (status == CRLIB_CHECKPOINTABLE_NEVER) {
      if (_generation > 1) {
```

src/hotspot/share/runtime/crac.cpp
```diff
@@ -86,6 +80,24 @@ jlong crac::uptime_since_restore() {
...
+        return status;
+      } else {
+        // Engine doesn't support checkpoint after restore,
+        // but it's a first checkpoint. So return READY.
+        return CRLIB_CHECKPOINTABLE_READY;
```
TimPushkin, line 92: We should not override what the engine reports, it is very unintuitive

src/hotspot/share/runtime/crac.hpp
TimPushkin, line 60: Nitpick: since we have `crlib_checkpointable_status_t` we should return that and convert to `jint` in the JNI layer

src/java.base/linux/native/libcriuengine/criuengine.cpp
```diff
@@ -495,6 +496,18 @@ static char* get_relative_file(const char* rel) {
...
 
+static crlib_checkpointable_status_t get_checkpointable_status(crlib_conf_t * conf) {
+  if (!conf->direct_map()) {
+    return CRLIB_CHECKPOINTABLE_READY;
+  }
+
+  // With direct_map, restored memory pages are mmap'd directly from the
+  // checkpoint image file rather than copied into anonymous memory. A second
+  // checkpoint would overwrite that same image, corrupting the live mappings
+  // of the running process. CRIU has no safe way to re-checkpoint in this mode.
```
TimPushkin, line 507: Just to clarify: you can change `CRaCCheckpointTo` on restore and thus then checkpoint to another place saving the old image — some time ago I tested that this works and can be restored from, not sure if it still does though. But there seems to be no practical reason to do this, unless the second image is only the difference from the first one making the second checkpoint faster — BTW, are you sure this is not the case?

TimPushkin (commented):
The comments are mostly minor. Major things that I think need to be addressed before integration:
1. Add this to `CRaCMXBean`. I'd remove it from `Core.checkpointRestore1()` completely because the API is prone to races (status can change between the check and the actual checkpoint) so it should be up to the user whether to use it on per-engine basis.
2. Add a simple test, e.g. that we cannot checkpoint after direct-mapped restore.

src/java.base/linux/native/libcriuengine/criuengine.cpp
```diff
@@ -495,6 +496,18 @@ static char* get_relative_file(const char* rel) {
   return buf;
 }
 
+static crlib_checkpointable_status_t get_checkpointable_status(crlib_conf_t * conf) {
+  if (!conf->direct_map()) {
+    return CRLIB_CHECKPOINTABLE_READY;
+  }
```
TimPushkin, line 502: Now that I think of it, I do not think this would actually work. On restore there are two instances of the engine: (1) the restoring one and (2) the one being restored. Options are not propagated from (1) to (2) this way. `direct_map` is set on restore i.e. only in (1), while this check is run before checkpoint i.e. in (2). So I believe this would work only if you set `direct_map` on the initial start up (which is incorrect and should give you a warning on checkpoint).

I think you would need to inspect the memory of the process to see if it comes from a mmaped file to make this work.

rvansa: I wouldn't do any introspection, that's not in line with what CRIU does. The restoring engine should rather transfer data to checkpointed engine (we do that for env and system properties, though the implementation passes only the PID number and the 'data' is actually transferred at JVM level)

TimPushkin: FYI currently the restoring JVM does not record `CRaCEngineOptions` in restore-data at all, it and some other restore-settable options that affect only the restoring JVM/engine are filtered out here: https://github.com/openjdk/crac/blob/aabad2d0ffb44bcc3c951d3ecd65c1e48c500736/src/hotspot/share/runtime/arguments.cpp#L2698-L2710

TimPushkin: Making the engine parse `CRaCEngineOptions` from restore-data is also not universal since the JVM can set additional options. A separate engine-internal mechanism would need to be implemented to make it pass info to its counterpart.

antonvoznia: I looked at the code, it seems I could use `user_data` to save the value of current argument of `direct_map` on each restore() call.
Extract the value once the `get_checkpointable_status()` is called.
WDYT?

TimPushkin: I don't exactly remember how `user_data` is implemented in `criuengine`, but using `user_data` itself does not fit well semantically as it's meant to hold stuff set by the user of the engine, and this is implementation details of `checkpointable_status`. But implementing it in some way similar to `user_data` would probably work (again, I don't remember how it's implemented at the moment).

TimPushkin: Also `user_data` is for passing from checkpoint to restore IIRC, and what needs to be done here is the other way around

TimPushkin: And I am personally still not fully convinced if we need to forbid checkpoint after direct-map of CRIU at all: https://github.com/openjdk/crac/pull/317#discussion_r3301828326. Especially since it is not trivial.

antonvoznia: > Also user_data is for passing from checkpoint to restore IIRC

can we extend the purpose of the mechanism? To avoid creating similar things a code generation for no reason

TimPushkin: I believe `user_data` has been unused for a while because we moved to more task-specific APIs, initially it was for CPU features AFAIR. So it should probably be deprecated and dropped rather than extended.

Anyway, relaying `direct_map` is an implementation detail of `checkpointable_status` in `criuengine` and should not touch user-facing APIs. It can use a similar implementation under the hood, just should not be user-visible.

TimPushkin: Refreshed my knowledge on user-data. It is used to add arbitrary data to checkpoint image, in `criuengine` case, by adding files to the image. This is suitable for checkpoint->restore data passing but not the other way around because a particular restore should not mutate an image possibly used for many restores.

antonvoznia: as @TimPushkin suggested at the first message, does it make sense to look at process memory mapping in `/proc/158992/map` and find if there the same path as image_location has?
By this, we can verify 
1. if the process mapped directly
2. if the new CheckpointTo path differs from RestoreFrom.

Do you guys think the approach makes sense or not?

rvansa: As I've said before, I don't think that criuengine should do *any* introspection. If you need to pass data from restoring VM to restored one on JVM level, this would go to `CracRestoreParameters`. Since we're talking about something implemented on engine level, this means probably extending the *implementation* of `crlib_restore_data_t::set_restore_data()` to pass both JVM data and engine data.

Right now the implementation works this way:
* restoring JVM creates a file at `/known/path/crac.<restoring pid>`
* passes `<restoring pid>` to the engine as 4-byte (int) data through `crlib_restore_data_t::set_restore_data()`
* engine sets the stringified version of this PID in `CRAC_NEW_ARGS_ID` env var
* this is still visible by post-restore CRIU hook that uses this as the 4-byte payload to signal restored process
* restored engine reads the value from signal, and makes this avaliable through `crlib_restore_data_t::get_restore_data()`
* restored JVM now knows the PID of restoring process and reads `/known/path/crac.<restoring pid>`

The `/known/path/...`  is modifiable. As Timofei said, we don't want to mutate checkpoint folder on restore, so this temporary path is preferred. Eventually this could work like this
* restoring JVM passed arbitrary number of bytes (not just 4) to the engine
* criuengine writes `/known/path/<xxx>engine.<restoring pid>`, combining JVM-level data and own data
* passing the PID through `CRAC_NEW_ARGS_ID` and signals happens as before
* restore engine reads  `/known/path/<xxx>engine.<restoring pid>`, stores contents in an allocated buffer
* JVM now reads its serialized data, does not touch any file at all

This approach will require modifications to engines that don't support arbitrary length of restore data, but I believe it is architecturally correct approach.

PR #320, opened 2026-05-28
8385573: [CRaC] Fix use-after-free in engine options parsing

TimPushkin:
Fixes the bug by using a single buffer for concatenated engine options.

---------

---------
### Progress

### Issue
 * [JDK-8385573](https://bugs.openjdk.org/browse/JDK-8385573): [CRaC] Fix use-after-free in engine options parsing (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/320/head:pull/320` \
`$ git checkout pull/320`

`$ git checkout pull/320` \
`$ git pull https://git.openjdk.org/crac.git pull/320/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 320`

`$ git pr show -t 320`

</details>

<a href="https://git.openjdk.org/crac/pull/320.diff">https://git.openjdk.org/crac/pull/320.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/320#issuecomment-4564171368)
</details>


TimPushkin:
The test failures are unrelated. I reported the two `jdk/crac` ones internally, and regarding Windows it looks like my repo just got "lucky" to use newer runner images earlier than the others'.

src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -240,18 +204,64 @@ static crlib_conf_t *create_conf(const crlib_api_t &api, bool for_restore) {
...
 
+  const char *extra_options = for_restore ? CRaCRestoreEngineOptions : CRaCCheckpointEngineOptions;
+  const size_t main_options_len = CRaCEngineOptions != nullptr ? strlen(CRaCEngineOptions) : 0;
+  const size_t extra_options_len = extra_options != nullptr ? strlen(extra_options) : 0;
+  if (main_options_len == 0 && extra_options_len == 0) {
+    return conf;
+  }
+
+  const bool options_need_sep = main_options_len > 0 && extra_options_len > 0;
+  char * const options = NEW_C_HEAP_ARRAY(char, main_options_len + extra_options_len + (options_need_sep ? 1 : 0) + 1, MemTag::mtInternal);
```
rvansa, line 215: This is quite verbose code, could we use `stringStream` instead?

src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -240,18 +204,64 @@ static crlib_conf_t *create_conf(const crlib_api_t &api, bool for_restore) {
...
+    const char *key = strsep(&key_value, "=");
+    const char *value = key_value != nullptr ? key_value : "";
+    assert(key != nullptr, "Should have terminated before");
+    if (vm_controlled_keys.contains(key)) {
+      log_warning(crac)("VM-controlled CRaC engine option provided, skipping: %s", key);
+      continue;
+    }
+    {
+      bool is_new_key;
+      user_keys.put_if_absent(key, &is_new_key);
```
rvansa, line 250: In case of duplicate options we issue only warning, and first value wins, rather than last value. I think that the more specific VM options should win, but currently you first apply main, then extra.
(And yes, this is inline with existing behaviour, I didn't realize that at first)

src/hotspot/share/runtime/crac_engine.cpp
TimPushkin: `stringStream` does not allow accessing its contents as non-const. So we would need to work around its declared interface and `const_cast` to use `strsep`, or do an additional copy from the stream, which I'd prefer not to do.

rvansa: I see. I had in mind something with resource allocation (cheap) during growth and final `as_string()` copy to heap-allocated array, but apparently there isn't one.

You could avoid the `> 0` checks below, and maybe use `stpcpy` to reduce verbosity further.

TimPushkin: `> 0` is needed because `CRaC*EngineOptions` can be nulls and `memcpy` from null is UB even with 0 size according to the standard. And `stpcpy` is not available on Windows.

src/hotspot/share/runtime/crac_engine.cpp
TimPushkin: Actually, the last value wins. Here we just record the key, it does not make sense to replace the key with the same key. A few lines below we still call `configure`.

rvansa (approved):

PR #321, opened 2026-06-02
8385815: [CRaC] Fix aarch64 failing testcases

jankratochvil:
test/jdk/jdk/crac/CPUFeatures/SimpleCPUFeaturesTest.java
test/jdk/jdk/crac/engine/CheckCPUFeaturesTest.java

It needs some updates for the new CPU Features support. 

---------

---------
### Progress

### Issue
 * [JDK-8385815](https://bugs.openjdk.org/browse/JDK-8385815): [CRaC] Fix aarch64 failing testcases (**Bug** - P4)

### Reviewers
 * [Timofei Pushkin](https://openjdk.org/census#tpushkin) (@TimPushkin - Committer) ⚠️ Review applies to [7a7e742e](https://git.openjdk.org/crac/pull/321/files/7a7e742e19f99f251d02130eb5823bf9e9f0fa34)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/321/head:pull/321` \
`$ git checkout pull/321`

`$ git checkout pull/321` \
`$ git pull https://git.openjdk.org/crac.git pull/321/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 321`

`$ git pr show -t 321`

</details>

<a href="https://git.openjdk.org/crac/pull/321.diff">https://git.openjdk.org/crac/pull/321.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/321#issuecomment-4602502035)
</details>


jankratochvil:
The failed CI is a false positive:
- [jdk/crac/ContainerOOMETest](https://github.com/jankratochvil/crac/actions/runs/26819480974#user-content-jdk_crac_containeroometest)

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -898,6 +898,9 @@ VM_Features VM_Version::CPUFeatures_generic() {
   if (_cpu_features.supports_feature(CPU_NOTPACA)) {
     retval.set_feature(CPU_NOTPACA);
   }
+  if (_cpu_features.supports_feature(CPU_LSE)) {
```
rvansa: I am starting to think that we made an error accepting the code above: In fact `aarch64` should reject `generic` because we cannot create universally applicable image (which is the meaning of `generic`).

jankratochvil: I agree. And `PACA` is the same case.

jankratochvil:
With upstream CRIU it somehow does not work now:
```
+ bin/java -XX:CRaCCheckpointTo=cr -XX:+ShowCPUFeatures -XX:CRaCEngine=criuengine -XX:CPUFeatures=0x21461805ddfbf7,0xfcc0000000000000 CPUFeaturesAWS
This machine's CPU features are: -XX:CPUFeatures=0xe14e1b9dfdfbff,0xfcc0000000000000
CPU features being used are: -XX:CPUFeatures=0x21461805ddfbf7,0xfcc0000000000000
Original environment variable: GLIBC_TUNABLES=glibc.pthread.rseq=0
Re-exec of java with new environment variable: GLIBC_TUNABLES=glibc.pthread.rseq=0:glibc.cpu.hwcaps=,-ERMS,-AVX512F,-AVX512CD,-AVX512BW,-AVX512DQ,-AVX512VL
This machine's CPU features are: -XX:CPUFeatures=0xe14e1b9dfdfbff,0xfcc0000000000000
CPU features being used are: -XX:CPUFeatures=0x21461805ddfbf7,0xfcc0000000000000
CPUFeaturesCheck 0
CPUFeaturesCheck 1
+ set +e
+ bin/jcmd CPUFeaturesAWS JDK.checkpoint
CPUFeaturesCheck 2
CPUFeaturesCheck 3
CPUFeaturesCheck 4
CPUFeaturesCheck 5
CPUFeaturesCheck 6
CPUFeaturesCheck 7
Could not find any processes matching : 'CPUFeaturesAWS'
+ wait 2084
CPUFeaturesCheck 8
CPUFeaturesCheck 9
...
CPUFeaturesCheck 367
CPUFeaturesCheck 368
CPUFeaturesCheck 369
...
```

src/hotspot/cpu/x86/vm_version_x86.hpp
```diff
@@ -40,101 +40,104 @@ class VM_Feature_Flag {
...
+    decl(XSAVE,             "xsave",             122) \
+    decl(CMPXCHG16,         "cmpxchg16",         123) /* Also known in cpuinfo as cx16 and in glibc as cmpxchg16b */ \
+    decl(LAHFSAHF,          "lahfsahf",          124) /* Also known in cpuinfo as lahf_lm and in glibc as lahf64_sahf64 */ \
+    decl(HTT,               "htt",               125) /* hotspot calls it 'ht' but that is affected by threads_per_core() */ \
+    decl(XSAVEC,            "xsavec",            126) \
+    decl(AVX_Fast_Unaligned_Load, "avx_fast_unaligned_load", 127) \
     /**/
 
-#define DECLARE_CPU_FEATURE_FLAG(id, name) CPU_##id,
+#define DECLARE_CPU_FEATURE_FLAG(id, name, bit) CPU_##id = (bit),
```
TimPushkin, line 125: Would it be possible to have two versions: with and without an explicit ID to minimize the difference with the mainline? Currently CPU-features is already the most frequently conflicting part when merging with the mainline...

jankratochvil: Done. Did I understood you correctly? I would not call the change too great or did you have some better idea?

TimPushkin: I haven't thought an implementation through, so not sure if this is possible to implement at all, but what I'd ideally like to see is for the mainline features to be declared in exactly the same way as they were and only for our additional features to use another macro with explicit value:
```
#define CPU_FEATURE_FLAGS(decl, decl3) \
    decl(CX8,               cx8               )  /*  next bits are from cpuid 1 (EDX) */ \
    decl(CMOV,              cmov              )  \
...
    decl3(FMA4,             fma4,          117) \
    decl3(MOVBE,            movbe,         118) \
...
```

TimPushkin: The current version indeed seems worse than the previous: it's more complex and still changes declarations of the mainline features, i.e. still introduces conflicts

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
TimPushkin: BTW, if you now know that there's no mandatory features on arm64 this TODO can be removed, or replaced with a comment explaining that

test/jdk/jdk/crac/CPUFeatures/CPUFeaturesAWS.sh
```diff
@@ -287,8 +287,14 @@ if echo "$arch"|grep -q ', x86-64, ';then
...
-# ZULU-84672: CPUFeatures: Intel/AMD image portability problem; non-contained intersection; a!=(a&b)!=b
-checkpoint_restore "$LINENO" t3a.micro t3.micro "1:Restore failed due to incompatible or missing CPU features, try using -XX:CPUFeatures=0x8510c02eefdf7,0x3f3 on checkpoint." "" ""
+# Test 2 issues:
+# 1: ZULU-84672: CPUFeatures: Intel/AMD image portability problem; non-contained intersection; a!=(a&b)!=b
+# 2: IgnoreCPUFeatures is not inherited from snapshot to restore.
+checkpoint_restore "$LINENO" t3a.micro t3.micro "1:Restore failed due to incompatible or missing CPU features, try using -XX:CPUFeatures=0x21461805ddfbf7,0xfcc0000000000000 on checkpoint." \
+  "-XX:+UnlockExperimentalVMOptions -XX:+IgnoreCPUFeatures" ""
+
+# Test IgnoreCPUFeatures - in this case it only works by luck.
+checkpoint_restore "$LINENO" t3a.micro t3.micro "" "" "-XX:+UnlockExperimentalVMOptions -XX:+IgnoreCPUFeatures"
```
TimPushkin: `IgnoreCPUFeatures` does not exist anymore, that's why I removed all references to it, and you are bringing them back now, I assume unintentionally

src/hotspot/cpu/x86/vm_version_x86.hpp
jankratochvil: One must include the bit number for each feature as that is a compatibility requirement of `-XX:CPUFeatures=0xnum1,0xnum2`. The only possibility is to give up on the bit mask and for example introduce a list of feature names - internal comment `https://azul.atlassian.net/browse/ZULU-84390?focusedCommentId=601289`.
OK so I will revert this change.

test/jdk/jdk/crac/CPUFeatures/CPUFeaturesAWS.sh
jankratochvil: I did undo the removal intentionally - if it fails I will find it out during the testcase reverification. So when `IgnoreCPUFeatures` has been removed why not to use in the testcase the new `CheckCPUFeatures=ignore` instead? I will try that.

src/hotspot/cpu/x86/vm_version_x86.hpp
TimPushkin: As I imagined it, `decl(CX8, cx8)` would be equivalent to `decl3(CX8, cx8 , 0)` if we rely on enum's default value assignment

TimPushkin: But if we want CPUFeatures compatibility between releases, maybe we actually want the conflicts to be able to notice and replace features removed from the mainline with `*_UNUSED` with the same bit placement...

jankratochvil: > But if we want CPUFeatures compatibility between releases, maybe we actually want the conflicts to be able to notice and replace features removed from the mainline with `*_UNUSED` with the same bit placement...

Yes, that's my plan. Do we agree that this is the least bad of all the options?

TimPushkin: Agree, but I think we should have a policy for their eventual removal, so that we won't be moving further and further from the mainline.

~~E.g. while an `*_UNUSED` feature exists, a warning (or even a fatal error?) is printed when it is specified, and on the next or one after the next major release it is removed (chosen so that there's at least one release with a warning/error).~~ That won't work because non-unused features with higher values will be moved and those who have been using just them will never get the warning. Then the only alternative I see is to use a few bits of CPUFeatures as a version, incremented when a non-backwards-compatible change occurs.

jankratochvil: Currently, OpenJDK uses 66 bits (including the abandoned `*_UNUSED` ones). That still leaves room for another 51 bits, which should be sufficient for quite some time.
If additional space is eventually needed, a completely new text format can be introduced while keeping the current `0xnum,0xnum` format for backward compatibility. For `PT_NOTE`, it can contain more than 16 bytes using a new arbitrary format (such as the bits for a version).
IMO this is only a concern in the distant future.

jankratochvil:
Known false positive failures:
- [jdk/crac/SecureRandom/InterlockTest](https://github.com/jankratochvil/crac/actions/runs/28683332742#user-content-jdk_crac_securerandom_interlocktest)
- windows-x64 / build

test/jdk/jdk/crac/CPUFeatures/CPUFeaturesAWS.sh
jankratochvil: `-XX:+IgnoreCPUFeatures` has been changed in the testcase to `-XX:CheckCPUFeatures=skip`.

jankratochvil:
`CPUFeaturesAWS.sh` has been tested on x86_64 and aarch64.

src/hotspot/share/runtime/vm_features.inline.hpp
```diff
@@ -26,6 +26,8 @@
 
 #include "utilities/ostream.hpp"
 
+#include <utility>
+
```
TimPushkin, line 37: Since this is pre-existing I'm OK with fixing this later, but this file certainly should have more includes:
- From HotSpot style guide: "All .inline.hpp files should include their corresponding .hpp file as the first include line"
- Includes for everything else directly used in the file

TimPushkin: Also having public methods declared only in this file probably breaks "Declarations needed by other files should be put in the .hpp file, and not in the .inline.hpp file" from the style guide

src/hotspot/share/runtime/vm_features.inline.hpp
```diff
@@ -204,6 +206,34 @@ class VM_Features : protected VM_Feature_Flag {
     print_numbers(ss);
     return buf;
   }
+
+  struct FeaturesNames {
```
TimPushkin, line 217: From HotSpot style guide: "Declarations needed by other files should be put in the .hpp file, and not in the .inline.hpp file".

Although as I wrote above, I believe this file had been breaking this rule even before this addition.

src/hotspot/share/runtime/vm_features.inline.hpp
```diff
@@ -204,6 +206,34 @@ class VM_Features : protected VM_Feature_Flag {
...
+    const char *a[MAX_CPU_FEATURES];
+
+    constexpr const char *operator[](size_t i) const {
+      const char *retval = a[i];
```
TimPushkin:
```suggestion
    const char *names[MAX_CPU_FEATURES];

    constexpr const char *operator[](size_t i) const {
      const char *retval = names[i];
```

src/hotspot/share/runtime/vm_features.inline.hpp
TimPushkin, line 217: Why do you think we need this class now? IIUC, it is only for out-of-bounds checks, but all its usages seem to use `MAX_CPU_FEATURES` as before, so what changed? I am not against the asserts, but it moves us further from the mainline...

test/jdk/jdk/crac/engine/CheckCPUFeaturesTest.java
```diff
@@ -63,8 +63,8 @@ public class CheckCPUFeaturesTest implements CracTest {
     @Override
     public void test() throws Exception {
         boolean success = "pass".equals(result);
-        if (!Platform.isX86() && !Platform.isX64()) {
-            success = success || "fail-x86".equals(result);
+        if (!Platform.isX86() && !Platform.isX64() && !Platform.isAArch64()) {
```
TimPushkin, line 66: BTW, the `Platform.isX64()` check in `test/jdk/jdk/crac/ignoreRestore/MissingMetadataTest.java` can probably also be extended to include aarch64 (and X86?)

src/hotspot/share/runtime/vm_features.inline.hpp
jankratochvil: This is similar to `src/hotspot/share/gc/parallel/psClosure.inline.hpp` where is a line:
```
// No psClosure.hpp
```
There are no separate methods or functions in this file.

src/hotspot/share/runtime/vm_features.inline.hpp
jankratochvil: It is there because there is a gap of 51 `nullptr` entries in the middle of `_features_names` which I did not want to list literally. And to prevent running a constructor during startup one needs to generate the array as `constexpr`. And OpenJDK should not use `std::array` as it is dependent on `new` etc. I haven't found how to `constexpr`-initialize an array without wrapping it into a `struct`.

src/hotspot/share/runtime/vm_features.inline.hpp
TimPushkin: `psClosure.inline.hpp` is only included in a `.cpp` file, while this is included into `.hpp` as well.

As I see it, `inline.hpp` are closer to `.cpp` than to `.hpp`, ideally they should only implement inline stuff declared in some `.hpp`.

jankratochvil: Most of the methods in `src/hotspot/share/runtime/vm_features.inline.hpp` are a single-line or a few lines code. By the current practice all their declarations will have to be duplicated in `src/hotspot/cpu/x86/vm_version_x86.hpp` and `src/hotspot/cpu/aarch64/vm_version_aarch64.cpp` as they are not suitable for `src/hotspot/share/runtime/abstract_vm_version.hpp` as they depend on `MAX_CPU_FEATURES` and some other arch-specific items.
There are already some declaration duplications among `src/hotspot/cpu/x86/vm_version_x86.hpp` and `src/hotspot/cpu/aarch64/vm_version_aarch64.cpp` and other arch include files. I could make it worse but I rather made this new style.

TimPushkin: Fair enough. Fixing the includes should not have such consequences though, at least for those things that do not come from arch-dependent files.

jankratochvil: I do not see what to change, could you suggest? There exists no file `vm_features.hpp`. And when I provide in the beginning
```
class VM_Feature_Flag { public: enum Feature_Flag { MAX_CPU_FEATURES }; };
#define CPU_FEATURE_FLAGS(x)
```
The file `src/hotspot/share/runtime/vm_features.inline.hpp` compiles fine.

test/jdk/jdk/crac/engine/CheckCPUFeaturesTest.java
jankratochvil: Added aarch64 to `test/jdk/jdk/crac/ignoreRestore/MissingMetadataTest.java` but I tested it only on x86_64. Kept it without `isX86()` as I have some doubts it works. For the test I had to change this line, unaware how it can ever work without it:
```
-            .forwardClasspathOnRestore(true);
+            .forwardClasspathOnRestore(true)
+	    .env("PATH", "/home/azul/azul/crac-git/build/fastdebug-default/images/jdk/lib" + ":" + System.getenv().getOrDefault("PATH", ""));
```
as otherwise it was complaining:
```
[49315 ERR] criuengine: Cannot find CRIU executable: criu_location option not set, not found on PATH
```

src/hotspot/share/runtime/vm_features.inline.hpp
```diff
@@ -26,6 +26,8 @@
 
 #include "utilities/ostream.hpp"
 
+#include <utility>
```
TimPushkin, line 36:
```suggestion
// No vm_features.hpp - this depends on arch-specific VM_Feature_Flag.

#include "memory/allocation.hpp"
#include "utilities/debug.hpp"
#include "utilities/globalDefinitions.hpp"
#include "utilities/macros.hpp"
#include "utilities/ostream.hpp"

#include <cstdint>
#include <utility>
```

src/hotspot/share/runtime/vm_features.inline.hpp
TimPushkin: https://github.com/openjdk/crac/pull/321/changes#r3529342130

test/jdk/jdk/crac/engine/CheckCPUFeaturesTest.java
TimPushkin: I believe it should work without this if you have CRIU in your `PATH`

jankratochvil: Ah OK, I had it just in `jdk/lib/`.

TimPushkin: As we now focus on upstream CRIU here that is not bundled with JDK, criuengine now only looks at `criu_location` engine option and if that is not set searches for `criu` in `PATH`.

TimPushkin (approved):

TimPushkin:
Since this has not been merged yet, could you please merge the target into your branch? `jdk-28+5` landed there, so just to check there are no new problems with this change. It will also fix the Windows build failures.

jankratochvil:
Known false positive:
- [jdk/crac/SecureRandom/InterlockTest](https://github.com/jankratochvil/crac/actions/runs/29023634707#user-content-jdk_crac_securerandom_interlocktest)

jankratochvil:
windows-aarch64 is a false positive failure:
- https://github.com/jankratochvil/crac/actions/runs/29212730408/job/86703144212#step:3:266
```
error: msys: signature from "Christoph Reiter (MSYS2 development key) <reiter.christoph@gmail.com>" is unknown trust
```

TimPushkin (approved):

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -3421,8 +3423,10 @@ bool VM_Version::is_intrinsic_supported(vmIntrinsicID id) {
 
 VM_Features VM_Version::CPUFeatures_mandatory() {
   VM_Features vm_features;
-  vm_features.set_feature(CPU_SSE);
```
TimPushkin, line 3424: As I wrote downstream, I think SSE should stay here as JVM assumes it is present just like SSE2

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -3421,8 +3423,10 @@ bool VM_Version::is_intrinsic_supported(vmIntrinsicID id) {
...
   vm_features.set_feature(CPU_SSE2);
+  // It is for compatibility of the set with older JDKs.
+  vm_features.set_feature(CPU_MMX_UNUSED);
```
TimPushkin, line 3430: In JDK 26 we don't have MMX in mandatory, so why is it here?

TimPushkin: MMX has been added to mandatory in JDK 26

src/hotspot/cpu/x86/vm_version_x86.cpp
```diff
@@ -3423,10 +3423,10 @@ bool VM_Version::is_intrinsic_supported(vmIntrinsicID id) {
 
 VM_Features VM_Version::CPUFeatures_mandatory() {
   VM_Features vm_features;
-  // SSE and SSE2 are set unconditionally in feature_flags().
+  // JDK-8383881 in jdk28+ removed supports_sse(), supports_sse2(), supports_mmx() and supports_clflush().
+  // JDK-8383881 has kept the CPU_SSE and CPU_SSE2 bits for an unknown reason.
```
TimPushkin: JFYI these were removed and then brought back by https://bugs.openjdk.org/browse/JDK-8384869: "...there are numerous tests that use these strings as pre-condition for performing IR verification. Instead of fixing all these tests, it is much simpler to just record these features in the cpu feature string"

TimPushkin (approved):

PR #324, opened 2026-06-12
Merge jdk:jdk-28+0

TimPushkin:
Merges `jdk-28+0`.

Notable parts:
- JDK-8383881 and JDK-8384869 removed `CPU_FLUSH`, `CPU_MMX`, made `CPU_SSE`, `CPU_SSE2` mandatory (`FLUSH` is also now mandatory but without an enum) — I dropped CRaC code that used `CPU_FLUSH`, inlined the only `CPU_MMX` check we had, made `CPU_SSE`, `CPU_SSE2` mandatory in `-XX:CPUFeatures`, updated hardcoded CPU features values in tests
- JDK-8382030 and JDK-8385011 enabled ARM64 testing on Windows and Linux which uncovered pre-existing bugs with our CPU features code for ARM64 — this is to be fixed separately, probably by https://github.com/openjdk/crac/pull/321
- JDK-8382582 removed JVMCI (the whole feature, I was surprised), so many small conflicts with CPU features and recompilations
- Removed `IgnoreCPUFeatures` that expires in this release

Output of `git show --remerge-diff`: https://gist.github.com/TimPushkin/1af925f55f841431c87aa31099925b43 — I dropped contents of JVMCI files that were fully removed because they are huge and of no interest.

---------

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/324/head:pull/324` \
`$ git checkout pull/324`

`$ git checkout pull/324` \
`$ git pull https://git.openjdk.org/crac.git pull/324/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 324`

`$ git pr show -t 324`

</details>

<a href="https://git.openjdk.org/crac/pull/324.diff">https://git.openjdk.org/crac/pull/324.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/324#issuecomment-4693194396)
</details>


TimPushkin:
The CI failures are unrelated:
- Build failures on windows-x64 — known issue JDK-8386551, fixed in the mainline just today, didn't make it into the tag being merged
- CPU features tests on ARM64 platforms — as described above, to be fixed separately
- `jdk/crac/SecureRandom/InterlockTest` — known issue

TimPushkin:
@jankratochvil, almost all conflicts are CPU-features-related, please review.

Also I didn't run the manual `CPUFeaturesAWS` test. The hardcoded values I changed there were calculated manually, same way as for the other tests.

jankratochvil:
> @jankratochvil, almost all conflicts are CPU-features-related, please review.

Why have you introduced `VM_Version::CPUFeatures_mandatory()`? A minimal valid CPUFeatures set for x86_64 was already `VM_Version::CPUFeatures_generic()`. For aarch64 there exists no minimal set which would work for all aarch64 CPUs - see https://github.com/openjdk/crac/pull/321/changes#diff-a87e260510f34ca7d9b0feb089ad982be8268c5c8aa5a71221f6738b051ea488R890 .

> Also I didn't run the manual `CPUFeaturesAWS` test. The hardcoded values I changed there were calculated manually, same way as for the other tests.

It does not matter much, it is being updated by https://github.com/openjdk/crac/pull/321 .

TimPushkin:
@jankratochvil, `VM_Version::CPUFeatures_mandatory()` is for this new check of parsed `-XX:CPUFeatures`: https://github.com/TimPushkin/crac/blob/44b08c197b31d231747bf8fe834cdac706432291/src/hotspot/share/runtime/abstract_vm_version.inline.hpp#L302-L314

Correct me if I'm wrong, without the check if the user passed a CPUFeatures string that does not include SSE and SSE2, which are now mandatory in OpenJDK, SSE and SSE2 will be disabled in glibc. But in OpenJDK they cannot be disabled, so they will continue to be used, even though they will not be present in `VM_Version`, which is incorrect.

jankratochvil:
OK, so the new OpenJDK version makes `SSE` and `SSE2` mandatory even on x86 (previously, this was only the case on x86_64). I would rather make `Extended -XX:CPUFeatures` with missing mandatory features an error than an extension (the user explicitly requested this feature mask), but that is just my opinion; your solution is also valid.

And I see now that you have removed `CPU_MMX` and `CPU_FLUSH` from x86_64 `CPUFeatures_generic()` because upstream removed them from the enum definition. So that is OK; I had concerns about that before.

So yes, I am fine with it, thanks.

TimPushkin:
Agree, I should make this an error instead. I made it the way it is because it's now just impossible to launch a JVM on a machine that does not support these two features. But the user may try do disable them for a machine that does support them for some reason, and that should be an error.

TimPushkin:
Made missing mandatory features a error. The error messages now look like this:
```
$ java -XX:CPUFeatures=0xffff00,0x0 -version
[0.002s][error][os,cpu] This machine's CPU features are 0x8d30e02ee7ff7,0x6e6, they are missing the following features required by -XX:CPUFeatures: 0x118000,0x0 = tscinvbit, tscinv, erms
[0.002s][error][os,cpu] If you are sure it will not crash you can override this check by -XX:+UnlockExperimentalVMOptions -XX:CheckCPUFeatures=skip .
[0.002s][error][os,cpu] -XX:CPUFeatures is missing mandatory features sse, sse2, you can fix this by using -XX:CPUFeatures=0xffff60,0x0
Error occurred during initialization of VM
Invalid -XX:CPUFeatures=0xffff00,0x0, see error log for details
```

BTW, the "If you are sure it will not crash you can override this check by -XX:+UnlockExperimentalVMOptions -XX:CheckCPUFeatures=skip ." suggestion seems misleading, it does not override the CPUFeatures check that prints it. Since this is pre-existing I reported it as ZULU-92005.

TimPushkin:
Noticed `-XX:CRaCEngine=pause/pauseengine` should also be removed in JDK 28 — will do in a separate followup since it does not break anything and this PR is already big enough

TimPushkin:
@jankratochvil, please re-review.

The CI failures are the same as before.

jankratochvil:
> @jankratochvil, please re-review.

Fine with me, thanks.

TimPushkin:
Thanks. I pushed a micro-optimization, it's small so I won't ask for another review. Will integrate once CI passes.

PR #331, opened 2026-07-23
8385359: [CRaC] aarch64: c7g.medium <-> c8g.medium FAIL: vector length

jankratochvil:
It is a draft `src/hotspot/share/runtime/crac_engine.cpp` needs to be reworked. I believe `crlib_image_constraints::get_failed_bitmap` should become `get_bitmap` as this code needs to query image's CPUFeatures to find out its vector length to configure the CPU before restore. Otherwise one could also create some new user data. @rvansa?

---------

---------
### Progress

### Issue
 * [JDK-8385359](https://bugs.openjdk.org/browse/JDK-8385359): [CRaC] aarch64: c7g.medium &lt;-&gt; c8g.medium FAIL: vector length (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/331/head:pull/331` \
`$ git checkout pull/331`

`$ git checkout pull/331` \
`$ git pull https://git.openjdk.org/crac.git pull/331/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 331`

`$ git pr show -t 331`

</details>

<a href="https://git.openjdk.org/crac/pull/331.diff">https://git.openjdk.org/crac/pull/331.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/331#issuecomment-5084341839)
</details>


rvansa:
The image constraints was meant as structured data that is validated by the engine. The check is performed in the engine as to move JVM out of the loop when there are more images and score-based criteria. On the other hand, `user_data` is meant to be data opaque to the engine - so this fits better into the picture.

The only downside of adding `get_bitmap` and `get_label` methods to the API is that in practice the values will have to be stored in the configuration until it is destroyed - this is not a strong con, the data is minimal. Duplicating entry storage in user data ain't ideal either, anyway. Let's not change `get_failed_bitmap`, though - the API should evolve append-only.

@TimPushkin what's your opinion on the suggested API change?

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -930,3 +931,32 @@ void VM_Version::CPUFeatures_apply_arch(VM_Features &parsed, VM_Features &missin
   missing.clear_feature(CPU_PACA);
   missing.clear_feature(CPU_NOTPACA);
 }
+
+bool VM_Version::restore_pre(VM_Features image_features, const char *image_location) {
+  bool boolvalue = image_features.supports_feature(CPU_SVE256);
```
rvansa: nitpick: `boolvalue` -> `image_supports_sve256` or some better name

src/hotspot/os_cpu/linux_aarch64/vm_version_linux_aarch64.cpp
```diff
@@ -226,6 +234,26 @@ void VM_Version::check_os_cpu_info() {
     // GLIBC_TUNABLES=glibc.cpu.hwcaps is unsupported on aarch64
     vm_exit_during_initialization(err_msg("LSE (%s) cannot be disabled via -XX:CPUFeatures on aarch64.", lse.print_numbers()));
   }
+  if (FLAG_IS_DEFAULT(CPUFeatures)) {
+    assert(_cpu_features.supports_feature(CPU_SVE256) == _features.supports_feature(CPU_SVE256), "CPU_SVE256 is not changed");
+  } else if (!_cpu_features.supports_feature(CPU_SVE256) && _features.supports_feature(CPU_SVE256)) {
+    stringStream ss;
+    VM_Features sve256;
+    sve256.set_feature(VM_Feature_Flag::CPU_SVE256);
+    ss.print("Specified -XX:CPUFeatures=%s have CPU_SVE256=%s set but this CPU does not support it", _features.print_numbers(), sve256.print_numbers());
```
rvansa: I understand that this provides a bit better message, but shouldn't this be handled by the general CPU features mismatch handling? Could we generate a better message in the general handling instead?

jankratochvil: I agree, this code cannot be reached, removed it.

TimPushkin:
Adding `get_bitmap` and `get_label` to the constraints API LGTM: if we have setters it feels natural to have getters.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -930,3 +931,31 @@ void VM_Version::CPUFeatures_apply_arch(VM_Features &parsed, VM_Features &missin
...
+  }
+  if (image_supports_sve256 && !_cpu_features.supports_feature(CPU_SVE256)) {
+    ResourceMark rm;
+    VM_Features sve256;
+    sve256.set_feature(CPU_SVE256);
+    VM_Features use = image_features & _cpu_features;
+    log_error(crac)("Image %s has -XX:CPUFeatures=%s with CPU_SVE256=%s, this CPU has CPUFeatures=%s not supporting CPU_SVE256, use -XX:CPUFeatures=%s during snapshot",
+                    image_location, image_features.print_numbers(), sve256.print_numbers(), _cpu_features.print_numbers(), use.print_numbers());
+    return false;
+  }
```
TimPushkin: This part should be skipped when `CheckCPUFeatures=skip`

jankratochvil: Do you agree it should be if:
```
VM_Version::check_cpu_features_skip() || (CheckCPUFeatures != nullptr && !strcmp(CheckCPUFeatures, "skip"))
```
It should be moved to some method (maybe even `check_cpu_features_skip`) but I have kept it this way so far copying `src/hotspot/share/runtime/crac.cpp`:
```
  bool ignore = VM_Version::check_cpu_features_skip();
  bool exact = false;
  if (CheckCPUFeatures == nullptr || !strcmp(CheckCPUFeatures, "compatible")) {
    // default, compatible
  } else if (!strcmp(CheckCPUFeatures, "skip")) {
    ignore = true;
```

TimPushkin: Yes, this should be the check. And yes, it would be good to either move the check to a centralized place (BTW, IMO `should_skip_cpu_features_check` would be a better name for `check_cpu_features_skip` — more obvious that it is a side-effect-free boolean getter) or the commented code should be extracted to a separate method and called after `CracEngine::require_cpuinfo`.

TimPushkin:
Nitpick: consider using `pre_restore` instead of `restore_pre` — the former convention seems more common in HotSpot

src/hotspot/share/runtime/crac.cpp
```diff
@@ -328,6 +328,8 @@ int crac::checkpoint_restore(int *shmid) {
 
   // Setup CPU arch & features only during the first checkpoint; the feature set
   // cannot change after initial boot (and we don't support switching the engine).
+  // should_skip_cpu_features_check() is not valid here as -XX:CheckCPUFeatures=skip
+  // does not apply for storing of CPUFeatures.
   if (_generation == 1 && !VM_Version::check_cpu_features_skip()) {
```
TimPushkin: Naming should be improved. The difference between `check_cpu_features_skip` and `should_skip_cpu_features_check` is not obvious at all from their names.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -930,3 +931,34 @@ void VM_Version::CPUFeatures_apply_arch(VM_Features &parsed, VM_Features &missin
   missing.clear_feature(CPU_PACA);
   missing.clear_feature(CPU_NOTPACA);
 }
+
+bool VM_Version::pre_restore(VM_Features image_features, const char *image_location) {
```
TimPushkin: `image_features` could be a const reference?

src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -378,11 +378,49 @@ int CracEngine::checkpoint() const {
   return _api->checkpoint(_conf);
 }
 
+static constexpr char cpuarch_name[] = "cpu.arch";
+static constexpr char cpufeatures_name[] = "cpu.features";
+
+bool CracEngine::pre_restore_bitmap_hook(const unsigned char *value, size_t value_size) const {
+  VM_Version::VM_Features features;
+  if (value_size != sizeof(features)) {
+    log_error(crac)("Invalid CPUFeatures length for image %s - got %zu, want %zu", _image_location, value_size, sizeof(features));
```
TimPushkin:
```suggestion
    log_error(crac)("Incompatible CPUFeatures length in image %s - got %zu, want %zu", _image_location, value_size, sizeof(features));
```

src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -378,11 +378,49 @@ int CracEngine::checkpoint() const {
...
+    return true;
+  }
+  if (Abstract_VM_Version::should_skip_cpu_features_check()) {
+    return true;
+  }
+  crlib_image_constraints_t *ics = CRLIB_EXTENSION_IMAGE_CONSTRAINTS(_api);
+  if (ics == nullptr) {
+    log_error(crac)("Cannot initialize constraints extension for image %s", _image_location);
+    return false;
+  }
```
TimPushkin: 1. Would be better to use `prepare_image_constraints_api()` + `_image_constraints_api` since it could already be loaded.
2. in `crac.cpp` we warn instead of failing when constraints API is unsupported by the engine. We should do the same here for consistency. Or maybe it would be better to change the behavior of `crac.cpp` to fail unless `Abstract_VM_Version::should_skip_cpu_features_check()`.

src/hotspot/share/include/crlib/crlib_image_constraints.h
```diff
@@ -69,6 +69,13 @@ typedef const struct crlib_image_constraints {
   // Returns the size of the data, in bytes — it can be more, equal to or less than
   // 'value_size'. Returned value of 0 represents an error.
   size_t (*get_failed_bitmap)(crlib_conf_t *, const char *name, unsigned char *value_return, size_t value_size);
+
+  // Register callbacks before restore. Callbacks are called only during restore.
+  // If the callback returns false the restore is aborted.
+  // name is not copied, its content must remain valid.
+  // user_data is an arbitrary pointer value which is passed along.
+  bool (*register_label_hook)(crlib_conf_t *, const char *name, bool (*hook)(const char *value, void *user_data), void *user_data);
+  bool (*register_bitmap_hook)(crlib_conf_t *, const char *name, bool (*hook)(const unsigned char *value, size_t value_size, void *user_data), void *user_data);
```
TimPushkin, line 78: The old engine API also did not do copies and eventually it became inconvenient to use — I'm afraid we may get the same here in the future in we'll have other use cases.

But then it would probably be logical to copy `user_data` as well. In the current use case it could be just `_image_location` which is copyable.

src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -378,11 +378,49 @@ int CracEngine::checkpoint() const {
   return _api->checkpoint(_conf);
 }
 
+static constexpr char cpuarch_name[] = "cpu.arch";
+static constexpr char cpufeatures_name[] = "cpu.features";
+
+bool CracEngine::pre_restore_bitmap_hook(const unsigned char *value, size_t value_size) const {
```
TimPushkin: `_image_location` is the only thing used from `CracEngine`, and it's only used in logging which I believe could be cut — we don't include it in any other logs. Then this could be static and `CracEngine::pre_restore_bitmap_hook_trampoline` wouldn't be needed.

src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -378,11 +378,49 @@ int CracEngine::checkpoint() const {
...
+  void *user_data = const_cast<CracEngine *>(this);
+  return ics->register_bitmap_hook(_conf, cpufeatures_name, pre_restore_bitmap_hook_trampoline, user_data);
+}
+
 int CracEngine::restore() const {
   precond(is_initialized());
   if (!check_engine(_name, _image_location)) {
     return -1;
   }
+  if (!pre_restore()) {
```
TimPushkin: I think the naming could be improved here, something like:
- `CracEngine::pre_restore` -> `CracEngine::register_constraints_hooks`
- `VM_Version::pre_restore` -> `VM_Version::process_image_cpu_features` — similar for `VM_Version::pre_restore_needed`

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
jankratochvil: I was thinking about that. It is 16 bytes now, for that ChatGPT says it is cheaper by value. Even for 24 bytes it is unclear whether value still is not cheaper.

src/hotspot/share/runtime/crac.cpp
jankratochvil: That's true, renamed:
- `check_cpu_features_skip()` -> `get_cpu_features_ignore()`
- `_ignore_glibc_not_using` -> `_cpu_features_ignore`

src/hotspot/share/include/crlib/crlib_image_constraints.h
jankratochvil: This and the next review comment seem to contradict each other. Here, storing the `name` string is suggested to make the API stable enough for possible future extensions.
I think the `name` string could always be copied later if needed, since that would be a backward-compatible API extension.

src/hotspot/share/runtime/crac_engine.cpp
jankratochvil: This comment suggests removing `user_data` to keep the API minimal for the current use case. My intention, however, was to make the API stable enough for future extensions.
If changes to the existing API are acceptable, the proposed `register_label_hook` is a superset of `require_label` and could therefore replace it. The same applies to `bitmap`.

src/hotspot/share/runtime/crac_engine.cpp
jankratochvil: 1. implemented
2. I just warn. The suggested change is simple and I think it should be rather done by @rvansa if really wanted.

src/hotspot/share/runtime/crac_engine.cpp
jankratochvil: The generic name was chosen with possible future extensions in mind. That said, since this does not define a stable ABI, renaming it later as part of a future extension would be straightforward.

src/hotspot/cpu/aarch64/vm_version_aarch64.hpp
```diff
@@ -122,8 +122,8 @@ class VM_Version : public Abstract_VM_Version, public VM_Feature_Flag {
   typedef ::VM_Features VM_Features;
   static void initialize();
   static bool cpu_features_binary(VM_Features *data);
-  static bool check_cpu_features_skip() {
-    return _ignore_glibc_not_using;
+  static bool get_cpu_features_ignore() {
+    return _cpu_features_ignore;
   }
```
TimPushkin, line 128: Style guide discourages using `get_`:
```
Getter accessor names are noun phrases, with no "`get_`" noise
word. Boolean getters can also begin with "`is_`" or "`has_`".  Member
function for reading data members usually have the same name as the
data member, exclusive of the leading underscore.
```

How about:
- `can_use_cpu_features` — false when `CPUFeatures=ignore` or the platform does not support CPU features — the opposite of current `get_cpu_features_ignore`
- `should_check_cpu_features` — false when `!can_use_cpu_features` or `CheckCPUFeatures=ignore` — the opposite of current `should_skip_cpu_features_check`

I used opposites as it makes the names cleaner.

Also maybe this getter should be closer to the field it returns?

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
TimPushkin: May worth a comment then

src/hotspot/share/include/crlib/crlib_image_constraints.h
TimPushkin: In the next comment I suggest that our current code should not need `void *user_data`, not that we should remove it from the API. So not a contradiction.

> the name string could always be copied later if needed, since that would be a backward-compatible API extension.

Agree.

src/hotspot/share/runtime/crac_engine.cpp
TimPushkin: As I wrote in the previous comment, I don't suggest removing `user_data` from the API, just not to use it in the way we do now (which means that for now it'll be unused by us, but that's OK).

> If changes to the existing API are acceptable, the proposed register_label_hook is a superset of require_label and could therefore replace it. The same applies to bitmap.

That would require adding more restrictions to how the new API can be used. The engine may use constraints API to select one of many images => the hooks can be called multiple times, including for images that won't end up being the one selected for restore => the hooks should not treat incompatible values as errors (e.g. in the current impl error logs should be replaced with debug logs). This is worth documenting in the API so that people don't call stuff like `exit` there.

src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -378,11 +378,54 @@ int CracEngine::checkpoint() const {
...
+      break;
+    case CracEngine::ApiStatus::ERR:
+      return true;
+    case CracEngine::ApiStatus::UNSUPPORTED:
+      log_warning(crac)("Cannot verify CPUFeatures for arch-specific part of restore "
+        "with the selected CRaC engine");
+      return true;
+  }
+  void *user_data = this;
+  return _image_constraints_api->register_bitmap_hook(_conf, cpufeatures_name, pre_restore_bitmap_hook_trampoline, user_data);
```
TimPushkin: If we remove the dependency on `this` (https://github.com/openjdk/crac/pull/331#discussion_r3666446451) it would be simpler to move this to the part of `crac.cpp` that calls `require_cpuinfo`.
- All CPUFeatures stuff would be together
- `CRaCEngine` would become sipler — it was supposed to be a thin wrapper around the engine library, mapping engine API to its methods directly

src/hotspot/share/runtime/crac.cpp
```diff
@@ -873,12 +875,12 @@ void crac::restore(crac_restore_data& restore_data) {
 
   // Since the check itself is delegated to the C/R Engine we will simply
   // skip the check here.
-  bool ignore = VM_Version::check_cpu_features_skip();
+  bool ignore = Abstract_VM_Version::should_skip_cpu_features_check();
```
TimPushkin: Looks like this would be better off using `get_cpu_features_ignore` directly. `CheckCPUFeatures` is interpreted below anyway.

src/hotspot/share/runtime/crac_engine.cpp
jankratochvil: I do not agree removing standard error reporting to save 8 normal lines of code but OK, implemented.

src/hotspot/share/runtime/crac.cpp
jankratochvil: I disagree, as it replaces code verification via an assert with duplicated implementation but implemented it.

TimPushkin: That's the thing: the implementation was duplicated anyway. I don't push this, just seems better to me.

src/hotspot/share/runtime/crac_engine.cpp
TimPushkin: I'm not sure with which part of my previous comment you don't agree. None of them were about saving code really.

If with not using `user_data`, i.e. removing image path from the error messages: I suggest this because we do not do this in any other messages so it's weird to include it in only these particular ones. If you think we need to include it then I'd propose to do that in some more high-level message, e.g. the one reporting the overall restore failure (I think we should have one, not sure).

If with replacing error messages with debug messages: I suggest this because with multi-selection they are not really errors. I agree that this is not great / is incorrect for the non-multi-selection case, but I do not see a way to fix this, the JVM does not know what case this is.

jankratochvil: I expected `_image_location` to be used in another error message, but I see it isn't. In that case, it's fine that the new error messages don't print `_image_location` either. My mistake.

If `crlib_image_constraints_t` is called for multiple images, including those that are not ultimately selected, then I placed the CPU reconfiguration in the wrong place.

For example:
* We are running on a vec256 CPU.
* The selector checks a vec128 image.
  * `set_maximum_sve_vector_length(16)` is called.
* The selector then checks and selects a vec256 image.
  * The algorithm does nothing because both the image and the native CPU support vec256.
  * However, the CPU has already been reconfigured to vec128.

I'll fix this, but it means the CPU may be reconfigured multiple times while evaluating multiple candidate images.

Is there an existing JDK test case or documentation covering multiple images? If so, I should add a corresponding test to `test/jdk/jdk/crac/CPUFeatures/CPUFeaturesAWS.sh`.

src/hotspot/cpu/aarch64/vm_version_aarch64.hpp
jankratochvil: done

src/hotspot/share/runtime/crac_engine.cpp
jankratochvil: Done. Originally, I placed it as late as possible so that a failed restore would not leave the JVM with a misconfigured CPU. However, there is never a point late enough—for example, in the case of `CRaCIgnoreRestoreIfUnavailable`. So I had to implement a proper `restore_cpuinfo()`.

src/hotspot/share/runtime/crac_engine.cpp
TimPushkin: I believe we have multi-selection (called auto-selection) only downstream, and even there I'm not sure we have JDK-level tests for it (haven't checked) since it is an engine-level feature

TimPushkin: Maybe it makes sense downstreaming this before integrating the PR in case some other changes will surface

src/hotspot/share/runtime/crac.cpp
```diff
@@ -941,7 +954,9 @@ void crac::restore(crac_restore_data& restore_data) {
   }
 
   const int ret = engine.restore();
-  if (ret != 0) {
+  if (ret == 0) {
+    ShouldNotReachHere();
+  } else {
```
TimPushkin, line 959: Engine is a separate user-provided piece of software. We should not hard-crash if it misbehaves (where possible).

Actually, according to the engine API spec any return code of `restore` is treated as an error code, even 0, so the previous behavior was also incorrect. We should not check the return value at all, merely log it.

src/hotspot/share/runtime/crac_engine.cpp
```diff
@@ -543,6 +507,19 @@ const crlib_conf_option_t *CracEngine::configuration_options() {
   return _options;
 }
 
+bool CracEngine::bitmap_constraint_hook(const unsigned char *value, size_t value_size, void *user_data/*unused*/) {
```
TimPushkin, line 510: Since it does not use anything from `CracEngine`, why not make it `.cpp`-level static?

src/hotspot/share/runtime/crac_engine.cpp
jankratochvil: OK, though I still have to fix JDK-8385359 first.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -930,3 +931,56 @@ void VM_Version::CPUFeatures_apply_arch(VM_Features &parsed, VM_Features &missin
   missing.clear_feature(CPU_PACA);
   missing.clear_feature(CPU_NOTPACA);
 }
+
+bool VM_Version::process_image_cpu_features(const VM_Features *image_featuresp) {
+  int want;
```
TimPushkin, line 936: `want_sve_vector_length`

src/hotspot/os_cpu/linux_aarch64/vm_version_linux_aarch64.cpp
```diff
@@ -163,6 +172,10 @@ void VM_Version::get_os_cpu_info() {
   update_feature(auxv2, CPU_WFXT,       HWCAP2_WFXT      );
   update_feature(~auxv, CPU_NOTPACA,    HWCAP_PACA       );
 
+  if (supports_sve() && get_current_sve_vector_length() == 32) {
+    set_feature(CPU_SVE256);
+  }
```
TimPushkin, line 177: The rest of the code seems to assume that `!supports_feature(CPU_SVE256)` => supports SVE128 (or less?), but I read there are CPUs with SVE vector lengths higher than 256 bits. Not sure JDK can run on them though. Is it guaranteed somewhere that higher lengths are not possible in this code?

TimPushkin: AI also complains about smaller sizes, resending verbatim:
- "get_os_cpu_info sets the bit only when VL is exactly 32, so a VL=64 machine (A64FX etc.) is indistinguishable from VL=16. On restore, want = image_supports_sve256 ? 32 : 16 then hard-sets VL to 16 — including when checkpointing and restoring on the same VL=64 host, where the restored JVM's code was compiled for MaxVectorSize=64. That's a silent miscompile/SIGILL, not a clean failure."
- "process_image_cpu_features(nullptr) recomputes 16/32 from _features rather than restoring what get_current_sve_vector_length() was before the hook ran. With -XX:MaxVectorSize=16 on an SVE256 CPU plus a failed restore and -XX:+CRaCIgnoreRestoreIfUnavailable (where the VM keeps running), the process ends up with VL=32 while its compiled code assumed 16."

src/java.base/share/native/libcrcommon/image_constraints.cpp
```diff
@@ -257,6 +257,35 @@ bool ImageConstraints::validate(const char* image_location) const {
...
+      }
+    });
+    if (!found) {
+      LOG("Hook did not find its tag '%s'", hook.name);
+      hooks_result = false;
+    }
+  });
+  if (!hooks_result) {
+    return false;
+  }
```
TimPushkin, line 288: Maybe it would be a good idea to run all checks instead of failing fast so that the user sees the hole picture?

It will duplicate the error message for the SVE feature, but at least this way `CheckCPUFeaturesMessage` would be employed which can contain some useful info. BTW, if we make hooks' messages debug (as would be more correct for the multiple-selection case) the duplication could be avoided, otherwise `CheckCPUFeaturesMessage` should be incorporated into the hooks' messages somehow...

src/java.base/share/native/libcrcommon/image_constraints.cpp
```diff
@@ -257,6 +257,35 @@ bool ImageConstraints::validate(const char* image_location) const {
       return false;
     }
   }
+  bool hooks_result = true;
+  _hooks.foreach([&](const Hook& hook) {
+    bool found = false;
+    tags.foreach([&](const Tag& t) {
+      if (!hooks_result || strcmp(hook.name, t.name) != 0) {
```
TimPushkin, line 264: With this, if there are multiple hooks and one fails, every next hook will print the `Hook did not find its tag` line below

src/hotspot/os_cpu/linux_aarch64/vm_version_linux_aarch64.cpp
```diff
@@ -226,6 +239,20 @@ void VM_Version::check_os_cpu_info() {
...
+    assert(_cpu_features.supports_feature(CPU_SVE256) == _features.supports_feature(CPU_SVE256), "CPU_SVE256 is not changed");
+  } else if (_cpu_features.supports_feature(CPU_SVE256) && !_features.supports_feature(CPU_SVE256)) {
+    maximum_sve_vector_length = 16;
+    int got = set_and_get_current_sve_vector_length(maximum_sve_vector_length);
+    if (got != maximum_sve_vector_length) {
+      stringStream ss;
+      VM_Features sve256;
+      sve256.set_feature(VM_Feature_Flag::CPU_SVE256);
+      ss.print("Specified -XX:CPUFeatures=%s have unset CPU_SVE256=%s, this CPU has CPUFeatures=%s but it cannot be disabled as PR_SVE_SET_VL reports %d: %m",
+               _features.print_numbers(), sve256.print_numbers(), _cpu_features.print_numbers(), got);
```
TimPushkin, line 252: `print_numbers` does resource allocation, so `ResourceMark` could be used.

There are other usages of it without the mark, but they are pre-existing.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
```diff
@@ -930,3 +931,56 @@ void VM_Version::CPUFeatures_apply_arch(VM_Features &parsed, VM_Features &missin
...
+    // Cleanup after a failed restore.
+    want = supports_feature(CPU_SVE256) ? 32 : 16;
+  } else {
+    const VM_Features &image_features = *image_featuresp;
+    bool image_supports_sve256 = image_features.supports_feature(CPU_SVE256);
+    if (image_supports_sve256 && !_cpu_features.supports_feature(CPU_SVE256)) {
+      if (!Abstract_VM_Version::should_check_cpu_features()) {
+        return true;
+      }
+      ResourceMark rm;
```
TimPushkin, line 947: `#include "memory/resourceArea.hpp"` should be added

src/java.base/share/native/libcrcommon/crcommon.cpp
```diff
@@ -63,6 +63,14 @@ static size_t get_failed_bitmap(crlib_conf_t* conf, const char* name, unsigned c
   return COMMON(conf)->image_constraints.get_failed_bitmap(name, value_return, value_size);
 }
 
+static bool register_label_hook(crlib_conf_t* conf, const char *name, ImageConstraints::LabelHook hook, void *user_data) {
+  return COMMON(conf)->image_constraints.register_label_hook(name, hook, user_data);
+};
+
+static bool register_bitmap_hook(crlib_conf_t* conf, const char *name, ImageConstraints::BitmapHook hook, void *user_data) {
+  return COMMON(conf)->image_constraints.register_bitmap_hook(name, hook, user_data);
+};
```
TimPushkin, line 72:
```suggestion
}

static bool register_bitmap_hook(crlib_conf_t* conf, const char *name, ImageConstraints::BitmapHook hook, void *user_data) {
  return COMMON(conf)->image_constraints.register_bitmap_hook(name, hook, user_data);
}
```

src/hotspot/cpu/ppc/vm_version_ppc.hpp
```diff
@@ -57,7 +57,7 @@ class VM_Version: public Abstract_VM_Version {
   static void check_virtualizations();
   struct VM_Features: public Zero_Features {};
   static bool cpu_features_binary(VM_Features *data) { return false; }
-  static bool check_cpu_features_skip() { return true; }
+  static bool can_use_cpu_features() { return false; }
```
TimPushkin, line 60: The two new methods and `restore_failed_check` should be added to ppc/riscv/s390

test/jdk/jdk/crac/CPUFeatures/CPUFeaturesAWS.sh
```diff
@@ -57,6 +57,13 @@ internal_checkpoint() {
 }
 internal_restore() {
   javasetup
+  tid=$(echo cr/core-*.img|tr -cd ' 0-9'|sed 's/^.* //') # highest
+  if [ -n "$tid" ];then
+    ls -l /proc/$tid/exe || :
+    bash -c 'echo $$'
+    (set +x;while [ $(bash -c 'echo $$') -le $tid ];do :;done)
+    bash -c 'echo $$'
```
TimPushkin, line 65: Nitpick: since the script uses dash in its shebang, maybe use dash here as well?

test/jdk/jdk/crac/CPUFeatures/CPUFeaturesAWS.sh
```diff
@@ -57,6 +57,13 @@ internal_checkpoint() {
 }
 internal_restore() {
   javasetup
+  tid=$(echo cr/core-*.img|tr -cd ' 0-9'|sed 's/^.* //') # highest
```
TimPushkin, line 60: Breaks when crossing a power of ten because of the lexicographic sort

src/hotspot/share/runtime/crac_engine.cpp
jankratochvil: Because there is
```
friend class CracEngine;
```
and so a non-class function:
```
src/hotspot/share/runtime/crac_engine.cpp:518:48: error: ‘static bool VM_Version::process_image_cpu_features(const VM_Features*)’ is private within this context
```

jankratochvil:
> Nitpick: consider using `pre_restore` instead of `restore_pre` — the former convention seems more common in HotSpot

done

src/hotspot/share/runtime/crac_engine.cpp
jankratochvil: From this thread it remains to somehow optimize the auto-selection.

src/hotspot/os_cpu/linux_aarch64/vm_version_linux_aarch64.cpp
jankratochvil: Fixed by: +VM_Version::checkpoint_check() to verify >256 vector bits CPUs

src/java.base/share/native/libcrcommon/image_constraints.cpp
jankratochvil: Implemented by: Run all _hooks .

src/hotspot/os_cpu/linux_aarch64/vm_version_linux_aarch64.cpp
jankratochvil: Implemented by: Add missing ResourceMark.

src/hotspot/cpu/aarch64/vm_version_aarch64.cpp
jankratochvil: Implemented by: Add #include "memory/resourceArea.hpp"

src/java.base/share/native/libcrcommon/crcommon.cpp
jankratochvil: Implemented by: Remove excessive ;

test/jdk/jdk/crac/CPUFeatures/CPUFeaturesAWS.sh
jankratochvil: implemeted by: CPUFeaturesAWS.sh: call dash, not bash

test/jdk/jdk/crac/CPUFeatures/CPUFeaturesAWS.sh
jankratochvil: implemented by: Fix TID sorting

src/hotspot/cpu/ppc/vm_version_ppc.hpp
jankratochvil: Implemented by: Add the new methods to ppc/riscv/s390 .

jankratochvil:
@rvansa Could you take a look at it, as @TimPushkin is unavailable this week?

src/hotspot/share/runtime/crac.cpp
```diff
@@ -894,6 +898,14 @@ void crac::prepare_restore(crac_restore_data& restore_data) {
   restore_data.restore_nanos = os::javaTimeNanos();
 }
 
+template <typename F>
```
rvansa, line 901: In `crcommon.hpp` we have this pattern named `Deferred` - I wonder if we should use the same name? I wonder why OpenJDK does not have a reuse-able RAII pattern already.

src/hotspot/share/include/crlib/crlib_image_constraints.h
rvansa: It seems to me that if `require_label/bitmap` copies the name it would be pretty inconsistent to not copy it here. On the other hand it is not possible to copy an arbitrary user pointer. One way to work this around would be to not specify the name at all, and let engine pass it as a hook argument (for any tag stored).

However, there is a bigger problem with this API design: the whole point of image constraints API extension is to validate the image & current environment compatibility **externally**. The image lookup can happen out of the process, e.g. on a remote server storing multiple images - with this hook API we would need to retrieve the tag for each image, send it over to this process, invoke the hook (looking for a first match, or doing that will all images, in case we are validating multiple tags?) and then forward the choice back to the server, and then get the right image... The API does not say if the hook is mutating, or idempotent. This is totally invalidating the point of the API.

I think that we need to break `VM_Version::process_image_cpu_features` into two parts: one that does the compatibility check (and that would be 'descriptor match'), and the other part tries to invoke `prctl` and fails the restore when the result is unexpected.

src/java.base/share/native/libcrcommon/image_constraints.cpp
```diff
@@ -283,17 +283,25 @@ bool ImageConstraints::validate(const char* image_location) const {
     } else if (c.type == TagType::LABEL && strcmp(static_cast<const char*>(c.data), static_cast<const char*>(t->data))) {
       LOG("Label mismatch for tag %s: '%s' vs. '%s'", c.name,
         static_cast<const char*>(c.data), static_cast<const char*>(t->data));
-    } else if (c.type == TagType::BITMAP && !c.compare_bitmaps(static_cast<const unsigned char*>(t->data), t->data_size)) {
-      LOG("Bitmap mismatch for tag %s:", c.name);
-      print_bitmap("Constraint:   ", static_cast<const unsigned char*>(c.data), c.data_size);
-      print_bitmap("Image:        ", static_cast<const unsigned char*>(t->data), t->data_size);
+    } else if (c.type == TagType::BITMAP && t->data_size > c.data_size) {
+      LOG("Bitmap for tag %s in the image is too large - %zu > %zu bytes",
```
rvansa: If the size in image and constraints don't match, the shorter one should be logically extended with zeroes and then the comparison should happen normally.

src/java.base/share/native/libcrcommon/image_constraints.hpp
```diff
@@ -110,12 +110,31 @@ class ImageConstraints {
 
   LinkedList<Tag> _tags;
   LinkedList<Constraint> _constraints;
+  bool (*_callback)(crlib_conf_t *conf, void *user_data) = nullptr;
```
rvansa: nitpick: would be better to call it `prepare_restore_callback` or `restore_callback` to make it more obvious and inline with the C/R API.

rvansa (commented):
Mostly LGTM.

In the future it would be nice to move the static stubs into a base class, so we don't need to repeat then for every platform, but let's not slow down the integration here.

jankratochvil:
> It seems to me that if require_label/bitmap copies the name it would be pretty inconsistent to not copy it here. On the other hand it is not possible to copy an arbitrary user pointer. One way to work this around would be to not specify the name at all, and let engine pass it as a hook argument (for any tag stored).

I do not get all the parts but regarding the copying we were discussing it here with @TimPushkin and I find it would complicate the current code and the caller currently does not need it at all. If it is ever needed in the future one can start copying it in a fully backward compatible way.

Which reminds me also that I introduced the new API call `get_bitmap` to preserve existing `get_failed_bitmap` but IMO it would be perfectly backward compatible to just rename existing `get_failed_bitmap` to `get_bitmap`.

TBH I also do not understand why to currently be so strict with backward compatibility as it is all shipped in one bundle (even downstream). One can start considering backward compatibility if there is some 3rd party API implementation.

src/java.base/share/native/libcrcommon/image_constraints.cpp
```diff
@@ -283,17 +283,23 @@ bool ImageConstraints::validate(const char* image_location) const {
...
-      print_bitmap("Image:        ", static_cast<const unsigned char*>(t->data), t->data_size);
+    } else if (c.type == TagType::BITMAP) {
       free((void *) c.image_data);
       c.image_data = static_cast<unsigned char *>(malloc(c.data_size));
       if (c.image_data == nullptr) {
         LOG("Cannot allocate memory for a bitmap copy");
         result = false;
       } else {
-        memcpy(c.image_data, t->data, c.data_size);
+        size_t common_size = t->data_size < c.data_size ? t->data_size : c.data_size;
```
rvansa: If `t->data_size > c->data_size`, you allocate too short of a buffer above, and don't copy the rest in here.

src/java.base/share/native/libcrcommon/image_constraints.hpp
```diff
@@ -110,12 +110,31 @@ class ImageConstraints {
...
   bool check_tag(const char* type, const char* name, size_t value_size);
 
+  size_t get_any_bitmap(const char* name, unsigned char *value_return, size_t value_size, bool need_failed) const {
+    size_t result = 0;
+    _constraints.foreach([&](Constraint &c) {
+      if (c.type == TagType::BITMAP && !strcmp(c.name, name) && (!need_failed || c.failed)) {
+        if (c.image_data == nullptr) {
+          result = 0;
+        } else {
+          result = c.data_size;
```
rvansa: If the bitmap length in tag is > in constraint you will return incorrect value.

rvansa:
> Which reminds me also that I introduced the new API call `get_bitmap` to preserve existing `get_failed_bitmap` but IMO it would be perfectly backward compatible to just rename existing `get_failed_bitmap` to `get_bitmap`.

Yes, it would  be backwards-compatible in binary, and I would think that returning even non-failed bitmaps is an acceptable change in behaviour - calling it on non-failed bitmap is/was undefined behaviour. But it doesn't hurt to have two methods - I wouldn't mind even setting using single implementation in the API.

> TBH I also do not understand why to currently be so strict with backward compatibility as it is all shipped in one bundle (even downstream). One can start considering backward compatibility if there is some 3rd party API implementation.

We want to be welcome to 3rd party API implementations and keep a good history with the API stewardship. And if the API changes, the implementation should be able to detect it (struct sizes...) - I don't want to rely on packaging the correct bits together for code maintained as too separate projects. Sometimes I copy downstream engine into upstream bits and test something with that. TLDR: hygiene.

rvansa (approved):

PR #343, opened 2026-08-17
8390453: [CRaC] Custom image constraints

rvansa:
We can let the user define custom image constraints based on `jdk.crac.labels` and `jdk.crac.require-labels` properties, enforcing image choice on set strings or environment variables. This can help asserting e.g. the version of application in container image matches the CRaC image.

Along, we should record CPU features already in `crac::prepare_images()`, in case this is used in C/R API extension. This requires moving the method after `VM_Version` initialization. 

A small NPE fix is piggybacked.

---------

---------
### Progress

### Issue
 * [JDK-8390453](https://bugs.openjdk.org/browse/JDK-8390453): [CRaC] Custom image constraints (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/343/head:pull/343` \
`$ git checkout pull/343`

`$ git checkout pull/343` \
`$ git pull https://git.openjdk.org/crac.git pull/343/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 343`

`$ git pr show -t 343`

</details>

<a href="https://git.openjdk.org/crac/pull/343.diff">https://git.openjdk.org/crac/pull/343.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/343#issuecomment-5319483299)
</details>

### DIFF
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -943,7 +943,7 @@ void crac::restore(crac_restore_data& restore_data) {
     case CracEngine::ApiStatus::ERR:
       return;
     case CracEngine::ApiStatus::UNSUPPORTED:
-      log_warning(crac)("Cannot verify image constraints (e.g. CPUFeatures) for restore with the selected CRaC engine");
+      log_warning(crac)("Cannot verify image constraints (CPU features, labels) for restore with the selected CRaC engine");
       break;
   }
 
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -326,27 +326,6 @@ int crac::checkpoint_restore(int *shmid) {
     return JVM_CHECKPOINT_ERROR;
   }
 
-  // Setup CPU arch & features only during the first checkpoint; the feature set
-  // cannot change after initial boot (and we don't support switching the engine).
-  if (_generation == 1 && !VM_Version::check_cpu_features_skip()) {
-    VM_Version::VM_Features current_features;
-    if (VM_Version::cpu_features_binary(&current_features)) {
-      switch (_engine->prepare_image_constraints_api()) {
-        case CracEngine::ApiStatus::OK:
-          if (!_engine->store_cpuinfo(&current_features)) {
-            return JVM_CHECKPOINT_ERROR;
-          }
-          break;
-        case CracEngine::ApiStatus::ERR:
-          return JVM_CHECKPOINT_ERROR;
-        case CracEngine::ApiStatus::UNSUPPORTED:
-          log_warning(crac)("Cannot store CPUFeatures for checkpoint "
-            "with the selected CRaC engine");
-          break;
-      }
-    }
-  }
-
   const int ret = _engine->checkpoint();
   if (ret != 0) {
     log_error(crac)("CRaC engine failed to checkpoint to %s: error %i", image_location, ret);
@@ -605,7 +584,39 @@ public:
     _t = nullptr;
     return tmp;
   }
+  T *get() {
+    return _t;
+  }
 };
+
+static bool apply_labels(const char* property, CracEngine* engine, bool (CracEngine::*func)(const char*, const char*)) {
+  const char* labels = Arguments::get_property(property);
+  if (labels == nullptr) {
+    return true;
+  }
+  char* dup = os::strdup_check_oom(labels);
+  char *ptr = dup;
+  char *key_value;
+  bool retval = true;
+  while ((key_value = strtok_r(ptr, ",", &ptr)) != nullptr) {
+    char *eq = strchr(key_value, '=');
+    const char *value = nullptr;
+    if (eq == nullptr) {
+      value = getenv(key_value);
+    } else {
+      eq[0] = '\0';
+      if (eq[1] == '$') {
+        value = getenv(eq + 2);
+      } else {
+        value = eq + 1;
+      }
+    }
+    // if value is nullptr (e.g. env unset) func should ignore the call
+    retval = (engine->*func)(key_value, value) && retval;
+  }
+  os::free(dup);
+  return retval;
+}
 
 bool crac::prepare_checkpoint() {
   precond(CRaCCheckpointTo != nullptr);
@@ -624,6 +635,35 @@ bool crac::prepare_checkpoint() {
   }
   if (fixed_path && (!ensure_checkpoint_dir(image_location, true) || !engine->configure_image_location(image_location))) {
     return false;
+  }
+
+  switch (engine->prepare_image_constraints_api()) {
+    case CracEngine::ApiStatus::OK: {
+      VM_Version::VM_Features current_features;
+      if (!VM_Version::check_cpu_features_skip() && VM_Version::cpu_features_binary(&current_features) &&
+          !engine->store_cpuinfo(&current_features)) {
+        return false;
+      }
+      char java_version_buf[64];
+      guarantee((size_t) os::snprintf(java_version_buf, sizeof(java_version_buf), "%d.%d.%d",
+        JDK_Version::current().major_version(), JDK_Version::current().minor_version(),
+        JDK_Version::current().security_version()) < sizeof(java_version_buf), "version must fit");
+      // TODO: more built-in identifiers?
+      if (!engine->set_label("version", Arguments::get_property("jdk.crac.app.version")) ||
+          !engine->set_label("java.version", java_version_buf)) {
+        log_error(crac)("Cannot set common image labels");
+        return false;
+      }
+      if (!apply_labels("jdk.crac.labels", engine.get(), &CracEngine::set_label)) {
+        log_error(crac)("Cannot set some labels from the 'jdk.crac.labels' property");
+        return false;
+      }
+    } break;
+    case CracEngine::ApiStatus::ERR:
+      return false;
+    case CracEngine::ApiStatus::UNSUPPORTED:
+      log_warning(crac)("Cannot store image tags with the selected CRaC engine");
+      break;
   }
 
   _engine = engine.extract();
@@ -892,12 +932,15 @@ void crac::restore(crac_restore_data& restore_data) {
         if (VM_Version::cpu_features_binary(&current_features)) {
           engine.require_cpuinfo(&current_features, exact);
         }
-        } break;
+        if (!apply_labels("jdk.crac.require-labels", &engine, &CracEngine::require_label)) {
+          log_error(crac)("Cannot enforce some labels from the 'jdk.crac.require-labels' property");
+          return;
+        }
+      } break;
       case CracEngine::ApiStatus::ERR:
         return;
       case CracEngine::ApiStatus::UNSUPPORTED:
-        log_warning(crac)("Cannot verify CPUFeatures for restore "
-          "with the selected CRaC engine");
+        log_warning(crac)("Cannot verify image constraints (e.g. CPUFeatures) for restore with the selected CRaC engine");
         break;
     }
   }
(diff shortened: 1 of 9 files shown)


src/hotspot/share/runtime/crac.cpp
```diff
@@ -626,6 +637,35 @@ bool crac::prepare_checkpoint() {
...
+    case CracEngine::ApiStatus::OK: {
+      VM_Version::VM_Features current_features;
+      if (!VM_Version::check_cpu_features_skip() && VM_Version::cpu_features_binary(&current_features) &&
+          !engine->store_cpuinfo(&current_features)) {
+        return false;
+      }
+      char java_version_buf[64];
+      guarantee((size_t) os::snprintf(java_version_buf, sizeof(java_version_buf), "%d.%d.%d",
+        JDK_Version::current().major_version(), JDK_Version::current().minor_version(),
+        JDK_Version::current().security_version()) < sizeof(java_version_buf), "version must fit");
```
TimPushkin: There's `JDK_Version::to_string(char *buffer, size_t buflen)`. It also includes patch and build versions which I think is a good thing.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -605,8 +584,40 @@ template<class T> class FutureRef {
...
+    } else {
+      eq[0] = '\0';
+      if (eq[1] == '$') {
+        value = getenv(eq + 2);
+      } else {
+        value = eq + 1;
+      }
+    }
+    // if value is nullptr (e.g. env unset) func should ignore the call
+    retval = (engine->*func)(key_value, value) && retval;
```
TimPushkin: Maybe early-stop on failure? Or leave a comment why not.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -892,12 +932,15 @@ void crac::restore(crac_restore_data& restore_data) {
         if (VM_Version::cpu_features_binary(&current_features)) {
           engine.require_cpuinfo(&current_features, exact);
         }
-        } break;
+        if (!apply_labels("jdk.crac.require-labels", &engine, &CracEngine::require_label)) {
```
TimPushkin: I believe `jdk.crac.requireLabels` is a more common naming convention in JDK

src/hotspot/share/runtime/crac.cpp
```diff
@@ -626,6 +637,35 @@ bool crac::prepare_checkpoint() {
...
+      guarantee((size_t) os::snprintf(java_version_buf, sizeof(java_version_buf), "%d.%d.%d",
+        JDK_Version::current().major_version(), JDK_Version::current().minor_version(),
+        JDK_Version::current().security_version()) < sizeof(java_version_buf), "version must fit");
+      // TODO: more built-in identifiers?
+      if (!engine->set_label("version", Arguments::get_property("jdk.crac.app.version")) ||
+          !engine->set_label("java.version", java_version_buf)) {
+        log_error(crac)("Cannot set common image labels");
+        return false;
+      }
+      if (!apply_labels("jdk.crac.labels", engine.get(), &CracEngine::set_label)) {
```
TimPushkin: I'm not sure whether it is a good idea to make these Java properties. They have no relation to CRaC's Java API, and so far we used VM options for things used in the JVM. On the other hand, in the future we may add some interface for setting labels from Java, so maybe it's ok...

BTW, for properties I think we should start using `{@systemProperty}` javadoc tag to have them documented somehow. But it needs to be discussed where to put that doc, especially for such properties used by the JVM. Probably at the `jdk.crac` module level.

src/hotspot/share/runtime/threads.cpp
```diff
@@ -623,6 +623,10 @@ jint Threads::create_vm(JavaVMInitArgs* args, bool* canTryAgain) {
   }
   Arguments::free_restore_only_data(); // Not needed anymore
 
+  if (CRaCCheckpointTo && !crac::prepare_checkpoint()) {
```
TimPushkin, line 628: Please leave a comment explaining why it should be here exactly (need `JDKVersion`, no point in preparing for checkpoint if we are going to restore...)

test/jdk/jdk/crac/constraints/RequiredLabelsTest.java
```diff
@@ -0,0 +1,94 @@
+package constraints;/*
```
TimPushkin: We usually don't put package in tests, but if you do it should go after the license header

test/jdk/jdk/crac/constraints/RequiredLabelsTest.java
```diff
@@ -0,0 +1,94 @@
...
+ * Copyright (c) 2025, 2026, Azul Systems, Inc. All rights reserved.
```
TimPushkin:
```suggestion
 * Copyright (c) 2026, Azul Systems, Inc. All rights reserved.
```

test/jdk/jdk/crac/constraints/RequiredLabelsTest.java
```diff
@@ -0,0 +1,94 @@
...
+ * @library /test/lib
+ * @build RequiredLabelsTest
+ * @run driver/timeout=15 jdk.test.lib.crac.CracTest foo=bar  foo=bar  true
+ * @run driver/timeout=15 jdk.test.lib.crac.CracTest foo=bar  -        true
+ * @run driver/timeout=15 jdk.test.lib.crac.CracTest -        foo=bar  false
+ * @run driver/timeout=15 jdk.test.lib.crac.CracTest foo=bar  foo=goo  false
+ * @run driver/timeout=15 jdk.test.lib.crac.CracTest foo=bar  foo=$BAR false
+ * @run driver/timeout=15 jdk.test.lib.crac.CracTest foo=$BAR foo=GOO  true
+ * @run driver/timeout=15 jdk.test.lib.crac.CracTest foo      foo=xxx  true
+ * @run driver/timeout=15 jdk.test.lib.crac.CracTest        - foo      false
```
TimPushkin:
```suggestion
 * @run driver/timeout=15 jdk.test.lib.crac.CracTest -        foo      false
```

test/jdk/jdk/crac/constraints/RequiredLabelsTest.java
```diff
@@ -0,0 +1,94 @@
...
+    String setLabels;
+
+    @CracTestArg(1)
+    String requireLabels;
+
+    @CracTestArg(2)
+    boolean shouldSucceed;
+
+    private static CracBuilder newPauseBuilder() {
+        return new CracBuilder().engine(CracEngine.SIMULATE).engineOptions("pause=true")
```
TimPushkin, line 59: The test should require linux for `pause=true`

test/jdk/jdk/crac/constraints/RequiredLabelsTest.java
rvansa: Good spot, something my IDE has inserted and I didn't notice...

src/hotspot/share/runtime/crac.cpp
rvansa: You're right that we had the discussion about VM options used in native code vs. system properties used in Java. I am just always reluctant to add another VM option. We could also make this a `CRaCEngineOption=constraints.extra.labels=foo,bar=$BAR,goo=goo`, and `constraints.extra.require-labels`. So the very question whether this should be here in JVM is up to discussion.

If the labels are used by an C/R API extension early on (not just for checkpoint), there can't be any Java API. For checkpoint, yes, we could add that, but unlike metrics I imagine that labels are more 'static' properties.

src/hotspot/share/runtime/crac.cpp
TimPushkin: This is skipped when CPUFeatures check is skipped

src/hotspot/share/runtime/crac.cpp
```diff
@@ -605,8 +584,40 @@ template<class T> class FutureRef {
...
+    if (eq == nullptr) {
+      value = getenv(key_value);
+    } else {
+      eq[0] = '\0';
+      if (eq[1] == '$') {
+        value = getenv(eq + 2);
+      } else {
+        value = eq + 1;
+      }
+    }
```
TimPushkin, line 614: If an env var is not set `set_label`/`require_label` will silently skip it. We should at least warn, or even error for `require_label`.

src/hotspot/share/runtime/crac.cpp
TimPushkin: The `CRaCEngineOption` way would be great if we didn't have `set_label`/`require_label` API duplicating it. Deprecating those also would not be great because there will be asymmetry with `set_bitmap`/`require_bitmap` that have to stay there.

rvansa: Yes, engine exposes this programmatically, but I don't see a strong reason why it wouldn't offer configuration as an alternative.
We certainly don't want to remove those - we want JVM to record some labels automatically. But does it have to play a facade role for 'extra' tags?

TimPushkin: Currently the engine does not differentiate between `CRaCEngineOptions` and what JVM itself sets. If we add `constraints.extra.labels`/`constraints.extra.require-labels` it would be the first case of such differentiation — that's what I don't like.

Making JVM append its labels to whatever user passed as `constraints.labels` in `CRaCEngineOptions` would be inline with how we handle things now, but that would mean `set_label` is duplicated.

rvansa: My proposal was to take JVM entirely out of the loop; it wouldn't be setting these options. It would be the user who cannot do `-Djdk.crac.labels=...` nor `-XX:CRaCImageLabels=...` but would just set `-XX:CRaCEngineOptions=constraints.extra.labels=...` and JVM would pass it transparently.

TimPushkin: JVM would still use `set_label` e.g. for CPU architecture, so there will be two ways to set labels through the engine API: `constraints.extra.labels` and `set_label`. From the engine's perspective there's no difference between these, just duplication.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -871,35 +913,38 @@ void crac::restore(crac_restore_data& restore_data) {
...
-        log_warning(crac)("Cannot verify CPUFeatures for restore "
-          "with the selected CRaC engine");
-        break;
-    }
+      }
+    } break;
+    case CracEngine::ApiStatus::ERR:
+      return;
+    case CracEngine::ApiStatus::UNSUPPORTED:
+      log_warning(crac)("Cannot verify image constraints (e.g. CPUFeatures) for restore with the selected CRaC engine");
```
```suggestion
      log_warning(crac)("Cannot verify image constraints (CPU features, labels) for restore with the selected CRaC engine");
```

src/hotspot/share/runtime/globals.hpp
```diff
@@ -2064,6 +2064,14 @@ const int ObjectAlignmentInBytes = 8;
   product(bool, CRaCPauseOnCheckpointError, false, DIAGNOSTIC,              \
           "Pauses the checkpoint when a problem is found on VM level.")     \
                                                                             \
+  product(ccstr, CRaCImageLabels, nullptr, "Comma-separated list of labels "\
+          "to be recorded: label1=value,label2=$ENV,ENV (the last form "    \
+          "sets label ENV from environment variable ENV)")                  \
+                                                                            \
+  product(ccstr, CRaCRequiredImageLabels, nullptr, RESTORE_SETTABLE,         \
```
TimPushkin:
```suggestion
  product(ccstr, CRaCRequiredImageLabels, nullptr, RESTORE_SETTABLE,        \
```

TimPushkin (commented):
LGTM, just two minor things

TimPushkin (approved):

PR #345, opened 2026-08-19
8390670: [CRaC] Fix restart of streaming JFR recordings

rvansa:
When a recording does not have its destination set we hit a NPE.

There's more problems beyond the NPE, though: the EventDirectoryStream used by RecordingStream can keep a file open.

Also, currently all recordings are attempted to be restarted (recreated). We should do that only with the running ones; not started and closed recordings can be left as-is. 

---------

---------
### Progress

### Issue
 * [JDK-8390670](https://bugs.openjdk.org/browse/JDK-8390670): [CRaC] Fix restart of streaming JFR recordings (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/345/head:pull/345` \
`$ git checkout pull/345`

`$ git checkout pull/345` \
`$ git pull https://git.openjdk.org/crac.git pull/345/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 345`

`$ git pr show -t 345`

</details>

<a href="https://git.openjdk.org/crac/pull/345.diff">https://git.openjdk.org/crac/pull/345.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/345#issuecomment-5343401519)
</details>

### DIFF
--- a/src/jdk.jfr/share/classes/jdk/jfr/Recording.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/Recording.java
@@ -50,7 +50,7 @@ import jdk.jfr.internal.WriteablePath;
  *
  * @since 9
  *
- * @crac The recording is automatically stopped on checkpoint & re-created on restore.
+ * @crac The recording is automatically stopped on checkpoint and re-created on restore.
  */
 public final class Recording implements Closeable {
 
--- a/src/jdk.jfr/share/classes/jdk/jfr/Recording.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/Recording.java
@@ -49,6 +49,8 @@ import jdk.jfr.internal.WriteablePath;
  * {@snippet class="Snippets" region="RecordingOverview"}
  *
  * @since 9
+ *
+ * @crac The recording is automatically stopped on checkpoint & re-created on restore.
  */
 public final class Recording implements Closeable {
 
@@ -82,7 +84,7 @@ public final class Recording implements Closeable {
         }
     }
 
-    private final PlatformRecording internal;
+    private PlatformRecording internal;
 
     /**
      * Creates a recording with settings from a map of name-value pairs.
@@ -676,6 +678,10 @@ public final class Recording implements Closeable {
     // package private
     PlatformRecording getInternal() {
         return internal;
+    }
+
+    void setInternal(PlatformRecording pr) {
+        internal = pr;
     }
 
     private void setSetting(String id, String value) {
--- a/src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordingStream.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordingStream.java
@@ -60,7 +60,7 @@ import jdk.jfr.internal.management.StreamBarrier;
  *
  * @since 14
  *
- * @crac The recording is automatically stopped on checkpoint & re-created on restore.
+ * @crac The recording is automatically stopped on checkpoint and re-created on restore.
  */
 public final class RecordingStream implements AutoCloseable, EventStream {
 
--- a/src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordingStream.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordingStream.java
@@ -59,6 +59,8 @@ import jdk.jfr.internal.management.StreamBarrier;
  * {@snippet class="Snippets" region="RecordingStreamOverview"}
  *
  * @since 14
+ *
+ * @crac The recording is automatically stopped on checkpoint & re-created on restore.
  */
 public final class RecordingStream implements AutoCloseable, EventStream {
 
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
@@ -83,13 +83,19 @@ public final class PlatformRecorder {
         @Override
         public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
             synchronized (PlatformRecorder.this) {
-                ArrayList<PlatformRecording> copy = new ArrayList<>(recordings);
-                futureRecordings = copy.stream().map(r -> {
+                // Started, delayed and stopped + closed recordings can be left as-is
+                List<PlatformRecording> running = recordings.stream()
+                        .filter(r -> r.getState() == RecordingState.RUNNING).toList();
+                futureRecordings = running.stream().map(r -> {
                     // PlatformRecording has to have a matching Recording - otherwise we could not control those
                     // through jcmd
                     Recording rec = new Recording(r.getSettings());
-                    PlatformRecording pr = PrivateAccess.getInstance().getPlatformRecording(rec);
-                    if (r.getName().equals(String.valueOf(r.getId()))) {
+                    PrivateAccess access = PrivateAccess.getInstance();
+                    PlatformRecording pr = access.getPlatformRecording(rec);
+                    // Update the owning Recording (important if this was created programmatically)
+                    access.setPlatformRecording(r.getRecording(), pr);
+                    if (r.getName().equ
+                    als(String.valueOf(r.getId()))) {
                         // default name == id, use the new id as name as well
                         rec.setName(String.valueOf(rec.getId()));
                     } else {
@@ -113,8 +119,11 @@ public final class PlatformRecorder {
                     return pr;
                 }).collect(Collectors.toList());
                 recordings.removeAll(futureRecordings);
-                copy.forEach(r -> r.stop("Checkpoint"));
-                assert recordings.isEmpty();
+                running.forEach(r -> {
+                    r.stop("Checkpoint");
+                    r.close();
+                });
+                assert recordings.stream().allMatch(r -> r.getState() != RecordingState.RUNNING);
             }
         }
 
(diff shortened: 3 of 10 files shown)


src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordingStream.java
```diff
@@ -59,6 +59,8 @@
  * {@snippet class="Snippets" region="RecordingStreamOverview"}
  *
  * @since 14
+ *
+ * @crac The recording is automatically stopped on checkpoint and re-created on restore.
```
TimPushkin, line 63: There's no official recommendation but looks like we've been putting this as the very first tag.

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -83,12 +83,17 @@ public final class PlatformRecorder {
...
+                        .filter(r -> r.getState() == RecordingState.RUNNING).toList();
+                futureRecordings = running.stream().map(r -> {
                     // PlatformRecording has to have a matching Recording - otherwise we could not control those
                     // through jcmd
                     Recording rec = new Recording(r.getSettings());
-                    PlatformRecording pr = PrivateAccess.getInstance().getPlatformRecording(rec);
+                    PrivateAccess access = PrivateAccess.getInstance();
+                    PlatformRecording pr = access.getPlatformRecording(rec);
+                    // Update the owning Recording (important if this was created programmatically)
+                    access.setPlatformRecording(r.getRecording(), pr);
```
TimPushkin, line 96: Here we create a new `Recording` (let's call it R1) with the same settings as some old one (let's call it R0), it creates a new `PlatformRecording`, then we make the old R0 point to the new `PlatformRecording`. But we never make the new `PlatformRecording` point to R0, it remains pointing to R1. So, from the second checkpoint onwards we don't have a link to R0, the actual `Recording` that the user code holds, and it stops getting updated.

Also it's hard to understand the difference between `r`, `pr`, `rec`. I think even something like `oldRecording`, `oldPlatformRecording`, `newRecording`, `newPlatformRecording` (maybe shortened, but keeping the old/new platform/non-platform pairs visible) would make it more readable.

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -124,45 +132,52 @@ public void afterRestore(Context<? extends Resource> context) throws Exception {
...
                     }
                     r.start();
                 });
             }
         }
+
+        private void backupOriginalDestination(WriteablePath original) {
+            try {
+                File destFile = original.getReal().toFile();
+                if (destFile.exists()) {
```
TimPushkin, line 156: Maybe early return to reduce the nesting?

src/jdk.jfr/share/classes/jdk/jfr/consumer/RecordingStream.java
TimPushkin: Same comment for `Recording`

test/jdk/jdk/crac/jfr/FlightRecorderStreamingTest.java
```diff
@@ -0,0 +1,62 @@
...
+import jdk.test.lib.crac.CracTest;
+
+import java.time.Duration;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * @test FlightRecorderStreamingTest
+ * @library /test/lib
+ * @requires (os.family == "linux")
```
TimPushkin, line 35: Why do we need Linux for this, wouldn't simengine suffice?

TimPushkin: Same question for the other test

src/jdk.jfr/share/classes/jdk/jfr/internal/consumer/EventDirectoryStream.java
TimPushkin, line 229: This thing opens directories and files in a loop outside the lock. AI found this, worth checking the rest of the code outside the lock.

src/jdk.jfr/share/classes/jdk/jfr/Recording.java
```diff
@@ -82,7 +84,7 @@ public Map<String, String> toMap() {
         }
     }
 
-    private final PlatformRecording internal;
+    private PlatformRecording internal;
```
TimPushkin, line 87: Shouldn't it become `volatile` in case read concurrently? I'm not sure if that is possible, just asking to verify.

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -124,45 +132,52 @@ public void afterRestore(Context<? extends Resource> context) throws Exception {
                 futureRecordings.forEach(r -> {
```
TimPushkin, line 132: Pre-existing issue: `futureRecordings` should be cleared afterwards. Otherwise, if a future beforeCheckpoint fails, afterRestore receives a stale list. And there's no point in keeping the references after restore anyway.

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
```diff
@@ -124,45 +132,52 @@ public void afterRestore(Context<? extends Resource> context) throws Exception {
...
                 });
             }
         }
+
+        private void backupOriginalDestination(WriteablePath original) {
+            try {
+                File destFile = original.getReal().toFile();
+                if (destFile.exists()) {
+                    Path backup = null;
+                    for (int i = 0; backup == null && i < MAX_BACKUPS; ++i) {
```
TimPushkin, line 158: Maybe we should warn when `MAX_BACKUPS` gets exceeded?

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
TimPushkin, line 79: Should be `final`

TimPushkin, line 81: Should be `final`. Or maybe we could update it on restore?

src/jdk.jfr/share/classes/jdk/jfr/internal/consumer/EventDirectoryStream.java
```diff
@@ -212,6 +232,7 @@ protected void processRecursionSafe() throws IOException {
                     return;
                 }
                 currentChunkStartNanos = repositoryFiles.getTimestamp(path);
+                lock.lock();
                 input.setFile(path);
```
TimPushkin, line 236: IIUC the `PlatformRecorder`'s resource may change the path and delete the old file (by closing the recording) while this thread waits on the lock, then this `setFile` will fail since the file on the old path does not exist anymore

src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecorder.java
TimPushkin, line 117: `reports` should also be copied

src/jdk.jfr/share/classes/jdk/jfr/internal/consumer/RecordingInput.java
```diff
@@ -432,12 +432,14 @@ public void require(int minimumBytes, String errorMessage) throws IOException {
...
-        } catch (IOException e) {
-            // perhaps deleted
+        if (file != null) {
+            try {
+                file.close();
+            } catch (IOException e) {
+                // perhaps deleted
+            }
+            file = null;
         }
```
TimPushkin, line 442: Why is the null check needed? The only place where `file` is set to null is in this method which then sets in back to non-null unconditionally. The only way it remains null is if this methods throws, but the only caller (the `EventDirectoryStream` loop) propagates the exception out of the scope where this object exists, so another `setFile` call is not possible.

The null check here is good as a defense, but since it does not seem CRaC-related for us it's a needless difference from the mainline.

src/jdk.jfr/share/classes/jdk/jfr/internal/consumer/EventDirectoryStream.java
```diff
@@ -283,4 +304,15 @@ public StreamBarrier activateStreamBarrier() {
         barrier.activate();
         return barrier;
     }
+
+    @Override
+    public void beforeCheckpoint(Context<? extends Resource> context) throws Exception {
+        // Ensures that the thread in processRecursionSafe() does not have an open file
+        lock.lock();
```
TimPushkin, line 311: I wonder if we should put some timeout here. AI tells me this wait can last forever at least in these cases (the reproducers are also AI-generated, I didn't test them):
1. The stream has an `onEvent` handler that got blocked — probably a user's fault for which it's OK to stall.
```java
var queue = new ArrayBlockingQueue<RecordedEvent>(10);   // nobody draining it
try (var rs = new RecordingStream()) {
    rs.enable("jdk.CPULoad").withPeriod(Duration.ofMillis(1));
    rs.onEvent("jdk.CPULoad", e -> {
        try { queue.put(e); } catch (InterruptedException ignored) {}   // blocks once full
    });
    rs.startAsync();
    Thread.sleep(1000);                                  // let the queue fill
    CRaCMXBean.getCRaCMXBean().checkpointRestore();      // never returns
}
```
2. Streams with `recording == null` where the writer never finishes, e.g. created via `EventStream.openRepository` to a repository of another JVM that died — this seems like something we should not stall on.
```
$ java -XX:StartFlightRecording -XX:FlightRecorderOptions:repository=/tmp/repo Foo &
$ sleep 2 && kill -9 %1        # chunk left with fileState != 0, never finalized

// in the JVM being checkpointed
var es = EventStream.openRepository(Path.of("/tmp/repo/<subdir>"));
es.onEvent(e -> {});
es.startAsync();
CRaCMXBean.getCRaCMXBean().checkpointRestore();   // never returns
```

PR #349, opened 2026-09-08
8391957: [CRaC] Record restore-related metrics

rvansa:
We are reporting metrics that describe JVM+application state & performance, the warmed-up-ness. The concept of metrics can be useful for engine extensions and management at different moments, though - therefore we can report restore times (native & Java) and current generation as metrics, and configure image constraints sooner than when we are attempting a restore. 

---------

---------
### Progress

### Issue
 * [JDK-8391957](https://bugs.openjdk.org/browse/JDK-8391957): [CRaC] Record restore-related metrics (**Task** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/crac.git pull/349/head:pull/349` \
`$ git checkout pull/349`

`$ git checkout pull/349` \
`$ git pull https://git.openjdk.org/crac.git pull/349/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 349`

`$ git pr show -t 349`

</details>

<a href="https://git.openjdk.org/crac/pull/349.diff">https://git.openjdk.org/crac/pull/349.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/crac/pull/349#issuecomment-5584091707)
</details>

### DIFF
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -63,7 +63,7 @@
 
 static jlong _restore_start_time;
 static jlong _restore_start_nanos;
-static jlong _restore_native_end_nanos = -1;
+static jlong _restore_native_end_nanos;
 
 CracEngine *crac::_engine = nullptr;
 unsigned int crac::_generation = 1;
--- a/src/hotspot/share/runtime/crac.cpp
+++ b/src/hotspot/share/runtime/crac.cpp
@@ -625,6 +625,34 @@ static bool apply_labels(const char* labels, CracEngine* engine, bool (CracEngin
   return true;
 }
 
+static bool apply_constraints(CracEngine &engine, bool *exact) {
+  *exact = false;
+  // Since the check itself is delegated to the C/R Engine we will simply
+  // skip the check here.
+  bool ignore = VM_Version::check_cpu_features_skip();
+  if (CheckCPUFeatures == nullptr || !strcmp(CheckCPUFeatures, "compatible")) {
+    // default, compatible
+  } else if (!strcmp(CheckCPUFeatures, "skip")) {
+    ignore = true;
+  } else if (!strcmp(CheckCPUFeatures, "exact")) {
+    *exact = true;
+  } else {
+    log_error(crac)("Invalid value for -XX:CheckCPUFeatures=%s; available are 'compatible', 'exact' or 'skip'", CheckCPUFeatures);
+    return false;
+  }
+  if (!ignore) {
+    VM_Version::VM_Features current_features;
+    if (VM_Version::cpu_features_binary(&current_features)) {
+      engine.require_cpuinfo(&current_features, *exact);
+    }
+  }
+  if (!apply_labels(CRaCRequiredImageLabels, &engine, &CracEngine::require_label)) {
+    log_error(crac)("Cannot enforce some labels from CRaCRequiredImageLabels");
+    return false;
+  }
+  return true;
+}
+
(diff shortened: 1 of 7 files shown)


src/hotspot/share/services/management.cpp
```diff
@@ -1004,11 +1004,11 @@ static jlong get_long_attribute(jmmLongAttribute att) {
 
   case JMM_JVM_UPTIME_SINCE_RESTORE_MS:
     {
-      jlong ticks = crac::uptime_since_restore();
-      if (ticks == -1) {
+      jlong nanos = crac::uptime_since_restore();
+      if (nanos == -1) {
         return -1;
       }
-      return Management::ticks_to_ms(ticks);
```
rvansa, line 1011: Note: ticks equal to nanos on Linux, but Windows could use different granularity.

src/hotspot/share/runtime/crac.cpp
```diff
@@ -824,7 +860,10 @@ GrowableArray<crac::score> crac::get_image_scores_from_jvm() {
   // Uptime since initial boot
   scores.append({"vm.uptimeSinceBoot", uptime});
   // Uptime since start of latest restore or -1 if not restored - same as CRaCMXBean.getUptimeSinceRestore()
-  scores.append({"vm.uptimeSinceRestore", _generation > 1 ? TimeHelper::counter_to_millis(crac::uptime_since_restore()) : -1});
+  scores.append({"vm.uptimeSinceRestore", static_cast<double>(_generation > 1 ? crac::uptime_since_restore() / NANOSECS_PER_MILLISEC : -1)});
+
+  scores.append({"vm.crac.generation", static_cast<double>(_generation)});
+  scores.append({"vm.restore.nativeTime", static_cast<double>(_generation > 1 ? (_restore_native_end_nanos - _restore_start_nanos) / NANOSECS_PER_MILLISEC : -1)});
```
TimPushkin, line 866: Are the -1 fallbacks (and initializing `_restore_native_end_nanos = -1` at the top of the file) really needed when we guard the reads with `_generation > 1`?

rvansa: The initialization isn't necessary, this is the only read. But about the fallback - do you suggest to include the metric only if _generation > 1? I have followed the same pattern as with `vm.uptimeSinceRestore` above, rather than omitting this entirely.

Btw., with keep-running this value is set even if this is a booted (not restored) instance. In that case it's a bit confusing (has some low value) but I don't think it's critical.

TimPushkin: Looks like I misread something here, nevermind.

Regarding setting this for keep-running, currently we do not differentiate "real" restores from keep-running restores on JVM level, IIRC keep-running is just another engine option opaque to JVM, so this is the correct behavior. If we want to change this (which I don't oppose), it should be a part of a larger change, probably including something in the engine API to tell the JVM that we are still in the same process.

TimPushkin (approved):
LGTM, but there's a conflict

TimPushkin (approved):

src/hotspot/share/runtime/crac.cpp
rvansa: I would definitely leave `keep_running` as an opaque option, but we could have C/R API extension that would let us find the current state (and also maybe track generation in the engine). Related to https://github.com/openjdk/crac/pull/317/changes#diff-e1dc6928b2a9d695492bf6d17382545e053916ab26e4dcbcafd2b050330225ae

